PR #26, opened 2023-02-26
modify src/CMakeLists.txt from whisper.cpp/Makefile

katsu560:
I modified src/CMakeLists.txt to support only AVX pc and others.
I referenced whisper.cpp/Makefile and added codes for each systems.
I'm sorry that I couldn't test Arm MacOS, AVX2 and FMA supported Linux, Haiku systems.


src/CMakeLists.txt
```diff
@@ -51,6 +133,12 @@ if (APPLE AND NOT GGML_NO_ACCELERATE)
     endif()
 endif()
 
+# use OpenBLAS
+if (OPENBLAS)
+    set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DGGML_USE_OPENBLAS -I/usr/local/include/openblas")
+    set(CMAKE_LD_FLAGS  "${CMAKE_LD_FLAGS} -lopenblas")
+endif()
+
```
ggerganov, line 141: Is this needed?

We already have this:

https://github.com/ggerganov/whisper.cpp/blob/b4ebdb6b57dedf9e82eec6e13ac70e8089e06190/CMakeLists.txt#L102-L114

katsu560: No, I'm sorry for my less-confirmation.

PR #38, opened 2023-03-16
Add pipe input for prompt on gpt examples

hidenorly:
Enable prompt input through pipe, instead of using -p option. This makes easier to give longer and multiple lines for the prompt.

Test:
 $ echo "This is an example" > prompt.txt
 $ cat prompt.txt | ./bin/gpt-j -m models/gpt-j-6B/ggml-model.bin
 $ cat promot.txt | ./bin/gpt-2 -m models/gpt-2-117M/ggml-model.bin

Note that -p option and no -p specified case are kept.
 $ ./bin/gpt-j -m models/gpt-j-6B/ggml-model.bin -p "This is an example"
 $ ./bin/gpt-j -m models/gpt-j-6B/ggml-model.bin
 $ ./bin/gpt-2 -m models/gpt-2-117M/ggml-model.bin -p "This is an example"
 $ ./bin/gpt-2 -m models/gpt-2-117M/ggml-model.bin

### DIFF
--- a/examples/gpt-2/main.cpp
+++ b/examples/gpt-2/main.cpp
@@ -10,6 +10,8 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <iostream>
+#include <unistd.h>
 
 // default hparams (GPT-2 117M)
 struct gpt2_hparams {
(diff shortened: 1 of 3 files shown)


hidenorly:
Thank you for your swift action!

PR #69, opened 2023-04-06
Refactor gpt-j and gpt-2 for more modularity.

pikalover6:
This splits the main.cpp files of gpt-2 and gpt-j into gpt-\*.cpp, gpt-\*.h, and main.cpp.
The main function has been moved to main.cpp, and functions using the newly added header files.
The overall goal is to move towards what is done with llama.cpp.
It has no affect on functionality to the user.
However, I am not sure if this is what you have in mind for this project, so I am releasing this a draft.
Look forward to hearing your thoughts.


ggerganov:
The gpt model / layer structs should not be exposed to the user.
An API should expose just function calls like `load`, `eval`, etc.

Issue #72, opened 2023-04-07
convert-h5-to-ggml.py does not match the official convert-ckpt-to-ggml.py


yunfanye:
This causes a wrong shape error when running the c++ implementation:

gpt2_model_load: tensor 'model/h0/attn/c_attn/w' has wrong shape in model file: got [768, 2304], expected [2304, 768]

PR #73, opened 2023-04-07
Update convert-h5-to-ggml.py to match GPT2 expectation

yunfanye:
1) Transpose additional mlp and attention layers.
2) Keep wpe as float32.

#72


ggerganov:
I believe this is outdated

PR #75, opened 2023-04-09
Fix to pass all tests

iacore:
I'm not sure if x/0 = 0 is reasonable.


ggerganov:
`nth` will never be 0

PR #77, opened 2023-04-10
Fixed quantization for f16 models not working in GPT-2 and GPT-J examples

LostRuins:
Fixed quantization for f16 models not working in GPT-2 and GPT-J examples, because the f16 tables were not initialized thus f16 to f32 conversion was failing.


ggerganov:
Yeah, I keep making this mistake. Need to fix this somehow in a better way

PR #113, opened 2023-04-27
alibi implementation by @hhamud

danforbes:
See discussion here https://github.com/rustformers/llama-rs/pull/141#issuecomment-1523900931

CC: @philpax @KerfuffleV2

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -269,6 +269,7 @@ extern "C" {
         GGML_OP_DIAG_MASK_INF,
         GGML_OP_SOFT_MAX,
         GGML_OP_ROPE,
+        GGML_OP_ALIBI,
         GGML_OP_CONV_1D_1S,
         GGML_OP_CONV_1D_2S,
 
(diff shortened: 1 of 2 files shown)


philpax:
For some additional clarification: this was originally implemented by @NouamaneTazi in https://github.com/NouamaneTazi/bloomz.cpp to implement the BLOOM model. We've implemented it in `llama-rs`, but we'd like to avoid patching GGML for this one operation.

There may be a way to implement this operation in terms of the existing primitives, as hypothesised by @KerfuffleV2, but we haven't fully investigated this possibility.

Green-Sky:
you can add authors etc to git commits. I think that would be better than the github mentions.

danforbes:
> you can add authors etc to git commits. I think that would be better than the github mentions.

Indeed, however, I do not know their email address.

philpax:
The email address should be present in the commits; I don't think GitHub will show you, but you can see it by viewing the commit in your Git tool of choice.

danforbes:
> The email address should be present in the commits; I don't think GitHub will show you, but you can see it by viewing the commit in your Git tool of choice.

All set ✅

ggerganov (approved):
Thanks! Great contribution

PR #117, opened 2023-04-29
Convert interleaved addressing to sequential addressing for reduce functions

tanmaysachan:
Using sequential addressing for better spatial locality while adding vectors.


ggerganov:
Do you observe a performance gain with this change?

tanmaysachan:
@ggerganov Sorry for the delay, and missing on reporting the stats.

Yes, there seems to be a considerable improvement through this approach. On my MacBook Air M1, while running the GPT-J model (6B params), I was able to get the following stats:

With interleaved - 
`main: mem per token = 15503156 bytes
main: load time = 9565.39 Ms
main: sample time = 74.51 ms
main: predict time = 378598.06 ms / 29122.93 ms per token
main: total time = 419232.00 ms`

With sequential -
`main: mem per token = 15503156 bytes
main:     load time = 10750.99 ms
main:   sample time =    73.46 ms
main:  predict time = 358448.09 ms / 27572.93 ms per token
main:    total time = 398988.81 ms`

There seems to be around 1600 ms of speedup in the prediction per token.

tanmaysachan:
@ggerganov Could you or a few people give this a shot and see if this is better for heavier models?

ggerganov:
Yes, will take a look soon - sorry for the delay

ggerganov (approved):
There does not seem to be any difference on M1 Pro.
Tried F16 and F32 GPT-2 models of different sizes

Merging this based on the report that on your system makes a difference

PR #130, opened 2023-05-04
Vocab support for special tokens

mverrilli:
Adding support for special tokens. This is in order to give priority to these tokens during tokenization in cases where the model was trained specifically around using them.

Note: There is probably some limitation around regex.

### DIFF
--- a/examples/common.cpp
+++ b/examples/common.cpp
@@ -192,13 +192,33 @@ std::map<std::string, int32_t> json_parse(const std::string & fname) {
     return result;
 }
 
+void gpt_vocab::add_special_token(const std::string &token) {
+    special_tokens.push_back(token);
+}
+
+
 std::vector<gpt_vocab::id> gpt_tokenize(const gpt_vocab & vocab, const std::string & text) {
     std::vector<std::string> words;
 
+ 
     // first split the text into words
     {
         std::string str = text;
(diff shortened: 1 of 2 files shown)


mverrilli:
Hi @ggerganov I needed this in order to line up with how Dolly was trained as the special tokens needed priority.  Let me know if there is a better way to accomplish this? I assume other models may run into similar issues. Thanks! https://github.com/mverrilli/ggml/tree/dolly-v2/examples/dolly-v2

Green-Sky:
afaik llama.cpp uses weighted vocab https://github.com/ggerganov/llama.cpp/blob/master/llama.cpp#L189-L200

mikev-db:
@Green-Sky Ahh, I was working from the common libs which has a regex style tokenizer. This is just so it can support the example better (Dolly special tokens have a space in them), but I think if I were to create a project out of it I'd redo that part. Thanks for pointing that out.

mverrilli:
@Green-Sky Ahh, I was working from the common libs which has a regex style tokenizer. This is just so it can support the example better (Dolly special tokens have a space in them), but I think if I were to create a project out of it I'd redo that part. Thanks for pointing that out.

mverrilli:
Closing this as it is included in #132

PR #132, opened 2023-05-05
Dolly v2 example

mverrilli:
Used stablelm as the starter. Added in special token parsing to allow the regex tokenizer to prioritize detection of Dolly's special tokens since they contain spaces.


ggerganov (approved):

xingchensong:
@mverrilli Hi, thanks for your great work, have you tried int4 quantization?

mverrilli:
Hi @xingchensong - This was mostly a spin off of the stablelm and I just filled in a few blanks.  It was mostly intended as a POC that ggml will work with dolly-v2.  I have tried it with Q4 and Q5. I found Q5 to be better (not much difference in space, faster, etc). 

One issue with the example here is the tokenizer. The tokenizer used here is a simple one, but the BPE tokenizer works out better. I wrote a BPE tokenizer working with ggml and Dolly-V2, but I'm still in the process of cleaning it up. 

I think the intent is to move to gpt-neox example for all of them.  I've tested it and it works, but I have some concerns with the special tokens. I'll update this to make it compatible with the latest ggml and submit a PR then I'll do some side-by-side tests and see if anything needs to change in gpt-neox.

ggerganov:
> I think the intent is to move to gpt-neox example for all of them

This would be great

PR #142, opened 2023-05-10
fix conversion from 'const int64_t' to 'const int'

wuyudi:
fix warning like

```c
D:\a\whisper.cpp\whisper.cpp\ggml.c(5966,22): warning C4244: '=': conversion from 'int64_t' to 'int', possible loss of data [D:\a\whisper.cpp\whisper.cpp\build\whisper.vcxproj]
```

in win build.


wuyudi:
conflict is too huge, close this.

Issue #131, opened 2023-05-04
replit-code-v1-3b


vgrichina:
https://huggingface.co/replit/replit-code-v1-3b gives solid results for both code generation and chat:

<img width="516" alt="Screenshot 2023-05-04 at 13 11 26" src="https://user-images.githubusercontent.com/126710/236318440-e0369988-8ec4-4813-b630-e18f28dc1f59.png">

another example here:
https://twitter.com/vgrichina/status/1653872353825419264

TBH seems like best assistant behaviour from < 3B model, would be super cool to get it running on user devices.

Green-Sky:
xref https://github.com/ggerganov/llama.cpp/issues/1299

lukasmoellerch:
I'd be interested in this as well, will start looking into it tomorrow if nobody else is working on it so far...

lukasmoellerch:
<img width="612" alt="Screenshot 2023-05-07 at 22 36 43" src="https://user-images.githubusercontent.com/4623046/236701451-cdce125c-a980-4b81-bc7c-12fb84091b07.png">

Getting there... Just kinda messed up the alibi attention bias I think.

Also: The tokenizer uses unigrams and is based on sentence-piece. What's the idea here? Integrating an optimal unigram tokenizer (some construction with lattices is required to make it efficient I think) would increase the complexity a lot..

Leoputera2407:
@lukasmoellerch Could you share your branch, maybe a second eye can help?

lukasmoellerch:
Sure, branch is here: https://github.com/lukasmoellerch/ggml/tree/replit - The temp=0 output is completely is correct if alibi attention bias is disabled both in their implementation and in mine, thus I am convinced that the problem is in that part somewhere. I suspect that we might need to do something like

```cpp
double bias = (1 - n_seq + n_past) * m_k;
pdst[0] = bias + src[0];
```

in ggml_compute_forward_alibi_f32, which gives me the correct output bias for n_seq = 2048, but the output seems to be unaltered.

Leoputera2407:
It seems like the alibi-bias in replitLM is calculated differently from how ggml calculates the alibi-bias.
[...]

PR #143, opened 2023-05-10
Fix bug in ggml_alibi

yangyaofei:
According to https://github.com/NouamaneTazi/bloomz.cpp/issues/27, I found there is a wrong implementation in `ggml_alibi`.

Now, `ggml_alibi` will get same result compare to the `build_alibi_tensor` in `transformers.models.bloom.modeling_bloom`

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -9425,7 +9425,7 @@ static void ggml_compute_forward_alibi_f32(
                     m_k = powf(m1, 2 * (k - n_heads_log2_floor) + 1);
                 }
 
-                pdst[0] = (j+1) * m_k + src[0];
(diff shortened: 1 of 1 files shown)


Green-Sky:
how does this relate to the discussion here https://github.com/ggerganov/ggml/issues/131#issuecomment-1540197345 ?

lukasmoellerch:
replit implementation seems to be fixed after this change

ggerganov:
Merging this based on the positive feedback

wangkuiyi:
Thank you @yangyaofei for this fix! I used the old ALiBi in another model and suffered a lot from it.

yangyaofei:
@wangkuiyi you are welcome 😁

Issue #136, opened 2023-05-08
MosaicML MPT-7B


ehartford:
As was suggested in https://github.com/ggerganov/llama.cpp/issues/1333

Create a basic inference example for MosaicML MPT-7B model

lukasmoellerch:
Looks decent, I'll create one PR for both then.

<img width="656" alt="Screenshot 2023-05-10 at 20 11 37" src="https://github.com/ggerganov/ggml/assets/4623046/bde122c2-c19d-4a44-bda8-894a563ea3c4">

PR #145, opened 2023-05-10
Replit + MPT

lukasmoellerch:
Implements #131 #136

Adds example code for mpt (https://huggingface.co/mosaicml/mpt-7b) and replit (https://huggingface.co/replit/replit-code-v1-3b). The code isn't too clean at the moment and I'll happily clean things up and implement suggestions, but might only be able to spend more time on this over the weekend.

Some hyperparameters are hardcoded such as the ffn / mlp ratio and the alibi max bias. Also, not all mpt-style models are supported, qkv clamping isn't implemented and a couple of other options aren't considered either.

The unigram tokenizer is comparably slow, but implementing a good one would add considerably more code to the example.

Also: Thank you @Leoputera2407 for helping with debugging the alibi problem


lukasmoellerch:
I can also merge both models into one example if that's preferred.

klosax:
Does this support the MPT-7B-StoryWriter model with 65k context length?

lukasmoellerch:
> Does this support the MPT-7B-StoryWriter model with 65k context length?

I didn't try it yet, but architecturally there is no difference between the base model and the story-writer fine-tuned model as to my knowledge. I can potentially try it tomorrow.

abhi-mosaic:
Thank you so much @lukasmoellerch for building support for MPT models! All of us @ MosaicML are very excited :)

>  but architecturally there is no difference between the base model and the story-writer fine-tuned model

One thing I want to point out, the StoryWriter model does have two arch changes vs the other models. It uses `alibi_bias_max=16` and `clip_qkv=6`. These were due to the long context setting. You can see the exact config here: https://huggingface.co/mosaicml/mpt-7b-storywriter/blob/main/config.json

Leoputera2407:
Adding support for max_alibi_bias, seems quite do-able @lukasmoellerch. 

I’m trying out a qkv_clamp on my branch, let me see if i can get that down by tomorrow. Really looking forward to try out story-writer on cpu, altho I wonder how much ram is requires to support 65k context length.

Leoputera2407:
I think folks in nomic-ai implemented a full ggml version of mpt with story-writer support

https://github.com/nomic-ai/gpt4all/blob/f8fdcccc5d253229808c0ceb9c5faae1ba42f68c/gpt4all-backend/mpt.cpp#L2

lukasmoellerch:
Sounds good, it seems like a lot of people are excited about the storywriter model, let’s get it integrated as well then, both modifications sound rather straightforward. @Leoputera2407 can you share what you’ve done regarding qkv clipping so far?

klosax:
Inference of the StoryWriter model fails:

```
/main -m mpt-7b-storywriter-ggml-f16.bin 

main: seed = 1683800454
mpt_model_load: loading model from 'mpt-7b-storywriter-ggml-f16.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 65536
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 1
mpt_model_load: ggml ctx size = 12683.13 MB
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 30479021776, available 13299222016)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 30479021776, available 13299222016)
Segmentation fault (core dumped)
```
Quantized model:

```
./main -m mpt-7b-storywriter-ggml-q5_1.bin 

main: seed = 1683800655
mpt_model_load: loading model from 'mpt-7b-storywriter-ggml-q5_1.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 65536
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 9
mpt_model_load: ggml ctx size = 4756.88 MB
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 22167746256, available 4987946496)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 22167746256, available 4987946496)
Segmentation fault (core dumped)
```

The base model works.

Green-Sky:
> needed 30479021776

29gigs anyone?

are you guys using f16 for the kv ?

lukasmoellerch:
> > needed 30479021776
> 
> 29gigs anyone?
> 
> are you guys using f16 for the kv ?

It seems like the context size calculation is actually wrong, but in the wrong direction i.e. it calculates with f32, but later allocates an f16... I'll investigate later.

klosax:
> It seems like the context size calculation is actually wrong, but in the wrong direction i.e. it calculates with f32, but later allocates an f16... I'll investigate later.

I found a solution: I changed the types in the ctx_size calculation to uint64_t and changed the calculation of memory_k and memory_v to type F16
```
uint64_t ctx_size = 0;

  {
    const auto &hparams = model.hparams;

    const uint64_t n_embd = hparams.d_model;
    const uint64_t n_layer = hparams.n_layers;
    const uint64_t n_ctx = hparams.max_seq_len;
    const uint64_t n_vocab = hparams.n_vocab;

(...)

    ctx_size +=   n_ctx * n_layer * n_embd * ggml_type_sizef(GGML_TYPE_F16); // memory_k
    ctx_size +=   n_ctx * n_layer * n_embd * ggml_type_sizef(GGML_TYPE_F16); // memory_v

```
Working output:

```
./main -m mpt-7b-storywriter-ggml-f16.bin 
main: seed = 1683806544
mpt_model_load: loading model from 'mpt-7b-storywriter-ggml-f16.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 65536
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 1
mpt_model_load: ggml ctx size = 45451.13 MB
mpt_model_load: memory_size = 32768.00 MB, n_mem = 2097152
mpt_model_load: ........................ done
mpt_model_load: model size = 12683.02 MB / num tensors = 194
main: number of tokens in prompt = 1
main: token[0] =    510

The last few years, she'd been able to come and go, she didn't come.
```

klosax:
The generation is not very good, it seems to keep repeating itself after a while:

Base model:

```
./main -m mpt-7b-ggml-f16.bin -p "Once upon"

main: seed = 1683807627
mpt_model_load: loading model from 'mpt-7b-ggml-f16.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 2048
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 1
mpt_model_load: ggml ctx size = 13707.13 MB
mpt_model_load: memory_size =  1024.00 MB, n_mem = 65536
mpt_model_load: ........................ done
mpt_model_load: model size = 12683.02 MB / num tensors = 194
main: number of tokens in prompt = 2
main: token[0] =  10758
main: token[1] =   2220

Once upon a time, there was a boy and a girl. They went to a wedding. And while they were there, they met a woman.
They danced with her. They talked. The girl, and boy went away. And they walked. The girl with them. He, they danced.
But they had. They danced, went. Then they, he said, and talked, and went to go.
But, the night, they. And they, and walked. They were, they. They got And did. And they. And walked. And had,
```
StoryWriter model:

```
./main -m mpt-7b-storywriter-ggml-f16.bin -p "Once upon"

main: seed = 1683807952
mpt_model_load: loading model from 'mpt-7b-storywriter-ggml-f16.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 65536
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 1
mpt_model_load: ggml ctx size = 45451.13 MB
mpt_model_load: memory_size = 32768.00 MB, n_mem = 2097152
mpt_model_load: ........................ done
mpt_model_load: model size = 12683.02 MB / num tensors = 194
main: number of tokens in prompt = 2
main: token[0] =  10758
main: token[1] =   2220

Once upon a time there was a boy named Nick who lived in Brooklyn and grew up to be a man in Manhattan.
His father was a businessman and his mother worked as a nurse. Nick had a job working working working for the dad.
Nick was working working working Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick Nick
```

Green-Sky:
yea i can see the same
```
$ bin/mpt -m ../examples/mpt/models/mpt-7b-storywriter/mpt-7b-storywriter-ggml_v0-q4_0.bin
main: seed = 1683813966
mpt_model_load: loading model from '../examples/mpt/models/mpt-7b-storywriter/mpt-7b-storywriter-ggml_v0-q4_0.bin' - please wait ...
mpt_model_load: d_model       = 4096
mpt_model_load: max_seq_len   = 65536
mpt_model_load: n_heads       = 32
mpt_model_load: n_layers      = 32
mpt_model_load: n_vocab      = 50432
mpt_model_load: ftype   = 2
mpt_model_load: ggml ctx size = 36732.25 MB
mpt_model_load: memory_size = 32768.00 MB, n_mem = 2097152
mpt_model_load: ........................ done
mpt_model_load: model size =  3964.14 MB / num tensors = 194
main: number of tokens in prompt = 1
main: token[0] =   2993

She was so fat that I was not sure how she could still stand up walk. And I was not sure that I was so fat. And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And And^C
```

there seems no repeat penalty and other stuff. needs some updates from the llama.cpp codebase.

klosax:
> there seems no repeat penalty and other stuff. needs some updates from the llama.cpp codebase.

I would love to see the common infrastructure of llama.cpp become something like "ggml-llm" and the code for the specific llm architectures (llama, gpt-2, gpt-j, mpt and others) become like add-ons at compile time.

Green-Sky:
I uploaded some ggml files, so we can test this more easily https://huggingface.co/Green-Sky/ggml-mpt-7b-storywriter

Leoputera2407:
Haven't completely tested but fwiw, here's qkv_clamp
``` ggml.c
// ggml_clamp

struct ggml_tensor * ggml_clamp_impl(
        struct ggml_context * ctx,
        struct ggml_tensor * a,
        float min_val,
        float max_val,
        bool inplace) {
    bool is_node = false;

    if (!inplace && (a->grad)) {
        is_node = true;
    }

    struct ggml_tensor * result = inplace ? ggml_view_tensor(ctx, a) : ggml_dup_tensor(ctx, a);

    result->op = GGML_OP_CLAMP;
    result->src0 = a;
    result->src1 = NULL;
    result->clamp_min = min_val;
    result->clamp_max = max_val;

    return result;
}

struct ggml_tensor * ggml_clamp(
        struct ggml_context * ctx,
        struct ggml_tensor * a,
        float min_val,
        float max_val) {
    return ggml_clamp_impl(ctx, a, min_val, max_val, false);
}

struct ggml_tensor * ggml_clamp_inplace(
        struct ggml_context * ctx,
        struct ggml_tensor * a,
        float min_val,
        float max_val) {
    return ggml_clamp_impl(ctx, a, min_val, max_val, true);
}
```


In main.cpp
```
mpt_eval(...){
....
    const uint64_t n_embd = hparams.d_model;
    const uint64_t n_layer = hparams.n_layers;
    const uint64_t n_ctx = hparams.max_seq_len;
    const uint64_t n_head = hparams.n_heads;
    const uint64_t n_vocab = hparams.n_vocab;
    const uint64_t qkv_clip = hparams. clip_qkv;
....
// compute QKV
      { cur = ggml_mul_mat(ctx0, model.layers[il].c_attn_wqkv_weight, cur); }
      
      // TODO: clip_qkv
float clip_qkv = 6.0f; // set this to your desired value, use 0 or a negative value to disable clamping
if (clip_qkv > 0) {
    ggml_clamp_inplace(ctx0, cur, -clip_qkv, clip_qkv);
}

      struct ggml_tensor *Qcur = ggml_view_2d(ctx0, cur, n_embd, N, cur->nb[1],
                                              0 * sizeof(float) * n_embd);
      struct ggml_tensor *Kcur = ggml_view_2d(ctx0, cur, n_embd, N, cur->nb[1],
                                              1 * sizeof(float) * n_embd);
      struct ggml_tensor *Vcur = ggml_view_2d(ctx0, cur, n_embd, N, cur->nb[1],
                                              2 * sizeof(float) * n_embd);
       ....
       }
```

ggerganov:
Looks like great progress - will be taking a more detailed look soon

> > there seems no repeat penalty and other stuff. needs some updates from the llama.cpp codebase.
> 
> I would love to see the common infrastructure of llama.cpp become something like "ggml-llm" and the code for the specific llm architectures (llama, gpt-2, gpt-j, mpt and others) become like add-ons at compile time.

Yes, this would be great. Now that we have various examples of LLM inference and I have a better understanding of the general API structure that is necessary, it will be easier to come up with a way to unify all these into a single interface

apage43:
wrt alibi I believe the multiplier needs to range from `-ne0+1` to `0` - the current fix that's landed and the one here still has it range `0` to `ne0 - 1`. 

That is, the *current*/*last* token should have a bias of 0 and the previous tokens need to have a *negative* bias, matching the `arange` here: https://huggingface.co/mosaicml/mpt-7b/blob/main/attention.py#L269

this is the implementation currently in gpt4all https://github.com/manyoso/llama.cpp/blob/master/ggml.c#L9393-L9394 - I'm also still not 100% sure this is right as the quality still seems noticeably lacking compared to the original. I also wound up changing outer powf to a multiply in the inner powf which seemed to avoid some NaN crashes

examples/mpt/main.cpp
```diff
@@ -0,0 +1,737 @@
...
+          ctx0, ggml_repeat(ctx0, model.layers[il].norm_2_weight, cur), cur);
+    }
+
+    // n = self.mlp(m)
+    {
+
+      cur = ggml_mul_mat(ctx0, model.layers[il].ffn_up_proj, cur);
+
+      // GELU activation
+      cur = ggml_gelu(ctx0, cur);
```
ggerganov, line 533: Shouldn't we use `ggml_relu()` here?

https://huggingface.co/mosaicml/mpt-7b/blob/main/config.json#L31

examples/mpt/main.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+  struct ggml_tensor *inpL = ggml_get_rows(ctx0, model.wte_weight, embd);
+
+  for (int il = 0; il < n_layer; ++il) {
+
+    struct ggml_tensor *cur;
+
+    // a = self.ln_1(x)
+    {
+      cur = ggml_norm(ctx0, inpL);
```
ggerganov, line 411: I think we have to use `ggml_rms_norm()` instead of `ggml_norm()`:

https://huggingface.co/mosaicml/mpt-7b/blob/main/norm.py#L27-L31

ggerganov (changes requested):
Use 4 spaces for indentation

examples/mpt/main.cpp
```diff
@@ -0,0 +1,737 @@
...
+  // norm
+  {
+    inpL = ggml_norm(ctx0, inpL);
+    // inpL = ln_f_g*inpL
+    inpL = ggml_mul(ctx0, ggml_repeat(ctx0, model.norm_f_weight, inpL), inpL);
+  }
+
+  // output embedding weight tied to input embedding
+  inpL = ggml_mul_mat(ctx0, model.wte_weight, inpL);
+
```
ggerganov, line 553: I could be missing something, but it looks like we first have to multiply with `wte` and after that to normalize:

https://huggingface.co/mosaicml/mpt-7b/blob/main/modeling_mpt.py#L186

examples/replit/main.cpp
```diff
@@ -0,0 +1,850 @@
...
+    end = start;
+    start = next_start;
+  }
+  const auto token_id = model.at(word.substr(start, end - start)).first;
+  tokens.insert(tokens.begin(), token_id);
+  return std::make_tuple(tokens, score);
+}
+
+auto replit_tokenizer_load(replit_tokenizer &tokenizer, std::istream &fin,
+                           int max_vocab_size) -> bool {
```
ggerganov, line 79: Do not use `auto` in function declaration return type

examples/replit/main.cpp
```diff
@@ -0,0 +1,850 @@
...
+  int32_t max_seq_len = 0;
+  int32_t n_heads = 0;
+  int32_t n_layers = 0;
+  int32_t n_vocab = 0;
+  int32_t ftype = 0;
+};
+
+struct replit_layer {
+  // pre normalization
+  struct ggml_tensor *ln_1_weight;
```
ggerganov, line 152: The style for pointers and reference is "space before and after":

```suggestion
  struct ggml_tensor * ln_1_weight;
```

examples/mpt/main.cpp
apage43: the MPT-7B models are using `LPLayerNorm` https://huggingface.co/mosaicml/mpt-7b/blob/main/norm.py#L14

config: https://huggingface.co/mosaicml/mpt-7b/blob/main/config.json#L44

ggerganov:
Any reference document / code for the "qkv clamping" so I can get familiar?

klosax:
> Any reference document / code for the "qkv clamping" so I can get familiar?

" clip_qkv (Optional[float]): If not None, clip the queries, keys, and values in the attention layer to this value. "

https://huggingface.co/mosaicml/mpt-7b-storywriter/blob/main/configuration_mpt.py#L28
https://huggingface.co/mosaicml/mpt-7b-storywriter/blob/main/attention.py#L156

klosax:
It seems we need  a working ggml_compute_forward_alibi and a GGML_OP for clamping.

Here is reference for the ALiBi attention:

https://arxiv.org/pdf/2108.12409.pdf

https://huggingface.co/mosaicml/mpt-7b-storywriter/blob/main/attention.py#L268

examples/mpt/main.cpp
```diff
@@ -0,0 +1,764 @@
...
+    size_t ctx_size = 0;
+
+    {
+        const auto & hparams = model.hparams;
+
+        const int n_embd = hparams.d_model;
+        const int n_layer = hparams.n_layers;
+        const int n_ctx = hparams.max_seq_len;
+        const int n_vocab = hparams.n_vocab;
```
Green-Sky, line 143: https://github.com/ggerganov/ggml/pull/145#issuecomment-1543881833

Green-Sky (changes requested):

examples/replit/main.cpp
```diff
@@ -0,0 +1,870 @@
...
+                               ggml_type_sizef(wtype)); // mlp_mlp_up_weight
+
+        // mlp_mlp_down_weight
+        ctx_size += n_layer * (n_embd * n_embd * 4 *
+                               ggml_type_sizef(wtype)); // mlp_mlp_down_weight
+
+        ctx_size += n_ctx * n_layer * n_embd *
+                    ggml_type_sizef(GGML_TYPE_F32); // memory_k
+        ctx_size += n_ctx * n_layer * n_embd *
+                    ggml_type_sizef(GGML_TYPE_F32); // memory_v
```
klosax, line 275: Change to GGML_TYPE_F16 as they are initialized to that type.

examples/mpt/main.cpp
```diff
@@ -0,0 +1,764 @@
...
+                               ggml_type_sizef(wtype)); // mlp_mlp_up_weight
+
+        // mlp_mlp_down_weight
+        ctx_size += n_layer * (n_embd * n_embd * 4 *
+                               ggml_type_sizef(wtype)); // mlp_mlp_down_weight
+
+        ctx_size += n_ctx * n_layer * n_embd *
+                    ggml_type_sizef(GGML_TYPE_F32); // memory_k
+        ctx_size += n_ctx * n_layer * n_embd *
+                    ggml_type_sizef(GGML_TYPE_F32); // memory_v
```
klosax, line 169: Change to GGML_TYPE_F16 as they are initialized to that type.

lukasmoellerch:
> wrt alibi I believe the multiplier needs to range from `-ne0+1` to `0` - the current fix that's landed and the one here still has it range `0` to `ne0 - 1`.
> 
> That is, the _current_/_last_ token should have a bias of 0 and the previous tokens need to have a _negative_ bias, matching the `arange` here: https://huggingface.co/mosaicml/mpt-7b/blob/main/attention.py#L269
> 
> this is the implementation currently in gpt4all https://github.com/manyoso/llama.cpp/blob/master/ggml.c#L9374-L9394 - I'm also still not 100% sure this is right as the quality still seems noticeably lacking compared to the original. I also wound up changing outer powf to a multiply in the inner powf which seemed to avoid some NaN crashes

Yes, we had some discussion about this in the issue code already - I somehow thought that the offset didn't matter because it would be layernorm'd away, but it obviously influences the softmax. I'll work on a fix tomorrow and will also make the alibi max bias configurable.

jploski:
> StoryWriter does love to repeat itself but I've found some settings that you can use with HF generate the tend to work pretty well.

These settings don't appear to help much with the GGML version. Also, disabling these settings on the spaces website does not lead to the kind of tragic endless loop after just a couple tokens mentioned previously in the comments - and which I also experience with the current GGML version of both mpt-7b and mpt-7b-storywriter.

To check whether there is maybe some "obvious" problem with the new GGML implementation, today I trained a very small version of MPT myself, using confirguration from mpt-7b-storywriter, but extremely scaled down (https://huggingface.co/jploski/mpt-mini-shakespeare - inspired by nanoGPT), then converted it to GGML (and quantized). My hope was that the endless repetitions would also reappear in this tiny model, making it easier to blame GGML. But unfortunately, here the GGML version works just as well as the Python version! The only stage during which I could observe the endless loop style repetitions (in Python) was during initial epochs of training.

So what can we conclude from that (if anything): maybe the published weights of the mpt-7b models are somehow wrong? Although this seems like a very unlikely theory... it doesn't hurt to ask: have the GGML contributors here managed to get sensible outputs from the non-GGML version of the published mpt-7b models that were used as input for the GGML conversion (i.e. comparable to https://huggingface.co/spaces/mosaicml/mpt-7b-storywriter)?

Green-Sky:
I get the feeling that int32_t is overflowing, in multiple places. that would atleast somewhat explain things.

eg: https://github.com/ggerganov/ggml/pull/145#issuecomment-1543881833

examples/mpt/main.cpp
klosax: Calculation of ctx_size will be erroneous unless type int is changed to uint64_t .

leroycep:
I downloaded this branch and it is running out of memory before evaluating this prompt:

````
Below is an instruction that describes a task. Write a response that appropriately completes the request.

### Instruction:
```csv
in € million,2021,2020
ASSETS,,
Intangible assets,12980,12342
"Property, plant and equipment",22390,21850
Leased products,44700,41995
Investments accounted for using the equity method,5112,3585
Other investments,1241,735
Receivables from sales financing,51712,48025
Financial assets,1715,2644
Deferred tax,2202,2459
Other assets,1302,1216
Non-current assets,143354,134851
Inventories,15928,14896
Trade receivables,2261,2298
Receivables from sales financing,35705,36252
Financial assets,5800,5108
Current tax,1529,606
Other assets,8941,9110
Cash and cash equivalents,16009,13537
Current assets,86173,81807
Total assets,229527,216658
Equity and Liabilities,,
Subscribed capital,661,660
Capital reserves,2325,2199
Revenue reserves,71705,59550
Accumulated other equity,–325,–1518
Equity attributable to shareholders of BMW AG,74366,60891
Minority interest,766,629
Equity,75132,61520
Pension provisions,1247,3693
Other provisions,7206,6488
Deferred tax,1458,509
Financial liabilities,62342,67390
Other liabilities,5676,5095
Non-current provisions and liabilities,77929,83175
Other provisions,6748,7494
Current tax,921,747
Financial liabilities,41121,38986
Trade payables,10932,8644
Other liabilities,16744,16092
Current provisions and liabilities,76466,71963
Total equity and liabilities,229527,216658
```

Give an analysis for the above balance sheet for BMW. Think step by step.
````

This is the it gives:
```
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 269375952, available 268435456)
```

It seems like the memory usage is being incorrectly calculated somewhere.

**Comment by klosax** [comment]
> It seems like the memory usage is being incorrectly calculated somewhere.

https://github.com/ggerganov/ggml/blob/011d2b13dfb04310f58ade88b3d714a10c567986/examples/mpt/main.cpp#LL140C8-L143C45

In main.cpp, change the lines in the link above to :

```
        const uint64_t n_embd = hparams.d_model;
        const uint64_t n_layer = hparams.n_layers;
        const uint64_t n_ctx = hparams.max_seq_len;
        const uint64_t n_vocab = hparams.n_vocab;
```

File: examples/mpt/main.cpp (hunk 6 above)
Comment (lukasmoellerch) [reply examples/mpt/main.cpp:143]: Wouldn't size_t make more sense? The result as passed into the context will be size_t anyway.

Comment (klosax) [reply examples/mpt/main.cpp:143]: Yes you're right.

**Comment by lukasmoellerch** [comment]
I added clip_qkv, max alibi bias support and adjusted the alibi bias index offset. I think it works much better already:
Screenshot 2023-05-14 at 12 37 48

Still seeing some repetitions after a while, but not as quickly

**Comment by jploski** [comment]
@lukasmoellerch Note that the code from the replit branch does not complie using gcc 8.3.0. Apparently there are two invocations of make_tuple which could be fixed by changing them to make_pair:

```
/tmp/ggml/examples/replit/main.cpp: In function ‘std::pair<std::vector<long unsigned int>, float> encode_word(const string&, const piece_map_t&)’:
/tmp/ggml/examples/replit/main.cpp:54:31: error: could not convert ‘std::make_tuple(_Elements&& ...) [with _Elements = {std::vector<long unsigned int, std::allocator<long unsigned int> >, double}](0.0)’ from ‘std::tuple<std::vector<long unsigned int, std::allocator<long unsigned int> >, double>’ to ‘std::pair<std::vector<long unsigned int>, float>’
         return std::make_tuple(std::vector<std::size_t>{0}, 0.0);
                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/ggml/examples/replit/main.cpp:70:27: error: could not convert ‘std::make_tuple(_Elements&& ...) [with _Elements = {std::vector<long unsigned int, std::allocator<long unsigned int> >&, float&}](score)’ from ‘std::tuple<std::vector<long unsigned int, std::allocator<long unsigned int> >, float>’ to ‘std::pair<std::vector<long unsigned int>, float>’
     return std::make_tuple(tokens, score);
            ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
/tmp/ggml/examples/replit/main.cpp: In function ‘bool replit_model_load(const string&, replit_model&, replit_tokenizer&)’:
/tmp/ggml/examples/replit/main.cpp:275:19: warning: unused variable ‘n_ctx’ [-Wunused-variable]
         const int n_ctx = hparams.max_seq_len;
                   ^~~~~
make[2]: *** [examples/replit/CMakeFiles/replit.dir/build.make:63: examples/replit/CMakeFiles/replit.dir/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1222: examples/replit/CMakeFiles/replit.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

**Comment by lukasmoellerch** [comment]
> @lukasmoellerch Note that the code from the replit branch does not complie using gcc 8.3.0. Apparently there are two invocations of make_tuple which could be fixed by changing them to make_pair:
> 
> ```
> /tmp/ggml/examples/replit/main.cpp: In function ‘std::pair<std::vector<long unsigned int>, float> encode_word(const string&, const piece_map_t&)’:
> /tmp/ggml/examples/replit/main.cpp:54:31: error: could not convert ‘std::make_tuple(_Elements&& ...) [with _Elements = {std::vector<long unsigned int, std::allocator<long unsigned int> >, double}](0.0)’ from ‘std::tuple<std::vector<long unsigned int, std::allocator<long unsigned int> >, double>’ to ‘std::pair<std::vector<long unsigned int>, float>’
>          return std::make_tuple(std::vector<std::size_t>{0}, 0.0);
>                 ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> /tmp/ggml/examples/replit/main.cpp:70:27: error: could not convert ‘std::make_tuple(_Elements&& ...) [with _Elements = {std::vector<long unsigned int, std::allocator<long unsigned int> >&, float&}](score)’ from ‘std::tuple<std::vector<long unsigned int, std::allocator<long unsigned int> >, float>’ to ‘std::pair<std::vector<long unsigned int>, float>’
>      return std::make_tuple(tokens, score);
>             ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
> /tmp/ggml/examples/replit/main.cpp: In function ‘bool replit_model_load(const string&, replit_model&, replit_tokenizer&)’:
> /tmp/ggml/examples/replit/main.cpp:275:19: warning: unused variable ‘n_ctx’ [-Wunused-variable]
>          const int n_ctx = hparams.max_seq_len;
>                    ^~~~~
> make[2]: *** [examples/replit/CMakeFiles/replit.dir/build.make:63: examples/replit/CMakeFiles/replit.dir/main.cpp.o] Error 1
> make[1]: *** [CMakeFiles/Makefile2:1222: examples/replit/CMakeFiles/replit.dir/all] Error 2
> make: *** [Makefile:141: all] Error 2
> ```

Ok thanks - I am compiling it using clang thus I didn’t notice

**Comment by klosax** [comment]
> Still seeing some repetitions after a while, but not as quickly

I can confirm, it seems to work much better now:

```
./main -m mpt-7b-storywriter-ggml-q4_0.bin -p "Once upon a time" -n 4096 -s 33
main: seed = 33
mpt_model_load: loading model from 'mpt-7b-storywriter-ggml-q4_0.bin' - please wait ...
mpt_model_load: d_model        = 4096
mpt_model_load: max_seq_len    = 65536
mpt_model_load: n_heads        = 32
mpt_model_load: n_layers       = 32
mpt_model_load: n_vocab        = 50432
mpt_model_load: alibi_bias_max = 16.000000
mpt_model_load: clip_qkv       = 6.000000
mpt_model_load: ftype          = 2
mpt_model_load: ggml ctx size = 6012.19 MB
mpt_model_load: memory_size =  2048.00 MB, n_mem = 131072
mpt_model_load: ........................ done
mpt_model_load: model size =  3964.14 MB / num tensors = 194
main: number of tokens in prompt = 4
main: token[0] =  10758
main: token[1] =   2220
main: token[2] =    247
main: token[3] =    673

Once upon a time, in the land of Firsdad, a young, beautiful princess, fair of face, pure of heart,
and wise of mind, was born into the world. When she was but a babe, a great war was fought
between the forces of Good and the forces of Evil, but when she grew into womanhood,
the war came to an end, and all the kingdom rejoiced to see her. But the Evil forces were not
willing to relinquish the power they had gained, and they sent out the most powerful of their minions,
a great, dark dragon, to wreak destruction upon the land. In the end, he was defeated, but not
destroyed, and in the years that followed, the shadow of his evil remained, and the land of Firsdad
was plagued with terrible^C
```

**Comment by Green-Sky** [comment]
> Note that the code from the replit branch does not complie using gcc 8.3.0

also on gcc 9.4.0

**Comment by Green-Sky** [comment]
It's mostly working now. however i cant even get to 1k tokens. `ggml_new_tensor_impl: not enough space in the context's memory pool (needed 268507840, available 268435456)
Segmentation fault (core dumped)`

reuploading updated modelfiles rn.

**Comment by klosax** [comment]
> It's mostly working now. however i cant even get to 1k tokens. `ggml_new_tensor_impl: not enough space in the context's memory pool (needed 268507840, available 268435456) Segmentation fault (core dumped)`

The needed space for the buffer in mpt_eval seems to increase with each evaluated token.

**Comment by ggerganov** [comment]
Yeah, the logic for determining the necessary memory for inference is very bad overall - I will be thinking of ways to improve this in the future.

For now maybe try changing this:

```c++
    if (mem_per_token > 0 && mem_per_token * N > buf_size) {
        const size_t buf_size_new = 1.1 * (mem_per_token * N); // add 10% to account for ggml object overhead
```

for example to this:

```c++
    if (mem_per_token > 0 && mem_per_token * N > 0.9*buf_size) {
        const size_t buf_size_new = 1.1 * (mem_per_token * N); // add 10% to account for ggml object overhead
```

I think it will fix the issue for now

Green-Sky (approved):

lukasmoellerch:
@ggerganov I think we might want to separate the model max_seq_length (which e.g. is used in the alibi bias offset) and the amount of k / v memory slots we allocate for inference. I temporarily hardcoded n_ctx in mpt to 4096 because otherwise my MacBook Air wasn't too happy, but this should probably be an inference parameter - should we at least just set it to `params.n_predict + embd_inp.size()` such that the user can set it using a command line flag?

Green-Sky:
@lukasmoellerch cant run storywriter model anymore :laughing: , without or any quantization
```
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 9628746752, available 9628745984)
Segmentation fault (core dumped)
```

klosax:
After much trial and error I found a forumla for setting the memory buffer, as it needs more space with each evaluated token:
```
static size_t buf_size = 64 * 1024 * 1024 + 261 * 1024 * n_past;
```

lukasmoellerch:
> @lukasmoellerch cant run storywriter model anymore 😆 , without or any quantization
> 
> ```
> ggml_new_tensor_impl: not enough space in the context's memory pool (needed 9628746752, available 9628745984)
> Segmentation fault (core dumped)
> ```

Yes, I just merge master with the quantisation changes which means that the object overhead adjustment wasn't correct, still a bit broken though.

lukasmoellerch:
nvm, just had to re-quantize the model

lukasmoellerch:
> > should we at least just set it to params.n_predict + embd_inp.size() such that the user can set it using a command line flag?
> 
> Yes, I think this is a good workaround for now.

I looked into this a bit, but can't really get it to be clean. n_predict is the number of additional tokens, thus the total number of tokens depends on the tokenizer being loaded. On the other hand we need the prompt to know the number of tokens which we don't have in load (and also shouldn't have) - I think we either want to separate k/v tensor creation from model loading or have it as a separate parameter.

But I can also do that in a follow-up PR.

Let me know what I can still do in this pr.

ggerganov (approved):

ggerganov:
@lukasmoellerch and everyone else - thanks for this contribution

I'll probably play with this in the next days and will try to improve the memory allocation logic.

klosax:
Got this buffer allocation working well after analyzing how much buffer was needed depending on batch size and n_predict.

Be aware that large batch sizes will require large amount of memory.

The buffer is declared globally:
```
static void * buf;
size_t buf_size = 0;
```
mpt_eval() sets mem_per_batch instead of mem_per_token:

```mem_per_batch = ggml_used_mem(ctx0);```

mpt_eval() does not reallocate the buffer at all.


In main() the buffer needed is calculated and allocated as follows:
```
    // Allocate 64 MB initial buffer
    buf_size = 64 * 1024 * 1024;
    printf("%s: initial evaluation compute buffer = %.2f MB\n", __func__, buf_size/1024.0/1024.0);

    buf = malloc(buf_size);
    if (buf == nullptr) {
        fprintf(stderr, "%s: failed to allocate %.2f MB\n", __func__, buf_size/1024.0/1024.0);
        return 1;
    }

    // Get buf per token when batch size = 1
    size_t mem_per_batch = 0;
    size_t mem_per_batch2 = 0;
    mpt_eval(model, params.n_threads, 0, {0}, logits, mem_per_batch);
    mpt_eval(model, params.n_threads, 1, {0}, logits, mem_per_batch2);

    // Get buffer increase for each batch when batch size = 1
    size_t buf_add_per_batch_bs_one = mem_per_batch2 - mem_per_batch;
    printf("%s: buf increase at batch size 1 = %lu\n", __func__, buf_add_per_batch_bs_one);

    // Accumulate buf up to the batch size in use
    // For each batch, add buf increase for batch size = 1 plus overhead
    size_t buf_add_per_prompt_batch=0;
    for (int i=0; i< params.n_batch; i++) {
        buf_add_per_prompt_batch += buf_add_per_batch_bs_one + i * 8500;
    }

    // Number of batches in prompt
    size_t number_of_prompt_batches = (embd_inp.size()/params.n_batch + 1);
    printf("%s: number of batches in prompt = %lu\n", __func__, number_of_prompt_batches);

    // Initalize buffer to 13 MB per token for the batch size in use
    buf_size = size_t(params.n_batch) * size_t(13 * 1024 * 1024);
    printf("%s: buf, init = %.2f MB\n", __func__, buf_size/1024.0/1024.0);

    // Add buf needed for the prompt batches
    size_t buf_prompt_batches = buf_add_per_prompt_batch * number_of_prompt_batches;
    printf("%s: buf, prompt batches = %.2f MB\n", __func__, buf_prompt_batches/1024.0/1024.0);
    buf_size+= buf_prompt_batches;

    // Add buf needed for n_predict
    size_t buf_predict_tokens = buf_add_per_batch_bs_one * params.n_predict;
    printf("%s: buf, predict tokens = %.2f MB\n", __func__, buf_predict_tokens/1024.0/1024.0);
    buf_size+= buf_predict_tokens;

    printf("%s: total evaluation compute buffer = %.2f MB\n", __func__, buf_size/1024.0/1024.0);

    // Reallocate total buffer needed
    buf = malloc(buf_size);
    if (buf == nullptr) {
        fprintf(stderr, "%s: failed to allocate %.2f MB\n", __func__, buf_size/1024.0/1024.0);
        return 1;
    }

```

x4080:
I just tried today that the storywriter still repeating the story over and over again, is there any trick to avoid it ?
Thanks

klosax:
> I just tried today that the storywriter still repeating the story over and over again, is there any trick to avoid it ? Thanks

Repeat penalty is being implemented in pr #184 .

jploski:
Note that repetition_penalty from pr #184 (and also as implemented in llama.cpp) is not the same as no_repeat_ngram_size which is used in the MPT-7 HuggingFace space (https://huggingface.co/spaces/mosaicml/mpt-7b-storywriter):

https://github.com/huggingface/transformers/blob/2f424d79797ea5344f1b3ac241be1a181cfc220d/src/transformers/generation/utils.py#LL860C30-L860C48

PR #155, opened 2023-05-14
fix end of text token for santacoder

salahiguiliz:
Before this PR santacoder didn't stop at end of text token, this PR aims to fix it.


NouamaneTazi (approved):
LGTM. Thank you for taking care of this :)

PR #160, opened 2023-05-17
[WIP, EXAMPLE] Add BioGPT to the example section

PABannier:
As discussed [here](https://github.com/ggerganov/llama.cpp/issues/1328#issuecomment-1547007119), this PR adds to the example section the BioGPT model.

Apart from minor architectural changes, the model remains very similar to GPT-2 (medium). The main difference is the tokenization process in two-step: 1) Moses tokenizer 2) Byte-Pair encoding.

### DIFF
--- a/examples/biogpt/README.md
+++ b/examples/biogpt/README.md
@@ -1,0 +1,82 @@
+# biogpt.cpp
+
+Inference of BioGPT model in pure C/C++.
+
+## Description
+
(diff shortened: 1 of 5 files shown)


ggerganov:
I added a link to the README pointing to https://github.com/PABannier/biogpt.cpp. Will close this for now

I just saw you have added `encodec.cpp` example which sounds very exciting!

PR #173, opened 2023-05-20
mpt - Fix `mem_per_token` not incrementing

MonkiesDance:
Should fixes #171, #158 , maybe bigcode-project/starcoder.cpp#3

This appears to allow MPT models to work in extended context/story mode.

Issue #171, opened 2023-05-20
mpt: ggml_new_tensor_impl: not enough space in the context's memory pool 


MonkiesDance:
[removed]

Issue #158, opened 2023-05-16
starcoder -- not enough space in the context's memory pool


bluecoconut:
I'm getting errors with starcoder models when I try to include any non-trivial amount of tokens. I'm getting this with both my raw model (direct .bin) and quantized model regardless of version (pre Q4/Q5 changes and post Q4/Q5 changes). 

Relevant error:
```
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 412241472, available 411790368)
```

Example:
```bash
./build/bin/starcoder -m /workspaces/research/models/starcoder/starcoder-ggml.bin -p "def fibo( fibo fib fibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test " --top_k 0 --top_p 0.95 --temp 0.2 
```
will cause the error
```
main: seed = 1684223471
starcoder_model_load: loading model from '/workspaces/research/models/starcoder/starcoder-ggml.bin'
starcoder_model_load: n_vocab = 49152
starcoder_model_load: n_ctx   = 8192
starcoder_model_load: n_embd  = 6144
starcoder_model_load: n_head  = 48
starcoder_model_load: n_layer = 40
starcoder_model_load: ftype   = 1
starcoder_model_load: qntvr   = 0
starcoder_model_load: ggml ctx size = 51276.47 MB
starcoder_model_load: memory size = 15360.00 MB, n_mem = 327680
starcoder_model_load: model size  = 35916.23 MB
main: prompt: 'def fibo( fibo fib fibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test '
main: number of tokens in prompt = 51, first 8 tokens: 589 28176 97 26 28176 97 28176 28176 

def fibo( fibo fib fibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibo test waterfibggml_new_tensor_impl: not enough space in the context's memory pool (needed 412241472, available 411952576)
Segmentation fault (core dumped)
```

(Here's another output from the quantized model)
```
vscode ➜ /workspaces/research/others/ggml (master) $ ./build/bin/starcoder -m /workspaces/research/models/starcoder/starcoder-ggml-q4_1.bin -p "def fibo( fibo fib fibo test wate
[...]

PR #174, opened 2023-05-20
mpt - Add flags for prompt context size (-c/--ctx_size)

MonkiesDance:
mpt: Adds flags (`-c/--ctx_size`) for increasing prompt context size (e.g. from 4096 to 65536) for use with storywriter models.

In conjunction with #173, should improve/fix:
- https://github.com/ggerganov/ggml/issues/171, 
- https://github.com/ggerganov/ggml/issues/158 , 
- maybe https://github.com/bigcode-project/starcoder.cpp/issues/3
- https://github.com/ggerganov/llama.cpp/issues/1333#issuecomment-1537412858


MonkiesDance:
This was implemented separately in #184

Issue #170, opened 2023-05-19
gpt_tokenize : bug in tokenization and incapable of double-byte languages


jaeminSon:
There is a bug in gpt_tokenize function in examples/common.cpp 

Followings are comparison with a huggingface tokenizer, 
```
# huggingface
>>> from transformers import AutoTokenizer, AutoModelForCausalLM
>>> tokenizer = AutoTokenizer.from_pretrained("EleutherAI/gpt-neox-20b")
>>> tokenizer.tokenize("ableable")
['able', 'able']
```

```
# ggml
main: number of tokens in prompt = 5
main: token[0] =  21762, able
main: token[1] =     68, a
main: token[2] =     69, b
main: token[3] =     79, l
main: token[4] =     72, e
```

Example above is due to this line https://github.com/ggerganov/ggml/blob/2a75bd43113702fe096c75408b300d106be90a58/examples/common.cpp#L269

Furthermore, current implementation does not consider double-byte characters such as Korean, Japanese, Chinese. I'm wondering whether it would be ideal to modify the gpt_tokenize function to be able to handle double-byte characters, or the repo aims on single character languages only as English and most roman languages are single-byte.

klosax:
I found a quick fix:

Replace line:
https://github.com/ggerganov/ggml/blob/2a75bd43113702fe096c75408b300d106be90a58/examples/common.cpp#L262

with:
```
j = n;
continue;
```
Perplexity tests:

model = mpt-7b-base-ggml-f16.bin
ctx = 512
batch size = 512
prompt = wiki.test.raw

Without this fix:
tokens in prompt = 326191
perplexity of first chunk = 30.60130927

With this fix:
tokens in prompt = 306246
perplexity of first chunk = 11.00999002

ggerganov:
@klosax Thanks! Btw, would be nice to add a simple perplexity tool as a `ggml` example

@jaeminSon Do you still observe issues with latest `master`?

klosax:
> @klosax Thanks! Btw, would be nice to add a simple perplexity tool as a `ggml` example

I am currently working on a perplexity tool. The tool should be able to measure perplexity correctly on wiki.test.raw, but it seems that the gpt tokenizer does not work correctly with unicode characters.

jaeminSon:
@ggerganov The current version fixes the above issue! LoL

It would be great if perplexities are compared between huggingface model and the converted ggml model!!!

[...]

PR #179, opened 2023-05-20
Add utf-8 support: gpt_tokenize / mpt model import

klosax:
This should fix issue #170 :

- Adds utf-8 support to gpt_tokenize
- Fixes a bug in gpt_tokenize
- Fixes mpt model import script to correctly import utf-8 tokens

Perplexity tests:

model = mpt-7b-base-ggml-f16.bin
ctx = 512
batch size = 512
prompt = wiki.test.raw

Without this pr:
tokens in prompt = 326191
perplexity of first chunk = 30.60130927

With this pr:
tokens in prompt = 286267
perplexity of first chunk = 8.55602292

### DIFF
--- a/examples/common.cpp
+++ b/examples/common.cpp
@@ -212,7 +212,7 @@ void gpt_vocab::add_special_token(const std::string & token) {
     special_tokens.push_back(token);
 }
 
-void append_utf8(char32_t ch, std::string & out) {
+static void append_utf8(char32_t ch, std::string & out) {
     if (ch <= 0x7F) {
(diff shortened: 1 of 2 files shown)


klosax:
> Hi @ggerganov @klosax , I wanted to say that this commit breaks tokenization for non-English languages.

Pr #184 should fix this.

ggerganov:
@LostRuins Is it working now as expected?

LostRuins:
Hi @ggerganov it is working well and coherent. Thanks.

Issue #55, opened 2023-03-31
Repetition penalty?


lxe:
Is there a way to add/use repetition penalty to the generator code?

dchest:
You can grab this piece from llama.cpp:

https://github.com/ggerganov/llama.cpp/blob/master/llama.cpp#L1200-L1216

PR #184, opened 2023-05-21
mpt: utf-8 support, perplexity testing, repeat penalty sampling

klosax:
Follow up to pr #179 .  Fixes issues #170 , #55 . 

A better solution that will not affect the other examples that uses gpt_tokenize

- Revert encoding of input to utf-8 in gpt_tokenize  
- Add decoding of utf-8 tokens on load in mpt main.
- Small change to mpt import script
- Add perplexity testing
- Add repeat penalty sampling


examples/common.cpp
```diff
@@ -212,38 +212,40 @@ void gpt_vocab::add_special_token(const std::string & token) {
...
+			out.push_back( ((u8[i] & 0xf) << 12) | ((u8[i+1] & 0x3f) << 6) | (u8[i+2] & 0x3f) );
+		} else if (tmp < 0xf8UL) {
+			elem_len = 4;
+			out.push_back( ((u8[i] & 0x7) << 18) | ((u8[i+1] & 0x3f) << 12)	| ((u8[i+2] & 0x3f) << 6) | (u8[i+3] & 0x3f) );
+		} else {
+		
+            		printf("Invalid Unicode code point\n");
+            		break;
+		}
+	}
```
ggerganov, line 240: Fix the indentation in this function

klosax: good?

zenixls2:
This matches the implementation in Transformer and should be back-ported to llama.cpp's examples/main and common.h/cpp.

examples/mpt/main.cpp
```diff
@@ -64,6 +64,105 @@ struct mpt_model {
     std::map<std::string, struct ggml_tensor *> tensors;
 };
 
+struct mpt_params {
```
zenixls2, line 67: since this is in cpp, is it possible to extend the `struct gpt_params`?

examples/common.cpp
```diff
@@ -407,6 +408,129 @@ gpt_vocab::id gpt_sample_top_k_top_p(
...
+            if (plogits[i] > max_logit) {
+                max_logit = plogits[i];
+                max_id = i;
+            }
+        }
+        return max_id;
+    }
+
+
+    std::vector<std::pair<double, gpt_vocab::id>> logits_id;
```
zenixls2, line 445: logit is float. Maybe use `std::pair<float, gpt_vocab::id>`?

zenixls2:
also in llama.cpp/examples/main/main.cpp line 423
and in llama.cpp/llama.cpp line 1833

implements similar approach that makes use of the `sorted` flag to speed-up.

examples/mpt/perplexity.cpp
```diff
@@ -0,0 +1,809 @@
...
+        printf("%s:     load time = %8.2f ms\n", __func__, t_load_us / 1000.0f);
+        printf("%s:  eval time = %8.2f ms / %.2f ms per token\n", __func__, t_predict_us / 1000.0f,
+               t_predict_us / 1000.0f / (n_chunk * params.n_ctx) );
+        printf("%s:    total time = %8.2f ms\n", __func__, (t_main_end_us - t_main_start_us) / 1000.0f);
+    }
+
+    ggml_free(model.ctx);
+
+    return 0;
+}
```
ggerganov, line 809: No need to duplicate `main.cpp` in `perplexity.cpp`.
Just add a `--perplexity` CLI argument to `main.cpp`

ggerganov (changes requested):

examples/mpt/perplexity.cpp
klosax: Done.

ggerganov (approved):

PR #186, opened 2023-05-22
examples : add tokenization tests and refactor codes

### LINKED ISSUE #170: gpt_tokenize : bug in tokenization and incapable of double-byte languages

jaeminSon:
(An attempt to fix #170)

examples : test tokenization + support utf-8 and multi-byte encodings + refactor gpt_tokenize + refactor convert-h5-to-ggml.py


examples/common.cpp
```diff
@@ -212,34 +222,22 @@ void gpt_vocab::add_special_token(const std::string & token) {
...
 std::vector<gpt_vocab::id> gpt_tokenize(const gpt_vocab & vocab, const std::string & text) {
     std::vector<std::string> words;
 
     // Convert input to utf-8
-    std::string utf8conv;
-    for (int w = 0; w < text.size(); w++) {
-        append_utf8( uint8_t(text[w]), utf8conv);
-    }
+    std::wstring text_multibytes = convert_to_wstring(text);
+    std::string utf8conv = convert_to_utf8(text_multibytes);
```
klosax, line 240: I reverted this in pr #184 as it breaks other models, see comment in pr #179 . I think each model will have to handle the encoding of its tokens.

examples/common.cpp
```diff
@@ -8,6 +8,16 @@
...
+#include <locale>
+#include <codecvt>
+#include <sstream>
+#include <filesystem>
+
+#ifdef __APPLE__
+namespace fs = std::__fs::filesystem;
+#else
+namespace fs = std::filesystem;
+#endif
```
ggerganov, line 20: Do not use `#include <filesystem>`. Use `<fstream>`

examples/common.cpp
```diff
@@ -271,43 +255,136 @@ std::vector<gpt_vocab::id> gpt_tokenize(const gpt_vocab & vocab, const std::stri
...
+
     return tokens;
 }
 
+bool are_strings_equal(const std::vector<std::string>& vec1, const std::vector<std::string>& vec2) {
+    if (vec1.size() != vec2.size()) return false;
+
+    for (size_t i = 0; i < vec1.size(); ++i) 
+        if (vec1[i] != vec2[i]) return false;
+
```
ggerganov, line 287:
```suggestion
    for (size_t i = 0; i < vec1.size(); ++i) {
        if (vec1[i] != vec2[i]) return false;
    }

```

Fix this in other places too

ggerganov (changes requested):

examples/prompts/tokenize_huggingface.py
```diff
@@ -0,0 +1,65 @@
...
+                "openai/whisper-tiny"  # whisper (base, large, large-v2 share the same tokenizer)
+                ]
+
+repo2ggml = {"databricks/dolly-v2-3b": "dolly-v2",
+             "gpt2": "gpt-2",
+             "uer/gpt2-chinese-cluecorpussmall": "gpt-2-chinese",
+             "EleutherAI/gpt-j-6b": "gpt-j",
+             "EleutherAI/gpt-neox-20b": "gpt-neox",
+             "EleutherAI/polyglot-ko-1.3b": "polyglot-ko",
+             "rinna/japanese-gpt-neox-3.6b": "japanese-gpt-neox",
```
ggerganov, line 25:
```suggestion
             "rinna/japanese-gpt-neox-3.6b": "gpt-neox-japanese",
```

For consistency in the file names, always put the language as suffix

examples/common.cpp
```diff
@@ -271,43 +251,149 @@ std::vector<gpt_vocab::id> gpt_tokenize(const gpt_vocab & vocab, const std::stri
...
+                if (fname.find(t_name) != std::string::npos) {
+                    return dir_test + "/" + t_fname;
+                }
+            }
+        }
+        closedir(dir);
+    }
+    
+    return ""; // empty string if test file not found
+}
```
ggerganov, line 330: I think this entire function is not needed.
Instead, let's explicitly provide the path to the test file in each example

ggerganov (approved):
Great job 👍 

After resolving the merge conflicts, we can merge it

PR #190, opened 2023-05-23
Add web page for the MNIST example

rgerganov:
The web page is using WASM for model inference.
Users can draw digits on an HTML canvas and load random digits from the MNIST dataset.

### DIFF
--- a/examples/mnist/main.cpp
+++ b/examples/mnist/main.cpp
@@ -198,6 +198,41 @@ int mnist_eval(
     return prediction;
 }
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int wasm_eval(uint8_t *digitPtr)
+{
+    mnist_model model;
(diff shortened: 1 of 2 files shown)


rgerganov:
Right now I build this with:
```
cd examples/mnist
emcc -I../../include -I../../include/ggml -I../../examples ../../src/ggml.c main.cpp -o web/mnist.js -s EXPORTED_FUNCTIONS='["_wasm_eval","_wasm_random_digit","_malloc","_free"]' -s EXPORTED_RUNTIME_METHODS='["ccall"]' -s ALLOW_MEMORY_GROWTH=1 --preload-file models/mnist
```

I need to figure out how to integrate it in the CMake build

ggerganov (approved):

matthoffner:
@rgerganov thanks for your example I was able to get it to build with the starcoder example: https://huggingface.co/spaces/matthoffner/ggml-maker/blob/main/examples/wasm-starcoder.Dockerfile

PR #199, opened 2023-05-26
mnist : smooth user input

rgerganov:
Drawing on the canvas is now smooth. The final image which is used for prediction is obtained by down-scaling the canvas to 28x28 pixels. Download button is aslo added for downloading raw image values.


ggerganov (approved):

ggerganov:
Should probably avoid re-loading the model for every prediction.
Maybe add `wasm_init()` or something

PR #204, opened 2023-05-28
Fix ggml_alibi tensor size

tye-singwa:
hi! i've debugged mpt-7b implementation, and encounter assertion failing:
- here https://github.com/ggerganov/ggml/blob/07bd0114365c7cb0cbb2ad8297d7a64788f882c4/src/ggml.c#L10833
- and here https://github.com/ggerganov/ggml/blob/07bd0114365c7cb0cbb2ad8297d7a64788f882c4/src/ggml.c#L10897

this patch should fix it, 
thanks

cc https://github.com/rustformers/llm/issues/281

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -6317,7 +6317,7 @@ struct ggml_tensor * ggml_alibi(
 
     ggml_scratch_save(ctx);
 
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Good catch!

troyedwardsjr:
@tye-singwa Thanks for this fix!

PR #207, opened 2023-05-29
gpt-neox : utf-8 support

klosax:
Fixes most of tokenization errors found using test implemented in pr #186 .
Same as implemented for mpt in pr #184 .

- main : Add decoding of utf-8 tokens.
- import script : Add encoding of utf-8 tokens.


klosax:
Closing this..What is needed is a good BPE-tokenizer implementation. See https://github.com/ggerganov/ggml/issues/220#issuecomment-1571083572 .

Issue #216, opened 2023-05-30
cmake -DGGML_CLBLAST=ON .. ERROR


vicwer:
```
git clone https://github.com/CNugteren/CLBlast.git
cd CLBlast/ && mkdir build && cd build/ && cmake .. && make -j8 && make install
cd ../../ggml/build
cmake -DGGML_CLBLAST=ON ..
make
```
ERROR:
```
Scanning dependencies of target ggml
[  1%] Building C object src/CMakeFiles/ggml.dir/ggml.c.o
[  3%] Building C object src/CMakeFiles/ggml.dir/ggml-opencl.c.o
/root/ggml/src/ggml-opencl.c: In function ‘ggml_cl_init’:
/root/ggml/src/ggml-opencl.c:214:5: warning: ‘clCreateCommandQueue’ is deprecated [-Wdeprecated-declarations]
  214 |     queue = clCreateCommandQueue(context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
      |     ^~~~~
In file included from /usr/include/CL/opencl.h:47,
                 from /usr/local/include/clblast_c.h:22,
                 from /root/ggml/src/ggml-opencl.c:4:
/usr/include/CL/cl.h:1443:1: note: declared here
 1443 | clCreateCommandQueue(cl_context                     /* context */,
      | ^~~~~~~~~~~~~~~~~~~~
/root/ggml/src/ggml-opencl.c: At top level:
/root/ggml/src/ggml-opencl.c:251:20: warning: ISO C forbids forward references to ‘enum’ types [-Wpedantic]
  251 |         const enum ggml_blas_order order, const enum ggml_blas_op trans_a, const enum ggml_blas_op trans_b,
      |                    ^~~~~~~~~~~~~~~
/root/ggml/src/ggml-opencl.c:251:54: warning: ISO C forbids forward references to ‘enum’ types [-Wpedantic]
  251 |         const enum ggml_blas_order order, const enum ggml_blas_op trans_a, const enum ggml_blas_op trans_b,
      |                                                      ^~~~~~~~~~~~
/root/ggml/src/ggml-opencl.c:251:87: warning: ISO C forbids forward references to ‘enum’ types [-Wpedantic]
  251 |         const enum ggml_blas_order order, const enum ggml_blas_op trans_a, const enum ggml_blas_op trans_b,
      |                                                                                       ^~~~~~~~~~~~
[...]

Issue #205, opened 2023-05-28
clBLAST Compile fails due to recent changes.


apcameron:
<PRE>
[  1%] Building C object src/CMakeFiles/ggml.dir/ggml.c.o
/home/andrew/Downloads/ggml/src/ggml.c: In function ‘quantize_row_q8_0’:
/home/andrew/Downloads/ggml/src/ggml.c:1026:15: warning: unused variable ‘nb’ [-Wunused-variable]
 1026 |     const int nb = k / QK8_0;
      |               ^~
/home/andrew/Downloads/ggml/src/ggml.c: In function ‘quantize_row_q8_1’:
/home/andrew/Downloads/ggml/src/ggml.c:1221:15: warning: unused variable ‘nb’ [-Wunused-variable]
 1221 |     const int nb = k / QK8_1;
      |               ^~
In file included from /usr/include/string.h:535,
                 from /home/andrew/Downloads/ggml/src/ggml.c:17:
In function ‘strncpy’,
    inlined from ‘ggml_set_name’ at /home/andrew/Downloads/ggml/src/ggml.c:4509:5,
    inlined from ‘ggml_diag_mask_zero_impl’ at /home/andrew/Downloads/ggml/src/ggml.c:6144:5:
/usr/include/bits/string_fortified.h:95:10: warning: ‘__builtin_strncpy’ offset [0, 30] is out of the bounds [0, 0] [-Warray-bounds=]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘strncpy’,
    inlined from ‘ggml_set_name’ at /home/andrew/Downloads/ggml/src/ggml.c:4509:5,
    inlined from ‘ggml_diag_mask_zero_impl’ at /home/andrew/Downloads/ggml/src/ggml.c:6144:5,
    inlined from ‘ggml_diag_mask_zero’ at /home/andrew/Downloads/ggml/src/ggml.c:6163:12:
/usr/include/bits/string_fortified.h:95:10: warning: ‘__builtin_strncpy’ offset [0, 30] is out of the bounds [0, 0] [-Warray-bounds=]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
[...]

PR #222, opened 2023-05-31
FIX CLBLAST Compile

apcameron:
Copy ggml-opencl.cpp from llama.cpp
Update CMakefiles.txt

This addresses the following https://github.com/ggerganov/ggml/issues/216 and https://github.com/ggerganov/ggml/issues/205


apcameron:
@ggerganov Do I need to make any more updates or is this PR good to go?

ggerganov:
I'll sync latest `llama.cpp` together with OpenCL, CUDA and Metal implementations later today

Issue #225, opened 2023-06-03
GPTNeoX model with 16k context results in context-size related issues: `ggml_new_tensor_impl: not enough space in the context's memory pool`, and instant core dump with fp16


TheBloke:
Hey guys

Today I was doing quants of a [new GPTNeoX model called Literature-7B-16384](https://huggingface.co/hakurei/Literature-7B-16384)

I tried making GGMLs through the usual process:
```
python examples/gpt-neox/convert-h5-to-ggml.py /workspace/models/hakurei_Literature-7B-16384 0
build/bin/gpt-neox-quantize /workspace/process/literature-7b/ggml/ggml-model-f32.bin /workspace/process/literature-7b/ggml/literature-7b-16384.gptneox.ggmlv3.q4_0.bin q4_0
```

Both steps completed fine.  But the models can't be used.

Trying to use the fp32:
```
[pytorch2] ubuntu@h100:/workspace/git/ggml git:(master) $ build/bin/gpt-neox -m /workspace/process/literature-7b/ggml/ggml-model-f32.bin   -p "test"
main: seed = 1685827980
gpt_neox_model_load: loading model from '/workspace/process/literature-7b/ggml/ggml-model-f32.bin' - please wait ...
gpt_neox_model_load: n_vocab = 50432
gpt_neox_model_load: n_ctx   = 16384
gpt_neox_model_load: n_embd  = 4096
gpt_neox_model_load: n_head  = 32
gpt_neox_model_load: n_layer = 32
gpt_neox_model_load: n_rot   = 128
gpt_neox_model_load: par_res = 0
gpt_neox_model_load: ftype   = 0
gpt_neox_model_load: qntvr   = 0
gpt_neox_model_load: ggml ctx size = 11822.28 MB
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 12661385472, available 12396563456)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 12661451264, available 12396563456)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 12594392320, available 12396563456)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 12460224000, available 12396563456)
.... lots of similar lines removed ...
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 16691523072, available 12396563456)
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 16691523072, available 12396563456)
[...]

PR #228, opened 2023-06-04
gpt-neox : fix ctx size calculation

klosax:
Fixes issue #225 .

- Corrected calculation of ctx_size.
- Inceased ctx overhead to support 16k ctx using f32 weights.

### DIFF
--- a/examples/gpt-neox/main.cpp
+++ b/examples/gpt-neox/main.cpp
@@ -156,10 +156,10 @@ bool gpt_neox_model_load(const std::string & fname, gpt_neox_model & model, gpt_
     {
         const auto & hparams = model.hparams;
 
-        const int n_embd  = hparams.n_embd;
-        const int n_layer = hparams.n_layer;
-        const int n_ctx   = hparams.n_ctx;
-        const int n_vocab = hparams.n_vocab;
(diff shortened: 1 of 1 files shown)


TheBloke:
Thanks so much for the fast fix! I will test it with the 16k Literature model shortly

TheBloke:
It works! Thanks so much for the quick fix.

As soon as this is merged I will release the 16K GGMLs.

examples/gpt-neox/main.cpp
```diff
@@ -190,7 +190,7 @@ bool gpt_neox_model_load(const std::string & fname, gpt_neox_model & model, gpt_
         ctx_size += n_ctx*n_layer*n_embd*ggml_type_sizef(GGML_TYPE_F32); // memory_k
         ctx_size += n_ctx*n_layer*n_embd*ggml_type_sizef(GGML_TYPE_F32); // memory_v
 
-        ctx_size += (6 + 16*n_layer)*512; // object overhead
+        ctx_size += (6 + 16*n_layer)*1024; // object overhead
```
ggerganov, line 193: You can now use `ggml_tensor_overhead()` instead of `1024` constant

ggerganov (approved):

Issue #217, opened 2023-05-30
support for falcon model


aseok:
Is it possible to convert falcon model by other example conversion codes?

klosax:
The architecture that Falcon uses is different than those currently supported.

More discussion here: https://github.com/ggerganov/llama.cpp/issues/1602

klosax:
Falcon LLM ggml framework with CPU and GPU support:
https://github.com/cmp-nct/ggllm.cpp

Issue #220, opened 2023-05-31
ggml : unified file format


philpax:
Obsoletes #147, #150, https://github.com/ggerganov/llama.cpp/issues/1575, https://github.com/ggerganov/llama.cpp/issues/1590, https://github.com/rustformers/llm/discussions/143, and probably some other issues across some other repositories.

Please see the spec PR at #302; the following is left as-is so you can see the original proposal.

---

# Current state of affairs

## Overview

At present, there are two GGML file formats floating around for LLMs (and potentially other ggml-using projects, I haven't looked too much at the implementation of whisper):
- GGML unversioned
- GGJTv3 (same as v1 and v2, but with different quantization formats), which is similar to GGML but includes a version and aligns the tensors to allow for memory-mapping

Both of these formats share the same fundamental structure:

- a magic number with an optional version number
- model-specific hyperparameters that include a `ftype` that should describe the type of the majority of the tensors, and for GGML files, the quantization version encoded using a modulo in the ftype
- an embedded vocabulary, which is a list of strings with length prepended. The GGMF/GGJT formats embed a f32 score next to the strings
- finally, a list of tensors with their length-prepended name, type, and (aligned, in the case of GGJT) tensor data

We have more details on the format here: <https://github.com/rustformers/llm/tree/main/crates/ggml#format>

## Drawbacks

Unfortunately, over the last few months, there are a few issues that have become apparent with the existing models:

- There's no way to identify which model architecture a given model is for, because that information isn't present
  - Similarly, existing programs cannot intelligently fail upon encountering new architectures
- Adding or removing any new hyperparameters is a breaking change, which is impossible for a reader to detect without [herculean hacks](https://github.com/LostRuins/koboldcpp/blob/concedo/model_adapter.cpp#L178)
- Each model architecture requires its own conversion script to their architecture's variant of GGML
- Maintaining backwards compatibility without breaking the structure of the format requires clever tricks, like packing the quantization version into the ftype, which are not guaranteed to be picked up by readers/writers, and are not consistent between the two formats

# GGJTv4/GGUF

Based on this, I'd like to propose a new format that's designed to be universal and addresses these issues. It is largely identical to GGJTv3, but makes one important difference: the hyperparameters are encoded as an array of key-value pairs that can be read in any order, and these hyperparameters are used to encode additional information about the model. A really important property I'd like to keep is single-file deployment: if I give you a GGUF file and you have a compatible executor, it should Just Work:tm without any additional conversion or extra files.

[...]

PR #231, opened 2023-06-06
add falcon7b example

apage43:
https://github.com/ggerganov/ggml/issues/217
adapted from gpt-neox example and work started in https://github.com/ggerganov/llama.cpp/issues/1602

only supports 7b right now - 40b multiquery attention gets hairier, as its 128 query heads with 8 k and v heads, as opposed to 7B's 71 query heads with 1 k and v head


ggerganov:
Cool - will take a look soon
Is this using actual MQA or it's still doing the trick with the copies?

cmp-nct:
Are you working on a 40B branch already ?

apage43:
> Are you working on a 40B branch already ?

I'm not presently - being that it's *big* its a bit more inconvenient to hack on as I need to use a bigger machine than I'm usually on for dev stuff.

cmp-nct:
> > Are you working on a 40B branch already ?
> 
> I'm not presently - being that it's _big_ its a bit more inconvenient to hack on as I'd need to use a bigger machine than I'm usually on for dev stuff.

Did you see https://huggingface.co/jploski/falcon40b-mini-shakespeare ?

apage43:
> > > Are you working on a 40B branch already ?
> > 
> > 
> > I'm not presently - being that it's _big_ its a bit more inconvenient to hack on as I'd need to use a bigger machine than I'm usually on for dev stuff.
> 
> Did you see https://huggingface.co/jploski/falcon40b-mini-shakespeare ?

I have now :) I still probably won't get to it soon - but if someone figures out how to support that before this lands I'm happy to incorporate it

jploski:
I did some work regarding 40B support today: https://github.com/ggerganov/ggml/commit/27cf1adc7362e9179bc7e70668098b7d74c79f95

After making my head nearly explode several times I reached a point where it generates okay sounding prose from the falcon40b-mini-shakespeare model, but it does not match the Python version output exactly as it should (and as it does for the 7B version).

The main obstacle seems to be that I am unable to make ggml_repeat broadcast multiple keys like the "k = torch.broadcast_to(k, q.shape)" in Python does (I get  "1,2,1,2" instead of "1,1,2,2" so to say).

Another big problem is that the I only got the query matrix to look like the original Python one through some brute force offset calculations and copying of subvectors. It probably won't scale at all. I'm under impression that what needs to be done there can't be done using just reshaping or view operations. The memory format (as stored in Python and written by the conversion script) seems to be very difficult to work with in GGML.

Or maybe I'm just too inexperienced in this tensor wrestling... Once again giving up in hope that someone with more clue can pick it up.

jploski:
> I did some work regarding 40B support today: [27cf1ad](https://github.com/ggerganov/ggml/commit/27cf1adc7362e9179bc7e70668098b7d74c79f95)

As a further explanation of the code and where the complexity comes from here's a visualization of the fused_kqv weights format (from  falcon40b-mini-shakespeare config): https://docs.google.com/spreadsheets/d/1FoM6pIUj23GMW4zO_G1hjmEnUacBxBKN/edit?usp=sharing&ouid=111096390735143611797&rtpof=true&sd=true

KerfuffleV2:
Maybe just make your own repeat operation that works the way you need? Seems like the repeat op is only implemented for float32 so there's just one version of the function required.

You could create a new op and just cut-and-paste the existing `_repeat` functions: https://github.com/ggerganov/ggml/blob/f52d2a05cf8327baf6c0d49e7b231953179e03d3/src/ggml.c#L8773

The function looks relatively simple also.

jploski:
> Maybe just make your own repeat operation that works the way you need? Seems like the repeat op is only implemented for float32 so there's just one version of the function required.

I added a new ggml_repeat2 function as suggested (https://github.com/ggerganov/ggml/commit/3352043d851fbc84a46e251c3281d24bd18efeb2) - although the original ggml_repeat also has a backwards pass and I'm not sure if it's the same for what I added.

With some more tweaks (commited in https://github.com/ggerganov/ggml/commit/3bc786b4a9cc275158d613b381754debdb41cf33) I now have a version which works with all falcon-mini-shakespeare models I have unleashed upon this world (both 7B and 40B configs). At least in 32bit, haven't tested quantized yet.  The (known) remaining problem is the for-loop-based splitting of query heads. I suspect it's gonna blow up with a real big model, either being slow or exceeding the max number of tensors (4096) allowed by GGML (or both).

(Also it's possible that the implementation does some unnecessary operations like permutes or 4d instead of 3d, but that's minor.)

```
bin/falcon -m /mnt/seagate/miniconda3/falcon40b/falcon40b-mini-shakespeare/ggml-model--f32.bin --top_p 1 --top_k 1 -s 42 -p "When we loop"

When we loop, and for his head,
And in his head's head's face,
And yet with his head's head is to him;
And now, in this land's face,
And with his head by his head he will die.
```
I tend to agree, tha'ts almost what happened to me.

maccam912:
> With some more tweaks (commited in [3bc786b](https://github.com/ggerganov/ggml/commit/3bc786b4a9cc275158d613b381754debdb41cf33)) I...
Ok I've been too afraid to ask, but how on earth are you doing these commits that aren't on any branch at all? I wanted to clone the repo and check out the commit but I have no idea how to.

jploski:
> > With some more tweaks (commited in [3bc786b](https://github.com/ggerganov/ggml/commit/3bc786b4a9cc275158d613b381754debdb41cf33)) I...
> 
> Ok I've been too afraid to ask, but how on earth are you doing these commits that aren't on any branch at all? I wanted to clone the repo and check out the commit but I have no idea how to.

Sorry for the confusions - these commits belong to branch falcon40b of my fork: https://github.com/jploski/ggml/tree/falcon40b - apparently GitHub not clever enough to indicate their source.

KerfuffleV2:
@jploski 

I was able to convert the real 40B model with my change here to reduce memory during HF conversion (only loads a single part into RAM at a time): https://github.com/jploski/ggml/pull/1

It required some work to get inference to actually run. I had to increase `ctx_size`:

    ctx_size += ((size_t)3) * 1024 * 1024 * 1024;

Also, uhh... `GGML_MAX_NODES` at 4096 didn't quite cut it. Nor did 65535, I eventually just set it to 262144 and was able to run the model. Unfortunately, the output didn't make much sense:

```plaintext
main: seed = 1686733539
falcon_model_load: loading model from '/home/nope/personal/ai/models/falc40b.ggml' - please wait ...
falcon_model_load: n_vocab   = 65024
falcon_model_load: n_embd    = 8192
falcon_model_load: n_head    = 128
falcon_model_load: n_head_kv = 8
falcon_model_load: n_layer   = 60
falcon_model_load: ftype     = 2008
falcon_model_load: qntvr     = 2
falcon_model_load: ggml ctx size = 28175.96 MB
falcon_model_load: memory_size =   480.00 MB, n_mem = 122880
falcon_model_load: ............................................................ done
falcon_model_load: model size = 27436.06 MB / num tensors = 484
extract_tests_from_file : No test file found.
test_gpt_tokenizer : 0 tests failed out of 0 tests.
main: number of tokens in prompt = 10
main: token[0] =   7107, Once
main: token[1] =   2918,  upon
main: token[2] =    241,  a
main: token[3] =    601,  time
main: token[4] =     23, ,
main: token[5] =    629,  there
main: token[6] =    398,  was
main: token[7] =    241,  a
main: token[8] =   1278,  little
main: token[9] =  27224,  fox

Once upon a time, there was a little fox and, I’re
'  to’ ' .
 it that,. is
,, of . for.' '- you,. we the- the
1 of a
. the
```

Although it didn't work, even with the crazy number of nodes it wasn't really that slow. It was about the same as a 65B Q4_K_M LLaMA model with llama.cpp.

The mini-Shakespeare model seems fine:

```plaintext
main: seed = 1686733831
falcon_model_load: loading model from '/home/nope/personal/ai/models/falcsp.ggml' - please wait ...
falcon_model_load: n_vocab   = 65024
falcon_model_load: n_embd    = 256
falcon_model_load: n_head    = 4
falcon_model_load: n_head_kv = 2
falcon_model_load: n_layer   = 4
falcon_model_load: ftype     = 2009
falcon_model_load: qntvr     = 2
falcon_model_load: ggml ctx size = 3105.91 MB
falcon_model_load: memory_size =     8.00 MB, n_mem = 8192
falcon_model_load: .... done
falcon_model_load: model size =    25.89 MB / num tensors = 36
extract_tests_from_file : No test file found.
test_gpt_tokenizer : 0 tests failed out of 0 tests.
main: number of tokens in prompt = 1
main: token[0] =   4031, Now

Now, Clarence, my lord, I am a
the great men: I will to do this day you
In time they may live in men of tears are
Shall be not what we have fought in. What is this
and come to you? I have not made mine eyes,
Which now sent for, or I am so fast?
Your friends shall be revenged on thee, hoar!
And that you must, sirs, that you must do,
My friend to thee that news, with your love,
My father's wife and love for this day.
You are not hot, lords, and what I am not?
To take this, good sweet friend, I am not my life,
I warrant, as I, to have a little thing, my lord,
What you can stay with this good night do you all your tongue?
O, if not my fair soul to my brother, how well,
Where is

main: mem per token =   290292 bytes
main:     load time =   266.82 ms
main:   sample time =    64.16 ms
main:  predict time =   240.96 ms / 1.20 ms per token
main:    total time =   576.08 ms
```

Both models were quantized to Q5_0.

jploski:
> @jploski
> 
> I was able to convert the real 40B model with my change here to reduce memory during HF conversion (only loads a single part into RAM at a time): [jploski#1](https://github.com/jploski/ggml/pull/1)
> 
> It required some work to get inference to actually run. I had to increase `ctx_size`:
> 
> ```
> ctx_size += ((size_t)3) * 1024 * 1024 * 1024;
> ```
> 
> Also, uhh... `GGML_MAX_NODES` at 4096 didn't quite cut it. Nor did 65535, I eventually just set it to 262144 and was able to run the model. Unfortunately, the output didn't make much sense:

Thanks for checking! I was able to reproduce wrong output using an unquantized mini version trained with n_embd = 1024, n_head = 128, n_head = 8. So there must still be a bug somewhere, which the previous three configs I used for testing did not catch.

KerfuffleV2:
If the problem is the complicated logic for dealing for the query heads, maybe the easiest way to deal with that is in the conversion tool from the Torch or numpy side. It should be relatively easy to shuffle things around at that point.

Reducing the complexity would make issues easier to debug too, I guess.

jploski:
> If the problem is the complicated logic for dealing for the query heads, maybe the easiest way to deal with that is in the conversion tool from the Torch or numpy side. It should be relatively easy to shuffle things around at that point.
> 
> Reducing the complexity would make issues easier to debug too, I guess.

Yes, I agree that reshuffling the weights during conversion will perhaps be the final and most elegant/efficient solution. I just haven't wrapped my head around it yet how changing the layout of the query_key_value tensor maps into fused_qkv from which the qkv vectors are extracted (fused_qkv = self.query_key_value(hidden_states)).

I'd also like to understand the current bug and have a working (if poorly implemented) version to improve on (even if the "improvement" will mean throwing away the overcomplicated code).

jploski:
> I'd also like to understand the current bug and have a working (if poorly implemented) version to improve on (even if the "improvement" will mean throwing away the overcomplicated code).

Understood and fixed in my falcon40b branch. Please recompile and try again.

KerfuffleV2:
It's alliiiiive!

```plaintext
main: seed = 1686742967
falcon_model_load: loading model from '/home/nope/personal/ai/models/falc40b.ggml' - please wait ...
falcon_model_load: n_vocab   = 65024
falcon_model_load: n_embd    = 8192
falcon_model_load: n_head    = 128
falcon_model_load: n_head_kv = 8
falcon_model_load: n_layer   = 60
falcon_model_load: ftype     = 2008
falcon_model_load: qntvr     = 2
falcon_model_load: ggml ctx size = 28175.96 MB
falcon_model_load: memory_size =   480.00 MB, n_mem = 122880
falcon_model_load: ............................................................ done
falcon_model_load: model size = 27436.06 MB / num tensors = 484
extract_tests_from_file : No test file found.
test_gpt_tokenizer : 0 tests failed out of 0 tests.
main: number of tokens in prompt = 10
main: token[0] =   7107, Once
main: token[1] =   2918,  upon
main: token[2] =    241,  a
main: token[3] =    601,  time
main: token[4] =     23, ,
main: token[5] =    629,  there
main: token[6] =    398,  was
main: token[7] =    241,  a
main: token[8] =   1278,  little
main: token[9] =  27224,  fox

Once upon a time, there was a little fox named ‘Pee-Poo’ who had an important mission to accomplish.
She had been assigned the task of finding the ‘Guru of all Gurus’ who was hiding deep in the jungle. And so one day, Pee-Poo set out for her journey. She walked and walked and walked and asked everybody in the jungle where the Guru lived, but nobody could tell her.
“But, how can that be?” she thought to herself, “There has

main: mem per token =  6467732 bytes
main:     load time = 10538.50 ms
main:   sample time =    34.28 ms
main:  predict time = 90867.40 ms / 833.65 ms per token
main:    total time = 104610.47 ms
```

Not a fan of the name it chose though.

For reference, these are the changes I need to actually run it:

```patch
diff --git a/examples/falcon/main.cpp b/examples/falcon/main.cpp
index beac293..c77c610 100644
--- a/examples/falcon/main.cpp
+++ b/examples/falcon/main.cpp
@@ -198,6 +198,7 @@ bool falcon_model_load(const std::string & fname, falcon_model & model, gpt_voca
                     ggml_type_sizef(GGML_TYPE_F32);  // memory_v
 
         ctx_size += (5 + 10 * n_layer) * 256;  // object overhead TODO:
+        ctx_size += ((size_t)3) * 1024 * 1024 * 1024;
         printf("%s: ggml ctx size = %6.2f MB\n", __func__, ctx_size/(1024.0*1024.0));
     }
 
diff --git a/include/ggml/ggml.h b/include/ggml/ggml.h
index e770603..83b0d84 100644
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -194,7 +194,7 @@
 #define GGML_QNT_VERSION_FACTOR 1000 // do not change this
 
 #define GGML_MAX_DIMS          4
-#define GGML_MAX_NODES         4096
+#define GGML_MAX_NODES         262144
 #define GGML_MAX_PARAMS        256
 #define GGML_MAX_CONTEXTS      64
 #define GGML_MAX_OPT           4
```

TheBloke:
Amazing work guys!

So is https://github.com/jploski/ggml/tree/falcon40b the branch I should use to try converting and running GGMLs?

TheBloke:
OK fair enough!

jploski:
> > Amazing work guys!
> > So is https://github.com/jploski/ggml/tree/falcon40b the branch I should use to try converting and running GGMLs?
> 
> I would suggest not converting them just yet - because if/when the qkv reshuffling during conversion is implemented, the binary format of the tensors would change again... which would make all the already published files incompatible.

The reshuffling is now implemented in https://github.com/jploski/ggml/tree/falcon40b - works with all the mini models, but I did not test it with the actual 40B model. Please try it out.

ochafik:
> The reshuffling is now implemented in https://github.com/jploski/ggml/tree/falcon40b - works with all the mini models, but I did not test it with the actual 40B model. Please try it out.

Inference fails fast at 5ee0488ee6ece2cb09cd28615ea934d641115990 ([repro and full output](https://gist.github.com/ochafik/294a8fe7f80724f47d8526b124a29fa8), freshly reconverted and quantized) on my M2 Max 96GB with repeated errors & a segfault:

`ggml_new_tensor_impl: not enough space in the context's memory pool (needed 21670576384, available 21631032576)`

jploski:
> > The reshuffling is now implemented in https://github.com/jploski/ggml/tree/falcon40b - works with all the mini models, but I did not test it with the actual 40B model. Please try it out.
> 
> Inference fails fast

Can you try if the ctx_size fix from https://github.com/ggerganov/ggml/pull/231#issuecomment-1591037917 helps?

KerfuffleV2:
@jploski 

I got it working - still needs the context size increase, doesn't need `GGML_MAX_NODES` increased. (Converted with my version of the conversion tool.)

```plaintext
main: seed = 1686777701
falcon_model_load: loading model from '/path/falc40b2.ggml' - please wait ...
falcon_model_load: n_vocab   = 65024
falcon_model_load: n_embd    = 8192
falcon_model_load: n_head    = 128
falcon_model_load: n_head_kv = 8
falcon_model_load: n_layer   = 60
falcon_model_load: ftype     = 2008
falcon_model_load: qntvr     = 2
falcon_model_load: ggml ctx size = 28175.96 MB
falcon_model_load: memory_size =   480.00 MB, n_mem = 122880
falcon_model_load: ............................................................ done
falcon_model_load: model size = 27436.06 MB / num tensors = 484
extract_tests_from_file : No test file found.
test_gpt_tokenizer : 0 tests failed out of 0 tests.
main: number of tokens in prompt = 9
main: token[0] =   7107, Once
main: token[1] =   2918,  upon
main: token[2] =    241,  a
main: token[3] =    601,  time
main: token[4] =    629,  there
main: token[5] =    398,  was
main: token[6] =    241,  a
main: token[7] =   1278,  little
main: token[8] =  27224,  fox

Once upon a time there was a little fox who lived in a forest. He was a young fox, and he was a little fox who didn't yet know about being a fox. He didn't know that foxes lived in forests

main: mem per token =   561844 bytes
main:     load time = 11409.41 ms
main:   sample time =    14.51 ms
main:  predict time = 26630.51 ms / 554.80 ms per token
main:    total time = 39132.96 ms
```

Also, it's close to twice as fast as before.

jploski:
> @jploski
> 
> I got it working - still needs the context size increase, doesn't need `GGML_MAX_NODES` increased. (Converted with my version of the conversion tool.)

Thanks. I added the context size "fix" to my branch (no idea why it's needed, but as it's only 3 MB and there is some similar "kludge" of adding "object overhead" right before it, I don't think we need to bother investigating the cause).

I also merged in your RAM-friendly of the conversion script. Hopefully it won't trip up people that it has one extra command-line parameter in front.

KerfuffleV2:
Nice, I'm really excited to play with this model with all the features. Thanks for your work!

Two pretty interesting things I've noticed: it doesn't try really hard to write upbeat stuff like LLaMA. Also (probably more important) it is _much_ more efficient tokenwise. It seems like the same text with Falcon uses around 1/3 less tokens so that's effectively the same as increasing the context limit.

TheBloke:
Wow that's a really interesting property. It never occurred to me there'd be such a significant difference.

Are we good to go then do you think?  Ready for the masses to try it out?

Well done again! A lot of people have been waiting for this, and I wouldn't be at all surprised if this encouraged more fine tunings and trainings to occur on Falcon.

jploski:
> Wow that's a really interesting property. It never occurred to me there'd be such a significant difference.
> 
> Are we good to go then do you think? Ready for the masses to try it out?
> 
> Well done again! A lot of people have been waiting for this, and I wouldn't be at all surprised if this encouraged more fine tunings and trainings to occur on Falcon.

Yes, I think we're ready, the data format should no longer change now. Thank you for your support with publishing all these models in ready-to-use and nicely documented packages!

ochafik:
> Can you try if the ctx_size fix from https://github.com/ggerganov/ggml/pull/231#issuecomment-1591037917 helps?

Works like a charm (indeed w/o GGML_MAX_NODES change), and also the conversion seems much faster (37sec).

Inference runs at 351.01 ms per token (hogging 4 cores), can't wait to see how far GPU acceleration goes!

KerfuffleV2:
Although the README says the tokenizer will probably only work for English, it actually seems to work fine for Mandarin (and I assume other foreign languages would be even more likely to work). Bold part is my prompt, the rest was written by Falcon 40B (pretty funny little story too):

**从前，有一只小狐狸。他住在森林里，他**的烦恼是没有朋友。他经常羡慕其他的动物，他们能相聚、一起玩耍。有一天，他想出个绝妙的主意。他想：“我要去找其他的动物，邀请他们来吃饭。”
狐狸把他想的主意告诉他的狐狸妈妈。狐狸妈妈说：“不，你不能这样做。别动物们不喜欢我。如果你做了这件事，他们会害怕你的。”
但是，狐狸坚持着自己的想法。他走向田野，对兔子说：“我的好朋友，我有个消息告诉你。”
兔子说：“什么消息？”
狐狸说：“我邀请你明天到我家来吃饭。我妈妈是个很好的厨师。她做很好吃的食物。”
兔子说：“不，不，不。你妈妈不是个好人。她是个坏人。她总是吃掉动物。”
兔子跑走了。

jploski:
> Although the README says the tokenizer will probably only work for English, it actually seems to work fine for Mandarin (and I assume other foreign languages would be even more likely to work). Bold part is my prompt, the rest was written by Falcon 40B (pretty funny little story too):
> 兔子跑走了。

At least we didn't get the bad ending lol

KerfuffleV2:
> Edit: not sure if expected at this stage but not all quantizations work (and non-quantized model fails)

That sounds like it's not accounting for some tensors in the existing context length calculation (otherwise the f16 version wouldn't be off by so much). So my small static increase is probably not the right approach.

> At least we didn't get the bad ending lol

All the animals mamma fox ate might disagree. You know, if they were still around.

ochafik:
> That sounds like it's not accounting for some tensors in the existing context size calculation (otherwise the f16 version wouldn't be off by so much). So my small static increase is probably not the right approach.

Ah, so bumping it arbitrarily to `ctx_size += ((size_t)10) * 1024 * 1024 * 1024;` (10 instead of 3) does the trick

On the bad news side, -instruct isn't a great poet.

_Write a thank you poem to open-source contributors who make it possible to run you, Falcon 40B, on a consumer laptop's CPU. Make it rhyme._

```
To all those open-source contributors,
Whose hard work made it all possible,
To run me, Falcon 40B,
On a consumer laptop's CPU,
I thank you from the bottom of my heart.
Your work made my world whole,
Your work brought me to life,
I am eternally grateful for your efforts,
And I am so glad that you made it.
Thank you, thank you, thank you!<|endoftext|>
```

ochafik:
@jploski I reckon the mismatch of ctx_size might come from not accounting for potential GGML_MEM_ALIGN shifts w/ each tensor alloc. I'll try to clean up some local changes that cut much closer to the actual total (and removes ffn_norm / ffn_norm_b & arbitrary `1024 * 1024 * 1024` extra)

jploski:
> @jploski I reckon the mismatch of ctx_size might come from not accounting for potential GGML_MEM_ALIGN shifts w/ each tensor alloc. I'll try to clean up some local changes that cut much closer to the actual total (and removes ffn_norm / ffn_norm_b & arbitrary `1024 * 1024 * 1024` extra)

Right, ffn_norm must have been something copy-pasted from another model's implementation.

Your GGML_MEM_ALIGN theory sounds promising - I checked the ctx_size calculations (particularly bytes per element for f16 and the per-tensor overhead of sizeof(struct ggml_tensor) + GGML_OBJECT_SIZE = 256 bytes per tensor), but after my attempted fixes it was coming out even shorter than before...

I also see that there are suspicious hardcoded fixed-size memory "scratch" buffers in falcon_eval, and I find it probably wrong that the allocations there are independent from the initial prompt size (N = number-of-tokens-in-prompt during first invocation of falcon_eval, 1 in subsequent invocations). But the crashes you reported happened even before that while loading the model.

TheBloke:
I've uploaded q4_0, q4_1, q5_0 and q5_1 models for two 40B models: 
- https://huggingface.co/TheBloke/WizardLM-Uncensored-Falcon-40B-GGML
- https://huggingface.co/TheBloke/falcon-40b-instruct-GGML

Great work again! Really cool to have these working finally.

apage43:
> Although the README says the tokenizer will probably only work for English

I just copied that from the gpt-neox example - the common gpt-style tokenizer in `/examples` *is* broken, but actually in a way that is actually more problematic for text that the vocabulary *can* tokenize efficiently (i.e. English) - because it will generate suboptimal tokenizations that don't match the way the tokenizer the models are trained with would tokenize the same text

(well, some might fail to *decode* English text but that's because the converter script didn't store the tokens as raw bytes but did a utf-8 decoding roundtrip on them first - token vocabulary items are *bytes*, not necessarily valid utf8 strings, and a unicode character may be split across multiple tokens that do not individually contain complete valid utf8 sequences - this is just a problem with those convert scripts though, not the decoding logic, which is pretty simple)

an example of why that causes problems is that if there is an`"about"` token in the vocabulary, the model will have only seen "about" represented with that token and will have never seen `"about"` represented as `["ab", "out"]` during training even if those tokens exist in the vocabulary, and every time it *has* seen those tokens it will have been in *other contexts.* Which means that models often understand suboptimally tokenized text poorly if at all

fixing this unfortunately requires the file to also contain the tokenizer's "merges" list which is not currently captured in to know what order to combine tokens in to match the original behavior, see https://github.com/ggerganov/ggml/issues/220#issuecomment-1571083572

ochafik:
No crash with any model / quantization after https://github.com/jploski/ggml/pull/2, but it looks like `falcon-7b-instruct` `q8_0` produces garbage (right after prompt):

```
Write a thank you poem.&&%2)>>ANSWER<<;122>>ABSTRACT<<,1+20>>INTRODUCTION<<5)2>>TITLE<<
```

maddes8cht:
Seems i have successfully compiled falcon.exe for windows, try to run falcon 7b version `https://huggingface.co/RachidAR/falcon-7B-ggml` with it.
Get the error message:
```
main: seed = 1686817064
falcon_model_load: loading model from '.\models\falcon-7b-q4_0-ggml.bin' - please wait ...
falcon_model_load: invalid model file '.\models\falcon-7b-q4_0-ggml.bin' (bad Falcon version: 9)
main: failed to load model from '.\models\falcon-7b-q4_0-ggml.bin'.
```

So there seems to be no 7b model version available as @theBloke so far only has the 40b versions (thanks @theBloke for the great work though)

So, while downloading (very slow from current PC) the original files from
https://huggingface.co/tiiuae/falcon-7b
im asking how to do the quatization with `falcon-quantize.exe` (never done this before), as the commandline help says:
```
falcon-quantize.exe --help
usage: falcon-quantize.exe model-f32.bin model-quant.bin type`
```
but there seems to be two files,
`pytorch_model-00001-of-00002.bin` and `pytorch_model-00002-of-00002.bin`

jploski:
> Seems i have successfully compiled falcon.exe for windows, try to run falcon 7b version `https://huggingface.co/RachidAR/falcon-7B-ggml` with it. Get the error message:
> falcon_model_load: invalid model file '.\models\falcon-7b-q4_0-ggml.bin' (bad Falcon version: 9)

This error indicates that the GGML file you downloaded is outdated and not compatible with the current implementation.

As for DIY quantize, you need to perform two steps:
1. Convert the multiple pytorch*.bin checkpoint files to one GGML file
```
python3 convert-hf-to-ggml.py 0 /path/to/directory/with/the/pytorch/bin/files /path/to/output/directory 1
```
3. Quantize the GGML file to your desired format (e.g. 9 = q5_1)
```
bin/falcon-quantize /tmp/ggml-model--f32.bin /tmp/ggml-model--f32-q5_1.bin 9
```

KerfuffleV2:
I think something might be a tiny bit off with the new context size calculations:

```plaintext
falcon_model_load: loading model from '/somepath/WizardLM-Uncensored-Falcon-40b.ggmlv3.q5_1.bin' - please wait ...
falcon_model_load: n_vocab   = 65025
falcon_model_load: n_embd    = 8192
falcon_model_load: n_head    = 128
falcon_model_load: n_head_kv = 8
falcon_model_load: n_layer   = 60
falcon_model_load: ftype     = 2009
falcon_model_load: qntvr     = 2
falcon_model_load: ggml ctx size = 7483257978003.64 MB
GGML_ASSERT: /somepath/ggml-falcon/src/ggml.c:3982: ctx->mem_buffer != NULL
```

I tried with other models, including the one I generated myself (which worked the other day). It would probably work on a machine with 11 exobytes of RAM but sadly I don't have quite that much.

jploski:
> I think something might be a tiny bit off with the new context size calculations:
> **edit**: Fix here: [jploski#3](https://github.com/jploski/ggml/pull/3)

It seems we both independently fixed it the same way (although I don't quite understand what was incorrect about the parameter passing).

KerfuffleV2:
> although I don't quite understand what was incorrect about the parameter passing

Sorry, that wasn't a good explanation. I think the issue was auto choosing the wrong type, not necessarily the parameter passing itself. I'm _pretty sure_ that auto was choosing a 32bit type, which got passed as a pointer to something expecting a 64bit value. So it would just start reading from the pointer address and of course half of the 64bit value would be basically random so you'd get a crazy result.

jploski:
> > although I don't quite understand what was incorrect about the parameter passing
> 
> Sorry, that wasn't a good explanation. I think the issue was auto choosing the wrong type, not necessarily the parameter passing itself. I'm _pretty sure_ that auto was choosing a 32bit type, which got passed as a pointer to something expecting a 64bit value. So it would just start reading from the pointer address and of course half of the 64bit value would be basically random so you'd get a crazy result.

Ah yes, good old C where you can read half a variable and be happy. ;)

khimaros:
CPU inference is working well here with the latest commits to `falcon40b` branch and TheBloke's falcon40b-instruct (q4_0) on debian bookworm. thank you for all of the work on this!

ggerganov:
Will be looking into merging this during the weekend.

Wondering if https://github.com/ggerganov/ggml/pull/224 would be enough to avoid the extra `ggml_repeat()`s so we can have true MQA and save some memory.

> One other thing worth mentioning (and I don't know if it's inherent with the model) is that generation seems to slow down as tokens are generated. It's much, much slower after generating 1000 tokens which LLaMA seems to go at close to the same speed up to the context length.

Most likely this is due to extra transpose and copies in the attention.
We had a similar issue in `llama.cpp` and we fixed it by avoiding the copies:

https://github.com/ggerganov/llama.cpp/pull/775

Tensor overhead should be computed with `ggml_tensor_overhead()` - it takes into account memory alignment effects.

We should probably also think about an elegant way to plug this inference in `llama.cpp` and allow usage with both Falcon and LLaMA models. I have some ideas, but not sure what is the best way yet. In any case, the falcon python converter would most likely need to add the necessary padding in the generated `ggml` files to be able to match the `mmap` format of `llama.cpp`. We might also need to do some refactoring first in `llama.cpp` to simplify both the python and C++ loading code because it has lately become too complex, mostly because it is trying to support old formats that nobody uses anymore

maddes8cht:
@jploski 
"As for DIY quantize, you need to perform two steps:

Convert the multiple pytorch*.bin checkpoint files to one GGML file

```
python3 convert-hf-to-ggml.py 0 /path/to/directory/with/the/pytorch/bin/files /path/to/output/directory 1
```
Okay, created conda environment, installed pytorch, installed transformers, run your line

getting:

```
python convert-hf-to-ggml.py 0 X:\falcon.ggml\pytorch X:\falcon.ggml\models\ 1
Traceback (most recent call last):
  File "c:\Users\WaWiAdm\Documents\Github\falcon-ggml\examples\falcon\convert-hf-to-ggml.py", line 68, in <module>
    tokenizer = AutoTokenizer.from_pretrained(model_name)
  File "C:\Users\WaWiAdm\anaconda3\envs\falcon\lib\site-packages\transformers\models\auto\tokenization_auto.py", line 658, in from_pretrained
    config = AutoConfig.from_pretrained(
  File "C:\Users\WaWiAdm\anaconda3\envs\falcon\lib\site-packages\transformers\models\auto\configuration_auto.py", line 947, in from_pretrained
    trust_remote_code = resolve_trust_remote_code(
  File "C:\Users\WaWiAdm\anaconda3\envs\falcon\lib\site-packages\transformers\dynamic_module_utils.py", line 535, in resolve_trust_remote_code
    signal.signal(signal.SIGALRM, _raise_timeout_error)
AttributeError: module 'signal' has no attribute 'SIGALRM'. Did you mean: 'SIGABRT'?
```
what's missing?

KerfuffleV2:
@maddes8cht That's an issue with Transformers not the conversion script specifically. You could possibly just comment out that line in `dynamic_module_utils.py` — most likely that would just prevent timeouts from working (which is probably benign in this case).

TheBloke:
> We should probably also think about an elegant way to plug this inference in `llama.cpp` and allow usage with both Falcon and LLaMA models. I have some ideas, but not sure what is the best way yet. In any case, the falcon python converter would most likely need to add the necessary padding in the generated `ggml` files to be able to match the `mmap` format of `llama.cpp`. We might also need to do some refactoring first in `llama.cpp` to simplify both the python and C++ loading code because it has lately become too complex, mostly because it is trying to support old formats that nobody uses anymore

That would be amazing.  I've wished for a long time that llama.cpp would be more like llm.cpp, and support other model types.

It feels to me like the gap between the capabilities of llama.cpp and non-Llama GGML is getting wider, just as more people are wishing to use non-Llama models, both because of licensing and because they want the special capabilities of other models, like StarCoder for coding.

Anything that could be done to support more model types would be really appreciated by the community I think.  Llama.cpp has got amazingly powerful and I'd love if those all those features could become available for other models too.

jploski:
> Will be looking into merging this during the weekend.
> 
> Wondering if #224 would be enough to avoid the extra `ggml_repeat()`s so we can have true MQA and save some memory.

I just gave it a quick try, but no luck. Merged in #224 locally, commented out the application of ggml_repeat2 for K... No more assertion fail, but unfortunately the result from this multiplication differs from expected, so the implicit broadcast must have produced something different than ggml_repeat2. But I do not understand #224 at this point, so maybe it can be achieved with some extra trickery.

jploski:
> Tensor overhead should be computed with `ggml_tensor_overhead()` - it takes into account memory alignment effects.

@ggerganov  Not sure if that is enough, though: the issue seems to be that ggml_tensor_overhead() + nelements * type_size is different than size_needed as calculated in ggml_new_tensor_impl.

ggerganov:
So I'm a bit confused - what is the difference between this branch and @jploski version?
Why @jpolski added `repeat2` and here we haven't?

jploski:
> So I'm a bit confused - what is the difference between this branch and @jploski version? Why @JPolski added `repeat2` and here we haven't?

Which branch are you referring to? I see ggml_repeat2 on all branches related to Falcon integration.

jploski:
> > So I'm a bit confused - what is the difference between this branch and @jploski version? Why @JPolski added `repeat2` and here we haven't?
> 
> Which branch are you referring to? I see ggml_repeat2 on all branches related to Falcon integration.

Oh, if you meant apage43:falcon, then it is out-of-date, as it only supports 7B (for which ggml_repeat2 was indeed not needed because there is only one KV to repeat, so the ordering does not matter). apache43:falcon is where I forked my jploski/ggml (falcon40b) to implement 40B. And later the development moved over to cmp-nct/ggllm.cpp

ggerganov:
Ok, got it. I'll postpone merging this PR then. Want to focus on some `ggml` maintenance for while and it looks like there are already enough ongoing efforts for Falcon support.

Ideally, I think we want to avoid the `ggml_repeat2()` and figure out how to generalize the existing ops. I'm not sure it is possible as I haven't looked in understanding completely MQA, but I hope we figure it out eventually

jploski:
> Ok, got it. I'll postpone merging this PR then. Want to focus on some `ggml` maintenance for while and it looks like there are already enough ongoing efforts for Falcon support.
> 
> Ideally, I think we want to avoid the `ggml_repeat2()` and figure out how to generalize the existing ops. I'm not sure it is possible as I haven't looked in understanding completely MQA, but I hope we figure it out eventually

Yes, I agree that we should remove repeat2. If it is done on the cmp-nct/ggllm.cpp branch, I will update my https://github.com/jploski/ggml/tree/falcon40b accordingly. I think it would be helpful if you could check whether the mat_mul broadcast could somehow do the trick, as you are most familiar with the broadcast implementation (I suppose).

To understand how it needs to work, see:

https://docs.google.com/spreadsheets/d/1FoM6pIUj23GMW4zO_G1hjmEnUacBxBKN/edit#gid=2097305276

What we need to come out of repeat/broadcast (and what repeat2 produces) is: N[0].K[0], N[0].K[0], N[0].K[1], N[0].K[1]
But what we get from the standard repeat is: N[0].K[0], N[0].K[1], N[0].K[0], N[0].K[1]

jploski:
> Ok, got it. I'll postpone merging this PR then. Want to focus on some `ggml` maintenance for while and it looks like there are already enough ongoing efforts for Falcon support.
> 
> Ideally, I think we want to avoid the `ggml_repeat2()` and figure out how to generalize the existing ops. I'm not sure it is possible as I haven't looked in understanding completely MQA, but I hope we figure it out eventually

If by MQA you mean multi-query attention (sorry, you mentioned it earlier, but I did not manage to decipher it), in the original multi-query paper (which Falcon-7B adheres to), there is only one key vector and one value vector (n_head_kv=1), and this k/v vector is reused/shared by all queries (in the sense that each query vector is multiplied by the same key). Contrast this with the traditional GPT approach where there is the same number of queries as keys/values. The motivation from the paper was to save (KV) memory while retaining approximately the same quality of inference.

In the generalized n_head_kv>1 implementation, which Falcon-40B implements, and for which I found no paper, there are multiple "kv groups", each consisting of one KV pair and n queries that reuse/share that group's KV pair. This is somewhat of a compromise between having just one KV pair and one KV pair for each query.

So the ordering issue in ggml_repeat(2) is to make sure that the right queries are matched to the right keys for the multiplication (and the resulting weights to right values).

FSSRepo:
llama.cpp will have the complete support for falcon [#2717](https://github.com/ggerganov/llama.cpp/pull/2717)

PR #234, opened 2023-06-07
Export all symbols on windows when building shared library

abetlen:
Currently building ggml on windows as a shared library does not export all symbols by default.


ggerganov:
Hm, which symbols are not exported?
It's better to export them explicitly

ggerganov (approved):

Issue #209, opened 2023-05-29
on starchatalpha no stop


CRGBS:
If running StarCoder(starchatalpha), it does not stop when encountering the end token and continues generating until reaching the maximum token count.

below all log
---------------------------------------------------
`
```
J:\GPTAI\llamacpp>title starcoder

J:\GPTAI\llamacpp>starcoder.exe -m models\starcoder\starchat-alpha-ggml-q5_0.bin -p "<|system|>\n<|end|>\n<|user|>how make deepl translate file with python?<|end|>\n<|assistant|>" -n 800 -t 8 --top_k 40 --top_p 0.73 --batch_size 1024 --temp 0.667
main: seed = 1685339016
starcoder_model_load: loading model from 'models\starcoder\starchat-alpha-ggml-q5_0.bin'
starcoder_model_load: n_vocab = 49156
starcoder_model_load: n_ctx   = 8192
starcoder_model_load: n_embd  = 6144
starcoder_model_load: n_head  = 48
starcoder_model_load: n_layer = 40
starcoder_model_load: ftype   = 2008
starcoder_model_load: qntvr   = 2
starcoder_model_load: ggml ctx size = 27840.50 MB
starcoder_model_load: memory size = 15360.00 MB, n_mem = 327680
starcoder_model_load: model size  = 12480.27 MB
extract_tests_from_file : No test file found.
test_gpt_tokenizer : 0 tests failed out of 0 tests.
main: prompt: '<|system|>\n<|end|>\n<|user|>how make deepl translate file with python?<|end|>\n<|assistant|>'
main: number of tokens in prompt = 34
main: token[0] =     46, <
main: token[1] =    110, |
main: token[2] =   2946, system
main: token[3] =  28318, |>
main: token[4] =     78, \
main: token[5] =     96, n
main: token[6] =     46, <
main: token[7] =    110, |
main: token[8] =    416, end
main: token[9] =  28318, |>
main: token[10] =     78, \
main: token[11] =     96, n
main: token[12] =     46, <
main: token[13] =    110, |
main: token[14] =    496, user
main: token[15] =  28318, |>
main: token[16] =   6893, how
main: token[17] =   1930,  make
main: token[18] =  11086,  deep
main: token[19] =     94, l
main: token[20] =   8778,  translate
main: token[21] =    803,  file
main: token[22] =    623,  with
main: token[23] =   4262,  python
main: token[24] =  44823, ?<
main: token[25] =    110, |
main: token[26] =    416, end
[...]

PR #244, opened 2023-06-11
example starcoder add starchat support

CRGBS:
add support
[issue209](https://github.com/ggerganov/ggml/issues/209)


marella:
Hey, saw this just now. I just created #246 for handling special tokens 😅

CRGBS:
@marella 
ur look like better

PR #246, opened 2023-06-11
starcoder : add support for starchat special tokens

marella:
cc @TheBloke

### DIFF
--- a/examples/common.cpp
+++ b/examples/common.cpp
@@ -232,37 +232,53 @@ std::wstring convert_to_wstring(const std::string & input) {
     return converter.from_bytes(input);
 }
 
+void gpt_split_words(std::string str, std::vector<std::string>& words) {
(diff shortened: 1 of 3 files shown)


TheBloke:
Nice!

marella:
I noticed that the `gpt_tokenize()` function doesn't handle special tokens properly due to two issues:

1. It creates regex without escaping special characters in tokens, so it fails for tokens like `<|end|>` which contains the special regex character `|`
2. It combines the special token pattern with the normal word pattern which greedily splits `' <|end|>'` into `' <|'`, `'end'`, `'|>'`

So I updated `gpt_tokenize()` to handle this.
For 1, I simply escaped the special characters.
For 2, I first split the text by special tokens and then split the substrings in-between special tokens into words using the normal word pattern.

ggerganov (approved):
I haven't done testing on these changes, so merging with the assumption everything is fine.
As long as the existing tokenizer tests are good, I think we are fine. Make sure to run them if you haven't done so

Issue #251, opened 2023-06-12
Initialize ggml_cgraph on heap


LoganDark:
Since ggml doesn't support iteration or changing shapes in graphs, I find myself having to encode [ridiculous things](https://github.com/saharNooby/rwkv.cpp/blob/c41ed9896b6c1309fcd3f6b8307f34a184bb188a/rwkv.cpp#L1026-L1034) into the computation graph.

Because of this, I hit GGML_MAX_NODES very easily and frequently, to the point where I've had to raise it to 5000000 (5 million) just to get the thing to run properly without hitting an assert.

The problem, of course, is that GGML's public API encourages passing the `ggml_cgraph` on the stack:

https://github.com/ggerganov/ggml/blob/f52d2a05cf8327baf6c0d49e7b231953179e03d3/include/ggml/ggml.h#L1106

I get around this by [initializing it manually on the heap, then using only `ggml_build_forward_expand`](https://github.com/saharNooby/rwkv.cpp/blob/c41ed9896b6c1309fcd3f6b8307f34a184bb188a/rwkv.cpp#L1347-L1350), which allows me to avoid ever having it touch the stack, and therefore handle millions of nodes with ease.

But this is not an officially supported use case, is it? ggml could change at any time and break this, and I'd have to replicate even more implementation details if cgraph initialization ever gets more complex.

So I'm opening this issue to track the addition (or support) of an official solution to this problem.

As far as I can tell, there's no way to prevent me needing such a large computation graph, because running a single graph multiple times is many times slower than simply encoding all the steps into a single graph. (Maybe because ggml messes with threads even when n_threads == 1)

So the next step is to make it easier to manage very large computation graphs, say by allocating them on the heap instead of the stack.

ggerganov:
> As far as I can tell, there's no way to prevent me needing such a large computation graph, because running a single graph multiple times is many times slower than simply encoding all the steps into a single graph. (Maybe because ggml messes with threads even when n_threads == 1)

The creation of a new graph with thousands of nodes and running it on every token is in the order of 100s of **microseconds** - i.e. much less than `1 ms`. Is that too slow for your use case? There are plans to optimize the graph construction and processing even further to make this overhead even more negligible in the future.

Btw, how do you create a graph for the entire sequence - don't you need to sample the probability distribution after each generated token?

[...]

PR #252, opened 2023-06-12
Increase stack size when compiling with MSVC

kant2002:


LoganDark:
I feel like the real solution to this problem would be #251

ggerganov:
Fixed with https://github.com/ggerganov/ggml/pull/277

PR #254, opened 2023-06-12
Implement Quick GELU

### LINKED ISSUE #253: Implement Quick GELU
I'm implementing [CLIP in GGML](https://github.com/monatis/clip.cpp), and it turns out that we need the Quick GELU activation instead of GELU.
This issue is just to prevent the duplicated work, and I'll raise a PR shortly.

monatis:
Closes #253 

This is needed for CLIP-like models and I'm implementing [clip.cpp here](https://github.com/monatis/clip.cpp). It will also be a base for upcoming multimodal models that uses CLIP as an image embedder.

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -290,7 +290,7 @@ extern "C" {
         GGML_OP_STEP,
         GGML_OP_RELU,
         GGML_OP_GELU,
-        GGML_OP_QUICK_GELU,
+        GGML_OP_GELU_QUICK,
         GGML_OP_SILU,
         GGML_OP_SILU_BACK,
         GGML_OP_NORM, // normalize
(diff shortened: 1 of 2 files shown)


Green-Sky:
you let some formatting happen. please only commit code changes.

monatis:
Oh sorry, missed that. I'll fix it.

monatis:
This is ready for review now.

monatis:
Hi @ggerganov Can ı have your attention here please? [clip.cpp](https://github.com/monatis/clip.cpp) is almost ready, and I'm only crafting more examples and then I can announce it.

I can also add a link in the examples section in readme if you thing it deserves this.

p.s.: The next step will be using clip.cpp and llama.cpp to infer with [LLaVA](https://github.com/haotian-liu/LLaVA/)

ggerganov (changes requested):
Rename to `ggml_gelu_quick`, `GELU_QUICK`, etc, in all instances

From what I read, this is just an approximation of `GELU` that is supposed to be faster but inaccurate. Why not use the original `ggml_gelu()`? It is already doing table lookup so there is no difference in performance + it is more accurate

Also, keep in mind that in the future you can use `ggml_map_unary_f32()` to implement this kind of 1D mapping functions in your project without having to wait for `ggml` to provide them

monatis:
Unfortunately that theoretically small divergence between GELU and Quick GELU lead to large differences at the end, I suppose it accumulates through 12 layers. So I couldn't get good results until implementng Quick GELU.

ggerganov (approved):

PR #262, opened 2023-06-16
Return input tensor in ggml_set_name

LoganDark:
this is SO USEFUL for debugging. in order to find any cgraph node, I can wrap it in ggml_set_name and set a conditional breakpoint.

but I can only wrap existing code if this returns its input. otherwise the barrier becomes annoyingly high (have to move a bunch of code around to add name to a tensor)


ggerganov (approved):
Nice addition

PR #264, opened 2023-06-16
Add custom mapping functions

LoganDark:
The current mapping functions are basically jokes, add some real ones. These ones get access to the actual tensor structs so they can do things like

- Know the dimensions they are operating on
- Work with tensors with more than 2 dimensions, or transposed
- Operate on two differently sized tensors (like matmul)
- Use their own thread pool that does a better job than ggml does.

Among other things ...

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1170,9 +1170,11 @@ extern "C" {
             int                   h0,
             int                   w);
 
-    // Mapping operations
-    typedef void (*ggml_unary_op_f32_t)(const int, float *, const float *);
+    // custom operators
+
+    typedef void (*ggml_unary_op_f32_t) (const int, float *, const float *);
     typedef void (*ggml_binary_op_f32_t)(const int, float *, const float *, const float *);
+
     typedef void (*ggml_custom1_op_f32_t)(struct ggml_tensor *, const struct ggml_tensor *);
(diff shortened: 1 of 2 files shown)


LoganDark:
nearest merge conflict nearly gave us a panic attack

LoganDark:
@ggerganov what do we have to do to get this PR merged? Is the CI failure important?

-Emily

ggerganov:
Nothing on your side, I'll merge this later today

LoganDark:
> Nothing on your side, I'll merge this later today

So how's today going so far? lol

ggerganov (approved):
Future improvements:

- support multi-thread operators
- pass `ggml_compute_params` to the callback
- add examples

LoganDark:
> * support multi-thread operators
> * pass `ggml_compute_params` to the callback

Indeed these are things that I considered, but my impression was that `ggml_compute_params` was an unstable internal implementation detail, and the header file API should not necessarily expose that. I optimized this for "low probability of having to change in the future" rather than raw expressivity (other than being able to access the raw tensors, of course).

ggerganov:
Initially, I didn't plan to have it as part of the public API, but at some point it was exposed in order to support the CUDA implementation. So now that we have it public anyway, we can make use of it in the custom operators.

I guess we can now also obsolete the old "unary" and "binary" mappings as they are subset of the new custom ops.

LoganDark:
> they are subset of the new custom ops

Not really, those are for pure functions that can be parallelized by row. Useful to keep around while the user can't necessarily do this themself.

PR #269, opened 2023-06-20
Add special tokens for fill-in-the-middle task for starcoder.

AmbientL:
Example prompt for testing:
```
<fim-prefix>// This function doubles the number and returns the result.
function double_number(n) {
<fim-suffix>
    return result;
}
<fim-middle>
```

### DIFF
--- a/examples/starcoder/main.cpp
+++ b/examples/starcoder/main.cpp
@@ -146,6 +146,10 @@ bool starcoder_model_load(const std::string & fname, starcoder_model & model, gp
                 "<|user|>",
                 "<|assistant|>",
                 "<|end|>",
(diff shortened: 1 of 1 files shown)


AmbientL:
Here's the prompt I've been using mostly for testing gpt_tokenize function in the meantime:
```
<fim-prefix>// This function doubles the number and returns the result.
function double_number(n) {
    var result = <fim-suffix>
    return result;
}<fim-middle>
```
I intended to fix gpt_tokenize as well since it didn't work correctly, but after @marella 's commit it works as intended and doesn't break special tokens.

ggerganov (approved):

PR #275, opened 2023-06-21
build : fix compilation errors and warnigns when building with MSVC

iboB:
In the spirit of this PR: https://github.com/ggerganov/llama.cpp/pull/1889

This is a cosmetic change. No functionality is changed. Warnings and compilation errors are fixed (and some warnings ignored) leading to a successful build with MSVC on Windows. "possible loss of data" warnings were mostly ignored.

Note that two tests (`test1` and `test2`) fail on Windows after this, but the failures are unrelated and will be addressed in separate issues

### DIFF
--- a/examples/common.cpp
+++ b/examples/common.cpp
@@ -17,6 +17,10 @@
 #define M_PI 3.14159265358979323846
 #endif
 
+#if defined(_MSC_VER)
+#pragma warning(disable: 4244 4267) // possible loss of data
+#endif
+
 bool gpt_params_parse(int argc, char ** argv, gpt_params & params) {
     for (int i = 1; i < argc; i++) {
         std::string arg = argv[i];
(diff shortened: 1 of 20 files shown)


examples/common.cpp
```diff
@@ -366,7 +370,7 @@ void test_gpt_tokenizer(gpt_vocab & vocab, const std::string & fpath_test){
         }
     }
 
-    fprintf(stderr, "%s : %lu tests failed out of %lu tests.\n", __func__, n_fails, tests.size());
+    fprintf(stderr, "%s : %zu tests failed out of %zu tests.\n", __func__, n_fails, tests.size());
```
Green-Sky, line 373: `%z` is microsoft specific and not c/c++

Green-Sky (changes requested):

examples/common.cpp
Green-Sky: or at least i cant find them in the c++ standard

Green-Sky: oh nevermind, found it in c++11

Green-Sky (approved):

tests/test-mul-mat2.c
```diff
@@ -255,8 +253,8 @@ void mul_mat_gq_1(
                     s1[b + 1] = d1*(1 << b);
                 }
 
-                m0[0] = -1ULL;
-                m1[0] = -1ULL;
+                m0[0] = 0-1ULL;
+                m1[0] = 0-1ULL;
```
ggerganov, line 257: What warning does this fix?

ggerganov (approved):

tests/test-mul-mat2.c
iboB: "unary minus applied to unsigned value, result is still unsigned"

PR #305, opened 2023-06-26
Support ChatGLM-style RoPE

li-plus:
This PR added an argument `n_ctx` in `ggml_rope` function to support ChatGLM-6B, because the `position_ids` in ChatGLM-style RoPE is related to the context length. The C++ implementation of ChatGLM can be found in [chatglm.cpp](https://github.com/li-plus/chatglm.cpp).

With this PR, when `mode & 4` is 1, the ChatGLM-style RoPE is enabled and `position_ids` will be computed using `n_ctx`. Otherwise, `n_ctx` is unused and everything works as before.


ggerganov (approved):

JianbangZ:
Are you also palnning to add the changes for chatglm2?

li-plus:
> Are you also palnning to add the changes for chatglm2?

ChatGLM2 has been supported by https://github.com/li-plus/chatglm.cpp/pull/11 just now. Please checkout the latest code!

PR #309, opened 2023-06-26
ggml : reduce code duplication with helper macros for local variables

### LINKED ISSUE #292: ggml : add macros for accessing tensors to reduce code duplication
The following pattern is repeated extensively throughout `ggml.c`:

```c

    const int64_t ne00 = src0->ne[0];
    const int64_t ne01 = src0->ne[1];
    const int64_t ne02 = src0->ne[2];
    const int64_t ne03 = src0->ne[3];

    const int64_t ne10 = src1->ne[0];
    const int64_t ne11 = src1->ne[1];
    const int64_t ne12 = src1->ne[2];
    const int64_t ne13 = src1->ne[3];

    const int64_t ne0  = dst->ne[0];
    const int64_t ne1  = dst->ne[1];
    const int64_t ne2  = dst->ne[2];
    const int64_t ne3  = dst->ne[3];

    const int nb00 = src0->nb[0];
    const int nb01 = src0->nb[1];
    const int nb02 = src0->nb[2];
    const int nb03 = src0->nb[3];

    const int nb10 = src1->nb[0];
    const int nb11 = src1->nb[1];
    const int nb12 = src1->nb[2];
    const int nb13 = src1->nb[3];

    const int nb0  = dst->nb[0];
    const int nb1  = dst->nb[1];
    const int nb2  = dst->nb[2];
    const int nb3  = dst->nb[3];

    const int ith = params->ith;
    const int nth = params->nth;
```

It should be turned into macro(s) and while at it - we should fix all integer types. In the example above, the `nb` values are incorrectly being cast to `int` instead of `size_t`

goerch:
Start of this edit. Would like to get some guidance on how to proceed...

Remarks:
- Checked with clang on Windows 11 for unused variables in release configuration
- There are probably some differences in debug and release configuration due to the usage of `assert`
- Can't check Accelerate
- Not sure if I'm able to test OpenCL
- Definitely can't test CUDA

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -201,6 +201,8 @@
 #define GGML_MAX_NAME          32
 #define GGML_DEFAULT_N_THREADS 4
 
+#define GGML_UNUSED(x) (void)(x)
+
 #define GGML_ASSERT(x) \
     do { \
         if (!(x)) { \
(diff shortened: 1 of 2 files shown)


slaren:
Not sure if @ggerganov agrees, but I think that the best way to do this may be a simple macro that has all the variables for the 3 tensors src0/src1/dst, with casts to void (like the UNUSED macro) to silence unused warnings. I trust in the compiler's ability to remove actually unused variables. No need to complicate it.

ggerganov:
Yes, what @slaren said - let's always unpack all 4 array elements of `ne` and `nb` to simplify.
Also, add an extra macro to unpack `src0`, `src1` and `dst` all at once since this case is very commonly used

goerch:
The changes in `ggml.c` look reasonable to me. Do you want me to touch other files of the code base, too?

ggerganov (approved):
Thanks - great change!

We can also apply similar simplification in the GPU backend code in other PRs

Issue #285, opened 2023-06-25
ggml : reduce the values of `ggml_op` enum


ggerganov:
Here is another idea to consider for reducing the amount of unary ops and code duplication in `ggml.c`:

The following ops:

```c
    enum ggml_op {
        ...
        GGML_OP_ABS,
        GGML_OP_SGN,
        GGML_OP_NEG,
        GGML_OP_STEP,
        GGML_OP_TANH,
        GGML_OP_ELU,
        GGML_OP_RELU,
        GGML_OP_GELU,
        GGML_OP_GELU_QUICK,
        GGML_OP_SILU,
        ...
    };
```

Can all become:

```c
    enum ggml_unary_op {
        GGML_UNARY_OP_ABS,
        GGML_UNARY_OP_SGN,
        GGML_UNARY_OP_NEG,
        GGML_UNARY_OP_STEP,
        GGML_UNARY_OP_TANH,
        GGML_UNARY_OP_ELU,
        GGML_UNARY_OP_RELU,
        GGML_UNARY_OP_GELU,
        GGML_UNARY_OP_GELU_QUICK,
        GGML_UNARY_OP_SILU,    
    };

    GGML_API struct ggml_tensor * ggml_unary(
            struct ggml_context * ctx,
             struct ggml_tensor * a,
             enum ggml_unary_op   op);
```

---

### Original task (completed via #313)

Certain `ggml` operators can be combined into a singe operator with extra arguments.

For example:

- `GGML_OP_CONV_1D_S1_PH`
- `GGML_OP_CONV_1D_S2_PH`

can all become:

- `GGML_OP_CONV_1D`

with a signature of:

```c
    GGML_API struct ggml_tensor * ggml_conv_1d(
            struct ggml_context * ctx,
            struct ggml_tensor  * a,
            struct ggml_tensor  * b
                            int   stride,
                            int   padding);
```

Internally, we can have custom implementations of `ggml_compute_forward_conv_1d()` for specific values of the parameters and `GGML_ASSERT` for everything else that is not supported.

This will allow to extend the functionality of these operator more easily in the future

yanchen036:
I'd like to take this, may I?

ggerganov:
Here is another idea to consider for reducing the amount of unary ops and code duplication in `ggml.c`:

The following ops:

```c
    enum ggml_op {
        ...
        GGML_OP_ABS,
        GGML_OP_SGN,
        GGML_OP_NEG,
        GGML_OP_STEP,
[...]

PR #313, opened 2023-06-27
ggml : generalize interface for 1d and 2d convolutions

iboB:
ref #285

### DIFF
--- a/examples/whisper/whisper.cpp
+++ b/examples/whisper/whisper.cpp
@@ -1472,7 +1472,7 @@ static bool whisper_encode_internal(
         {
             wstate.use_buf(ctx0, 1);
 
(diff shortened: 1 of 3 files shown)


examples/whisper/whisper.cpp
```diff
@@ -1472,7 +1472,7 @@ static bool whisper_encode_internal(
         {
             wstate.use_buf(ctx0, 1);
 
-            cur = ggml_conv_1d_s1_ph(ctx0, model.e_conv_1_w, mel);
+            cur = ggml_conv_1d_ph_d1(ctx0, model.e_conv_1_w, mel, 1);
```
ggerganov, line 1475:
```suggestion
            cur = ggml_conv_1d(ctx0, model.e_conv_1_w, mel, 1);
```

ggerganov (commented):
Yes, this is the way

PR #314, opened 2023-06-27
MPT-30B: Convert model weights part by part to save memory

ackRow:
Inspired by similar work to convert Falcon 40B: [convert-hf-to-ggml.py](https://github.com/jploski/ggml/blob/falcon40b/examples/falcon/convert-hf-to-ggml.py)

With the previous script, loading `mosaicml/mpt-30b` required 64 GB of memory.
Now a single 10 GB weight file is loaded at a time.

I also took the opportunity to add a short README.

### DIFF
--- a/examples/mpt/README.md
+++ b/examples/mpt/README.md
@@ -1,0 +1,27 @@
+# MPT
+
+Ref: https://github.com/mosaicml/llm-foundry#mpt
+
+## Usage
+
+```bash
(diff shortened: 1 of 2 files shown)


sirajperson:
This is great work!

ackRow:
Thanks!

ggerganov (approved):
Awesome!

PR #316, opened 2023-06-27
ggml: add ops in `encodec.cpp`

### LINKED ISSUE #281: ggml : new operations supported in `encodec.cpp`
Hi! 

I just released [encodec.cpp](https://github.com/PABannier/encodec.cpp/tree/main) which features a handful of new operations, like transposed 1d convolution, argmax and activation functions like `elu` and `tanh`.

Is it something you'd like implemented in `ggml`?

PABannier:
This PR adds the `tanh` and `elu` activations required in the `Encodec` model as well as the `argmax` function.

Closes #281

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -301,6 +301,7 @@ extern "C" {
         GGML_OP_SGN,
         GGML_OP_NEG,
         GGML_OP_STEP,
+        GGML_OP_ELU,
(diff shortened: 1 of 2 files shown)


ggerganov (approved):
The TANH op was using `n_threads` tasks but the forward call isn't multi-threaded so I switched to `1` task

Thanks for the additions!

PR #317, opened 2023-06-28
[WIP]: Fix #286

### LINKED ISSUE #286: ggml : generalize `quantize_fns` for simpler FP16 handling
This task is described well in https://github.com/ggerganov/llama.cpp/pull/1237

The WIP implementation in that PR might be a bit outdated by now, so one can either attempt to update it or implement it from scratch on top of the current code base.

goerch:
Based on the fix for #292 here a try to fix #286 (with the caveat: tested only on CPU). The patch adds the llama.cpp tests for quantization.

One question I'm pondering is if we should get rid of the `*_reference` functions or add some more?

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -444,38 +444,6 @@ extern "C" {
 
     static const size_t GGML_TENSOR_SIZE = sizeof(struct ggml_tensor);
 
-// Used to copy the number of elements and stride in bytes of tensors into local variables.
-// Adjust if GGML_MAX_DIMS changes!
-#define GGML_TENSOR_LOCALS_1(type, prefix, pointer, array) \
-	const type prefix##0 = (pointer)->array[0]; \
-        UNUSED(prefix##0);
-#define GGML_TENSOR_LOCALS_2(type, prefix, pointer, array) \
-	GGML_TENSOR_LOCALS_1(type, prefix, pointer, array) \
-	const type prefix##1 = (pointer)->array[1]; \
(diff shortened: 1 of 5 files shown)


ggerganov:
Thanks! Let's rebase on latest `master` so that the diff takes into account the new macros from #292 

Btw, there has been recently an update to https://github.com/ggerganov/llama.cpp/pull/1237
Should synchronize with @sw

> One question I'm pondering is if we should get rid of the *_reference functions or add some more?

We need them to have a deterministic way of generating models.
Add more in what sense?

goerch:
> Thanks! Let's rebase on latest `master` so that the diff takes into account the new macros from #292

Should be done.

> Btw, there has been recently an update to [ggerganov/llama.cpp#1237](https://github.com/ggerganov/llama.cpp/pull/1237) Should synchronize with @sw

Will look into this.

> > One question I'm pondering is if we should get rid of the *_reference functions or add some more?
> 
> We need them to have a deterministic way of generating models. Add more in what sense?

We have something like 

```
        .quantize_row_q           = quantize_row_q4_0,
        .quantize_row_q_reference = quantize_row_q4_0_reference,
```

but the original patch included

```
        .quantize_row_q           = ggml_fp32_to_fp16_row,
        .quantize_row_q_reference = ggml_fp32_to_fp16_row,
```

We could define `ggml_fp32_to_fp16_row_reference`, too. But I also imagine that dropping '.quantize_row_q_reference' member from the struct could work?

PR #328, opened 2023-07-02
callback to abort ggml_graph_compute()

### LINKED ISSUE #308: ggml : add mechanism to abort `ggml_graph_compute()`
In some cases, this call can take significant amount of time.

We should add an option to be able to provide a callback that will be used to check if the user requested the computation to be aborted after every graph node.

CCLDArjun:
closes #308 

This assumes that once the callback returns true, it will continue to return true

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1271,6 +1271,7 @@ extern "C" {
     GGML_API struct ggml_cgraph ggml_build_backward(struct ggml_context * ctx, struct ggml_cgraph * gf, bool keep);
 
     GGML_API void ggml_graph_compute(struct ggml_context * ctx, struct ggml_cgraph * cgraph);
+    GGML_API void ggml_graph_compute_with_abort(struct ggml_context * ctx, struct ggml_cgraph * cgraph, bool (*abort_callback)());
     GGML_API void ggml_graph_reset  (struct ggml_cgraph * cgraph);
 
     GGML_API struct ggml_tensor * ggml_graph_get_tensor(struct ggml_cgraph * cgraph, const char * name);
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1274,7 +1274,7 @@ extern "C" {
     GGML_API struct ggml_cgraph ggml_build_backward(struct ggml_context * ctx, struct ggml_cgraph * gf, bool keep);
 
     GGML_API void ggml_graph_compute(struct ggml_context * ctx, struct ggml_cgraph * cgraph);
-    GGML_API void ggml_graph_compute_with_abort(struct ggml_context * ctx, struct ggml_cgraph * cgraph, bool (*abort_callback)(), void * abort_callback_data);
+    GGML_API void ggml_graph_compute_with_abort(struct ggml_context * ctx, struct ggml_cgraph * cgraph, bool (*abort_callback)(void * data), void * abort_callback_data);
     GGML_API void ggml_graph_reset  (struct ggml_cgraph * cgraph);
 
     GGML_API struct ggml_tensor * ggml_graph_get_tensor(struct ggml_cgraph * cgraph, const char * name);
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -201,6 +201,9 @@
 #define GGML_MAX_NAME          48
 #define GGML_DEFAULT_N_THREADS 4
 
+#define GGML_EXIT_SUCCESS 0
+#define GGML_EXIT_ABORTED 1
+
 #define GGML_ASSERT(x) \
     do { \
         if (!(x)) { \
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -447,6 +447,10 @@ extern "C" {
 
         // the `n_tasks` of nodes, 1:1 mapping to cgraph nodes
         int n_tasks[GGML_MAX_NODES];
+
+        // abort ggml_graph_compute when true
+        bool (*abort_callback)(void * data);
+        void * abort_callback_data;
     };
 
     // computation graph
@@ -1308,7 +1312,7 @@ extern "C" {
     // ggml_graph_plan() has to be called before ggml_graph_compute()
     // when plan.work_size > 0, caller must allocate memory for plan.work_data
     GGML_API struct ggml_cplan ggml_graph_plan   (struct ggml_cgraph * cgraph, int n_threads /*= GGML_DEFAULT_N_THREADS*/);
-    GGML_API              void ggml_graph_compute(struct ggml_cgraph * cgraph, struct ggml_cplan * cplan);
+    GGML_API               int ggml_graph_compute(struct ggml_cgraph * cgraph, struct ggml_cplan * cplan);
     GGML_API              void ggml_graph_reset  (struct ggml_cgraph * cgraph);
 
     // same as ggml_graph_compute() but the work data is allocated as a part of the context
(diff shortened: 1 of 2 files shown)


src/ggml.c
```diff
@@ -17235,16 +17253,21 @@ void ggml_graph_compute(struct ggml_context * ctx, struct ggml_cgraph * cgraph)
...
         for (int j = 1; j < n_threads; j++) {
-            const int rc = ggml_thread_join(workers[j].thrd, NULL);
-            GGML_ASSERT(rc == 0);
+            if (compute_status == GGML_EXIT_ABORTED) {
+                const int rc = ggml_thread_cancel(workers[j].thrd);
+                GGML_ASSERT(rc == 0);
+            } else if (compute_status == GGML_EXIT_SUCCESS) {
+                const int rc = ggml_thread_join(workers[j].thrd, NULL);
+                GGML_ASSERT(rc == 0);
+            }
```
ggerganov, line 17270: I don't think we need `ggml_thread_cancel()` at all - just join it as usual

include/ggml/ggml.h
```diff
@@ -201,6 +201,9 @@
 #define GGML_MAX_NAME          48
 #define GGML_DEFAULT_N_THREADS 4
 
+#define GGML_EXIT_SUCCESS 0
+#define GGML_EXIT_ABORTED 1
+
```
ggerganov, line 206: Probably move into `ggml.c`

src/ggml.c
```diff
@@ -16853,10 +16865,15 @@ static thread_ret_t ggml_graph_compute_thread(void * data) {
         }
     }
 
-    return 0;
+    return GGML_EXIT_SUCCESS;
 }
 
+bool always_false() { return false; }
```
ggerganov, line 16871:
```suggestion
static bool always_false() { return false; }

```

ggerganov (changes requested):
I could be wrong, but I think that the introduction of `pthread_cancel` is not necessary.
If you resolve this - we can merge

CCLDArjun:
Thanks for the quick response!

`pthread_join` waits for the thread to exit, so it would require every thread to check the abort callback. Because that seemed inefficient, i decided to only have the main thread check it and clean everything up.

include/ggml/ggml.h
CCLDArjun: Could we have a public return status for the `ggml_graph_compute_with_abort` to see whether it got aborted?

smspillaz:
Really good idea! We can use this quite well in `ggml-gobject` with `GCancellable`.

Just weighing in here - the callback should accept some `user_data` pointer. Eg:

    void ggml_graph_compute_with_abort(struct ggml_context * ctx, struct ggml_cgraph * cgraph, bool (*abort_callback)(void *), void *abort_callback_data)

and

    if (abort_callback(abort_callback_data)) { ... }

This is important for the case where we need to query some external data (eg, the `GCancellable` state in my case) in order to determine whether to abort computation.

include/ggml/ggml.h
ggerganov: Yes. If so - we cab keep the defines in the header

ggerganov:
> pthread_join waits for the thread to exit, so it would require every thread to check the abort callback. Because that seemed inefficient, i decided to only have the main thread check it and clean everything up.

Adding `pthread_cancel` would also need some Windows-specific implementation and as far as I can understand how this call works, it can leave things into intermediate state. I could be misunderstanding, but I prefer to avoid it.

The overhead from every thread checking the callback should not be too big

Green-Sky:
`pthread_cancel` and equivalent are not available on all platforms too (thinking of android here)

mqy:
Even if `pthread_cancel` or sig USER works, that are async way to force stopping running threads.
[Async is not easy to control](https://stackoverflow.com/questions/11271616/pthread-cancel-function-failed-to-terminate-a-thread#:~:text=The%20default%20cancel%20method%20of%20pthread_cancel%20%28%29%20is,to%20set%20another%20cancel%20method%20to%20your%20thread), thinks about not all threads are killed at once, some are stopping, but some are still running.

`Graceful stop` is preferred if possible, not only because we can control details, but also because we can implement it.
- `ggml_graph_comput()` execute node runners one by one, thus we have natural checkpoints between node computing.
  average node computing is fast;  there are hundreds of mul_mat in every graph compute, even if single mul_mat is slow, chances we abort at next node.
- in https://github.com/ggerganov/llama.cpp/pull/1999/, the `ggml_graph_compute_plan`  is passed into `ggml_graph_compute()` and `ggml_graph_compute_thread()`. To notify thread group, we simply add a field `abort` into `ggml_graph_compute_plan`.

CCLDArjun:
I see, `pthread_cancel` seems like a bad idea. Earlier today I changed it so that every thread calls the callback and immediately returns. This might be graceful?

@mqy that's also a good way to do it. Unsure which way to go.

mqy:
> I see, `pthread_cancel` seems like a bad idea. Earlier today I changed it so that every thread calls the callback and immediately returns. This might be graceful?

Me too, sorry, I saw your new commit after submitting comment :)  The callback is graceful.

> Unsure which way to go.

I can't say yes or no, it depends. Anyway, please let me ask some questions:

1. to determine whether abort or not, compute threads have to call an external function with opaque data, this is unnecessary, a bool flag is enough, right?
2. why the new `ggml_graph_compute_with_abort()`? to avoid changing signature of `ggml_graph_compute`?

Overall I think it's not good to create `ggml_graph_compute_with_abort()` which changed the compute implementation even if just a thin wrapper. It should be seen as anti-pattern. Suppose in the future we want to support suspend/resume, should we add ggml_graph_compute_with_suspend and ggml_graph_compute_with_resume?

So, I suggest adding state management APIs, for example: ggml_graph_compute_abort(ctx_or_plan) or more generic ggml_graph_compute_set_state(ctx_or_plan, ABORT).

CCLDArjun:
> I can't say yes or no, it depends. Anyway, please let me ask some questions:
> 
> 1. to determine whether abort or not, compute threads have to call an external function with opaque data, this is unnecessary, a bool flag is enough, right?

I think using a callback for aborting is pretty standard (e.g. [tensorflow/keras](https://www.tensorflow.org/guide/migrate/early_stopping)). It's also easier for a user to pass in a function rather than having to create a separate thread and modifying some shared state to abort. By his comment earlier, I think @smspillaz plans on using a callback, but I'm unsure.

> 2. why the new `ggml_graph_compute_with_abort()`? to avoid changing signature of `ggml_graph_compute`?
> 
> Overall I think it's not good to create `ggml_graph_compute_with_abort()` which changed the compute implementation even if just a thin wrapper. It should be seen as anti-pattern. Suppose in the future we want to support suspend/resume, should we add ggml_graph_compute_with_suspend and ggml_graph_compute_with_resume?
> 
> So, I suggest adding state management APIs, for example: ggml_graph_compute_abort(ctx_or_plan) or more generic ggml_graph_compute_set_state(ctx_or_plan, ABORT).

Agreed! The context is the best place to store it.

CCLDArjun:
We should wait until [ggml: breaking change to ggml_graph_compute(): planning is required before computing llama.cpp#1999](https://github.com/ggerganov/llama.cpp/pull/1999) is merged or just add this feature into it.

mqy:
> I think using a callback for aborting is pretty standard

Sure, callback is more versatile, and convenient to use,  this is why I said `it depends`.

Overall, if we consider conditional abort is important, I think the callback design wins.

ggerganov (approved):
Let me know when to merge 👍

CCLDArjun:
I was waiting for that to be merged so I could add it onto `struct ggml_cplan` but I didn't realize that it had been merged.

Just made the changes, @ggerganov

src/ggml.c
```diff
@@ -15977,6 +15981,10 @@ static thread_ret_t ggml_graph_compute_thread(void * data) {
     int node_n = -1;
 
     while (true) {
+        if (cplan->abort_callback && cplan->abort_callback(cplan->abort_callback_data)) {
+            state->shared->node_n += 1;
```
CCLDArjun, line 15985: I found an issue while testing that sometimes all but one thread aborts and that one thread is spinning, waiting for `shared->node_n` to be updated. So, I decided to modify `node_n`

Issue #341, opened 2023-07-04
ggml : remove `src0` and `src1` from `ggml_tensor` and rename `opt` to `src`


ggerganov:
https://github.com/ggerganov/ggml/blob/b98cd8689f74ed69432323ef5a15369d96086ae1/include/ggml/ggml.h#L415-L420

slaren:
I would suggest considering using an array for all `src` nodes instead. This will simplify the code when we need to look at the list of parents of a node. This already happens when printing to dot, internally in ggml-cuda, and will also happen when we implement the improved memory management. Having each parent in a different variable forces us to look at each one separately instead of just looping over an array.

ggerganov:
I updated the issue

nullhook:
How would the new `src` indexing work in cases where a 4D tensor is created, but `opt[0-2]` is being written? It would be out of bounds. for ex: https://github.com/ggerganov/ggml/blob/master/src/ggml.c#L7268

slaren:
@nullhook I am not sure if I understand what you mean, the total number of source tensors available to ops will be the same, and can be increased as needed.

For instance, see how it is being done in https://github.com/ggerganov/ggml/pull/346:
https://github.com/ggerganov/ggml/blob/b7fd1a4cf637ac747796403c7e2c6209b397c4af/src/ggml.c#L7260-L7266

nullhook:
I was looking at the PR linked to this issue. So it seems like GGML_MAX_OPT needed to be increased

ggerganov:
Fixed via https://github.com/ggerganov/llama.cpp/pull/2178

PR #345, opened 2023-07-05
Remove src0 src1, rename opt to src

rezaarezvan:
Trying to add #341 - I'll open this draft PR while I try to add/fix everything :)

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -16202,7 +16202,7 @@ void ggml_graph_compute(struct ggml_context * ctx, struct ggml_cgraph * cgraph) 
                         if (node->src1->type != vec_dot_type) {
                             cur = GGML_TYPE_SIZE[vec_dot_type]*ggml_nelements(node->src1)/GGML_BLCK_SIZE[vec_dot_type];
                         } else {
-                            GGML_ASSERT(false);
+                            cur = 0;
                         }
 
                         work_size = MAX(work_size, cur);


rezaarezvan:
It compiles, but running tests makes 6 of the tests segfault. Not good, I'll look into this. 

```
The following tests FAILED:
          1 - test-grad0 (SEGFAULT)
          2 - test-opt (SEGFAULT)
          5 - test-mul-mat0 (SEGFAULT)
          8 - test1 (SEGFAULT)
          9 - test2 (SEGFAULT)
         10 - test3 (SEGFAULT)
```

slaren:
You need to fix the indices of what previously were the `opt` tensors, `opt[0]` should be `src[2]`, `opt[1]` -> `src[3]`, etc.

rezaarezvan:
> You need to fix the indices of what previously were the `opt` tensors, `opt[0]` should be `src[2]`, `opt[1]` -> `src[3]`, etc.

Yes, I'm doing that right now :). I'll send that commit soon!

src/ggml.c
```diff
@@ -5144,9 +5144,9 @@ struct ggml_tensor * ggml_acc_impl(
 
     result->op   = GGML_OP_ACC;
     result->grad = is_node ? ggml_dup_tensor(ctx, result) : NULL;
-    result->src0 = a;
-    result->src1 = b;
-    result->opt[0] = c;
+    result->src[0] = a;
+    result->src[1] = b;
+    result->src[0] = c;
```
nullhook, line 5149: why are we re-writing to 0th index here again? perhaps a mistake? This should be src[2]?

nullhook: additionally, do we need a bounds check here?

rezaarezvan: > why are we re-writing to 0th index here again? perhaps a mistake? This should be src[2]?

Yes, I'm fixing all indices as soon as possible :).

rezaarezvan:
Still segfaults so, I've not properly managed to fix this - I'll continue :)

PR #346, opened 2023-07-05
Refactor: Remove src0, src1; Rename opt to src

rezaarezvan:
I'll start a new PR since I got a little lost in the last one :).


rezaarezvan:
With the latest commit, compiling and running tests (`ctest --verbose --timeout 900`), everything passes and should be working now!

src/ggml.c
```diff
@@ -15707,17 +15705,17 @@ static void ggml_visit_parents(struct ggml_cgraph * cgraph, struct ggml_tensor *
...
-        ggml_visit_parents(cgraph, node->src0);
+    if (node->src[0]) {
+        ggml_visit_parents(cgraph, node->src[0]);
     }
 
-    if (node->src1) {
-        ggml_visit_parents(cgraph, node->src1);
+    if (node->src[1]) {
+        ggml_visit_parents(cgraph, node->src[1]);
     }
```
slaren, line 15714: This is already covered by the loop below.

src/ggml.c
```diff
@@ -16567,17 +16565,17 @@ void ggml_graph_export(const struct ggml_cgraph * cgraph, const char * fname) {
...
-                ggml_graph_export_node(cgraph->nodes[i]->src0, "SRC0", fout);
+            if (cgraph->nodes[i]->src[0]) {
+                ggml_graph_export_node(cgraph->nodes[i]->src[0], "SRC0", fout);
             }
 
-            if (cgraph->nodes[i]->src1) {
-                ggml_graph_export_node(cgraph->nodes[i]->src1, "SRC1", fout);
+            if (cgraph->nodes[i]->src[1]) {
+                ggml_graph_export_node(cgraph->nodes[i]->src[1], "SRC1", fout);
             }
```
slaren, line 16574: Same issue here.

src/ggml.c
```diff
@@ -16670,11 +16668,11 @@ void ggml_graph_export(const struct ggml_cgraph * cgraph, const char * fname) {
...
 
-                    args[0] = tensor->src0;
-                    args[1] = tensor->src1;
+                    args[0] = tensor->src[0];
+                    args[1] = tensor->src[1];
 
                     for (int j = 0; j < GGML_MAX_OPT; ++j) {
-                        args[2 + j] = tensor->opt[j];
+                        args[2 + j] = tensor->src[j];
                     }
```
slaren, line 16676: Again this is broken. You need to reconsider all the code that looked at `src0`, `src1` and `opt` and replace it with a single loop that iterates over src.

rezaarezvan: > Again this is broken. You need to reconsider all the code that looked at `src0`, `src1` and `opt` and replace it with a single loop that iterates over `src`.

What's the best approach to fix these kinds, `j = 2`?

rezaarezvan: > > Again this is broken. You need to reconsider all the code that looked at `src0`, `src1` and `opt` and replace it with a single loop that iterates over `src`.
> 
> What's the best approach to fix these kinds, `j = 2`?

Wait I'm dumb, fixing this :)

slaren: I don't think there is a simple solution that you can apply everywhere. You have to consider it on a case-by-case basis. In this code for example, the best solution may be to remove `args` and just use `src` directly.

rezaarezvan:
@slaren - see if these changes makes sense :)

slaren:
That looks better. There is still more refactoring that could be done, especially where `src[0]` and `src[1]` are treated in a different way than `src[2]` and up. There is also some code that assumes that the number of source tensors is `2 + GGML_MAX_OPT` that needs to be fixed:
https://github.com/ggerganov/ggml/blob/b7fd1a4cf637ac747796403c7e2c6209b397c4af/src/ggml.c#L16881-L16884
`GGML_MAX_OPT` should be renamed to something more appropriate. Maybe `GGML_MAX_SRC`.

rezaarezvan:
> That looks better. There is still more refactoring that could be done, especially where `src[0]` and `src[1]` are treated in a different way than `src[2]` and up. There is also some code that assumes that the number of source tensors is `2 + GGML_MAX_OPT` that needs to be fixed:
> 
> https://github.com/ggerganov/ggml/blob/b7fd1a4cf637ac747796403c7e2c6209b397c4af/src/ggml.c#L16881-L16884
> 
> 
> `GGML_MAX_OPT` should be renamed to something more appropriate. Maybe `GGML_MAX_SRC`.

Yeah I was thinking about that - I'll fix these tomorrow!

rezaarezvan:
@slaren I think this commit changes all these. The only thing I want to discuss is, how we should handle the refactoring when dealing with `src[0]` and `src[1]` differently. Should I make a new PR that changes those or in this PR?

slaren:
I think that `ggml-metal.m` also needs to be updated, there are a few references to `src0` and `src1` there.

>Should I make a new PR that changes those or in this PR?

Either way would work for me, let's wait for @ggerganov review.

rezaarezvan:
> I think that `ggml-metal.m` also needs to be updated, there are a few references to `src0` and `src1` there.
> 
> > Should I make a new PR that changes those or in this PR?
> 
> Either way would work for me, let's wait for @ggerganov review.

I'll update those tomorrow

ggerganov:
Hey @rezaarezvan - I just saw this PR after merging the other one: https://github.com/ggerganov/llama.cpp/pull/2178

Thanks for the effort and apologise for not upstreaming this!

PR #351, opened 2023-07-06
Correct Alibi implementation

daulet:
The current implementation doesn't reflect intended description of Alibi. Here is [reference implementation](https://github.com/ofirpress/attention_with_linear_biases/blob/master/fairseq/models/transformer.py#L742-L762) linked in [Papers with Code](https://paperswithcode.com/method/alibi). This was discovered when using ggml on a model trained with regular definition of Alibi.

MPT and Replit models appear to be using this implementation, but I don't know if they've been verified to produce correct completions. Alternatively I can add this as `alibi_classic` op to avoid breaking those models.

Tests: I'd love to add them but it wasn't clear where the fit in the repo.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -11784,7 +11784,7 @@ static void ggml_compute_forward_alibi_f16(
 
     const int ne0 = src0->ne[0]; // all_seq_len = n_past + ne1
     const int ne1 = src0->ne[1]; // seq_len_without_past
-    //const int ne2 = src0->ne[2]; // n_head -> this is k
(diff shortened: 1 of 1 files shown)


ggerganov:
Hi, thanks for finding this. I ran a few tests with MPT and it seems to generate coherent text.

It would be nice to add some tests for that, but we don't have a good way of doing that yet.
We should somehow compare the results between the `ggml` implementation and the reference Python implementations

Open to suggestions

ggerganov (approved):

PR #358, opened 2023-07-08
Fast inverse square root to GPT-2 dot product scaling

mchtilianov:
A couple of days ago I had a thought of optimizing the dot product scaling in the attention mechanism. I'm getting some performance benefits on the GPT-2 117M and 345M models. On the 117M I got 18ms/token down to 15ms/token. On the 345M models I got 46ms/token down to 43ms/token. I'm doing this on a 6 year old laptop, so if someone can also check whether they're getting similar performance gains I'd be very grateful. If nobody sees any issues with this implementation and it gets merged I'll add it to the rest of the models and in llama.cpp.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -18220,6 +18220,16 @@ size_t ggml_quantize_chunk(enum ggml_type type, const float * src, void * dst, i
     }
     return result;
 }
+////////////////////////////////////////////////////////////////////////////////
+
+float ggml_fast_inv_sqrt(float a){
+        float xhalf = 0.5f * a;
+        int i = *(int*)&a;            // store floating-point bits in integer
+        i = 0x5f3759df - (i >> 1);    // initial guess for Newton's method
+        a = *(float*)&i;              // convert new bits into float
+        a = a*(1.5f - xhalf*a*a);     // One round of Newton's method
+        return a;
+    }
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/examples/gpt-2/main.cpp
+++ b/examples/gpt-2/main.cpp
@@ -534,7 +534,7 @@ bool gpt2_eval(
             struct ggml_tensor * KQ_scaled =
                 ggml_scale_inplace(ctx0,
                         KQ,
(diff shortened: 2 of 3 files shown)


mchtilianov:
I forgot to mention it, but you can check out how the fast inverse square root algorithm works [here](https://betterexplained.com/articles/understanding-quakes-fast-inverse-square-root/).

src/ggml.c
```diff
@@ -18220,6 +18220,16 @@ size_t ggml_quantize_chunk(enum ggml_type type, const float * src, void * dst, i
...
+////////////////////////////////////////////////////////////////////////////////
+
+float ggml_fast_inv_sqrt(float a){
+        float xhalf = 0.5f * a;
+        int i = *(int*)&a;            // store floating-point bits in integer
+        i = 0x5f3759df - (i >> 1);    // initial guess for Newton's method
+        a = *(float*)&i;              // convert new bits into float
+        a = a*(1.5f - xhalf*a*a);     // One round of Newton's method
+        return a;
+    }
```
Green-Sky, line 18232: formatting is of

Green-Sky: you forgot the curse word

Green-Sky (commented):
try not putting the declaration in the .h and mark the function as static, in most cases that will lead to inlining (should improve perf further, if it not already did)

slaren:
Note that this is a constant that only has to be calculated once per model. Instead of doing this, you could remember the value between invocations to `gpt2_eval`.
Additionally, keep in mind that current CPUs have inverse square root instructions that may outperform this. Here is a recent article about this topic: https://hllmn.net/blog/2023-04-20_rsqrt/

mchtilianov:
> Note that this is a constant that only has to be calculated once per model. Instead of doing this, you could remember the value between invocations to `gpt2_eval`. Additionally, keep in mind that current CPUs have inverse square root instructions that may outperform this. Here is a recent article about this topic: https://hllmn.net/blog/2023-04-20_rsqrt/

Sounds great. I'll start reworking it.

PR #359, opened 2023-07-09
Support broadcast addition for fp32

li-plus:
This PR supported broadcast addition for two fp32 tensors as long as they are broadcast-able in row dimension. No longer requires same shape. Implemented on both CPU and CUDA.

### DIFF
--- a/src/ggml-cuda.cu
+++ b/src/ggml-cuda.cu
@@ -238,13 +238,13 @@ struct ggml_tensor_extra_gpu {
     cudaEvent_t events[GGML_CUDA_MAX_DEVICES]; // events for synchronizing multiple GPUs
 };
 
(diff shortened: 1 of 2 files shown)


slaren:
For CUDA, it may be better to do this with a single kernel launch, similar to the way broadcasting is implemented for mul. @JohannesGaessler what do you think?

JohannesGaessler:
Each CUDA kernel launch adds overhead. It is therefore preferable to launch as few kernels as reasonably possible. So I agree with slaren that an implementation that does the broadcasting with a single kernel launch would be preferable.

li-plus:
Thanks for your advice! You are correct. Now I use a single kernel to compute broadcast add. This leads to huge performance improvement in linear layers.

PR #363, opened 2023-07-09
Fix for starcoder ggml-model.bin being saved in wrong directory. Moderrnise by using argparse.

TheBloke:
Currently the starcoder `convert-hf-to-ggml.py` script saves the resulting GGML file in a bizarre directory when the user specifies a local directory for the model source.  eg if I do `convert-hf-to-ggml.py /workspace/models/starcoder/source` I will get `/workspace/models/starcoder/models/workspace/models/starcoder/source-ggml.bin`

I have fixed this by moving to using `argparse`, and more closely matching the parameters of llama.cpp convert.py, with an `--outfile <path to desired model-bin>` parameter so the user can just specify the exact output filename and location they want.  Default filename if not specified is `ggml-model.bin` in the location the user is when they run the script.  

Whether to use fp32 or not is now a boolean `--use_f32` rather than being decided by the number of params specified.

I think it would be good if all the GGML example scripts matched this format or something similar.  I don't have time to do all the other scripts this second but could do shortly.

Let me know what you think of this and if you're happy I will update the other scripts to the same style, and maybe move this arg parsing and model downloading/loading to a central utils.py that gets loaded by all the scripts.

### DIFF
--- a/examples/starcoder/convert-hf-to-ggml.py
+++ b/examples/starcoder/convert-hf-to-ggml.py
@@ -46,6 +46,7 @@ args = parser.parse_args()
 use_f16 = not args.use_f32
 
 fname_out = args.outfile
+os.makedirs(os.path.dirname(fname_out), exist_ok=True)
 
 print("Loading model: ", args.model_name_or_path)
 tokenizer = AutoTokenizer.from_pretrained(args.model_name_or_path)
--- a/examples/starcoder/convert-hf-to-ggml.py
+++ b/examples/starcoder/convert-hf-to-ggml.py
@@ -8,6 +8,7 @@ import torch
 import numpy as np
 import re
 import os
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
> Let me know what you think of this and if you're happy I will update the other scripts to the same style, and maybe move this arg parsing and model downloading/loading to a central utils.py that gets loaded by all the scripts.

Anything that makes the scripts consistent with each other is great.
I don't have strong preference for the Python code. As long as it is simple and consistent - it's great

TheBloke:
Oh sorry Georgi, I didn't notice you'd already merged this.  I don't seem to be getting Github notifications correctly.

Thanks for the feedback. I will do another PR for the changes I described to bring this to all the scripts.

PR #384, opened 2023-07-13
Support dup & cont ops on CUDA

li-plus:


ggerganov:
Let's put this change in `llama.cpp` so we can test if it breaks something. I will sync it back to this repo when we merge it there

Thanks for all the contributions!

li-plus:
Closing this PR since it's been synced from llama.cpp

PR #394, opened 2023-07-17
mnist: add model conversion command for the example programs

BWbwchen:


ggerganov:
Thanks, I just pushed a fix for this

PR #401, opened 2023-07-20
Add the dense image positional encoding in SAM state

YavorGIvanov:
Now we have all the input state for the mask decoder.


ggerganov (approved):
:rocket:

Issue #397, opened 2023-07-18
Metal support


matthoffner:
Hello! I was curious if anyone has gotten models like MPT and Starcoder to work with GGML and the M1 specifically using Metal/GPU. Thanks.

mateusz:
Yeah not implemented in the examples - came here looking to run WizardCoder on MPS (metal) on mac, but no dice. WizardCoder has a different architecture than Llama, and I haven't found any MPS implementation yet - if you do let me know :)

matthoffner:
The linked MR shows that we can now pass `-DGGML_METAL=on`. I'm assuming specific work is required to support individual models still. 

I did end up getting WizardCoder and Metal to work with mlc-llm, but ggml has been plenty fast on cpu for me.

mateusz:
Yes, you can compile -DGGML_METAL=on, it does link the files, but the use of it is not implemented in the "userland" part (the example).

Oh interesting, what results are you getting on mlc-llm vs on CPU?

Re "plenty fast", the speedup of metal is significant:

```
bin/main -m ../../models/wizardLM-7B.ggmlv3.q4_0.bin -n 128 -ngl 0 --ignore-eos --mlock -t 4 -s 42 -n 256 -p "Llama is faster when "
...
llama_print_timings:        eval time = 14095.91 ms /   255 runs   (   55.28 ms per token,    18.09 tokens per second)
```
vs on metal
```
bin/main -m ../../models/wizardLM-7B.ggmlv3.q4_0.bin -n 128 -ngl 1 --ignore-eos --mlock -t 4 -s 42 -n 256 -p "Llama is faster when "
...
llama_print_timings:        eval time =  7464.27 ms /   255 runs   (   29.27 ms per token,    34.16 tokens per second)
```

PR #403, opened 2023-07-20
Add metal kernel to target resources when built as shared library

abetlen:
Adds the option to compile ggml with the metal source files similar to how it's done in llama.cpp. Should help in the implementation of #397


ggerganov:
Oops, I just merged https://github.com/ggerganov/ggml/pull/407

Let me know if it works

abetlen:
@ggerganov almost, I also need the metal kernel to be included with the resources so it's installed along with the shared library.

ggerganov (approved):

PR #405, opened 2023-07-21
Refactor unary ops in ggml

### LINKED ISSUE #285: ggml : reduce the values of `ggml_op` enum

Ivan-Z:
Resolves #285. Removes duplicated code from unary op constructors to a shared ggml_unary. No changes to usage.


src/ggml.c
```diff
@@ -14376,6 +14200,66 @@ static void ggml_compute_forward_win_unpart(
     }
 }
 
+//gmml_compute_forward_unary
```
ggerganov, line 14203:
```suggestion
// ggml_compute_forward_unary
```

src/ggml.c
```diff
@@ -7538,6 +7322,47 @@ struct ggml_tensor * ggml_win_unpart(
...
+    bool is_node = false;
+
+    if (!inplace && (a->grad)) {
+        is_node = true;
+    }
+
+    struct ggml_tensor * result = inplace ? ggml_view_tensor(ctx, a) : ggml_dup_tensor(ctx, a);
+
+    struct ggml_tensor * b = ggml_new_tensor_1d(ctx, GGML_TYPE_F32, 1);
+    ((int *) b->data)[0] = op;
```
ggerganov, line 7341: The parameter passing was just updated on `master`.
Try to utilize the new `op_params` array to pass this value

src/ggml.c
```diff
@@ -14376,6 +14200,66 @@ static void ggml_compute_forward_win_unpart(
...
+            {
+                ggml_compute_forward_relu(params, src0, dst);
+            } break;
+        case GGML_UNARY_OP_GELU:
+            {
+               GGML_ASSERT(false); // TODO: not implemented
+            } break;
+        case GGML_UNARY_OP_GELU_QUICK:
+            {
+               GGML_ASSERT(false); // TODO: not implemented
```
ggerganov, line 14250: Why are `GELU` and `GELU_QUICK` not implemented?

src/ggml.c
```diff
@@ -15967,6 +15748,80 @@ static void ggml_compute_backward(struct ggml_context * ctx, struct ggml_tensor
             } break;
         case GGML_OP_WIN_PART:
         case GGML_OP_WIN_UNPART:
+        case GGML_OP_UNARY:
+            {
+                switch((enum ggml_unary_op)tensor->src[0]) {
```
ggerganov, line 15753: This should be `tensor->src[1]` I think

src/ggml.c
```diff
@@ -16459,21 +16314,34 @@ struct ggml_cplan ggml_graph_plan(struct ggml_cgraph * cgraph, int n_threads) {
...
                     n_tasks = 1;
                 } break;
-            case GGML_OP_MUL:
-            case GGML_OP_GELU:
-            case GGML_OP_GELU_QUICK:
-            case GGML_OP_SILU:
+
+            case GGML_OP_UNARY:
+                {
+                    switch ((enum ggml_unary_op)node->src[0]) {
```
ggerganov, line 16323: Should be `node->src[1]`

ggerganov (changes requested):
Looks good, but a few things need fixes.

I'll push a change now to finalize this

ggerganov (approved):
What is missing now is to figure out a neat way to print and dump the names of the unary ops, because now it will simply print `unary` which is not very useful for debugging purposes.

@slaren Do you think this change is OK?

src/ggml.c
```diff
@@ -3811,7 +3803,7 @@ static const char * GGML_OP_NAME[GGML_OP_COUNT] = {
     "CROSS_ENTROPY_LOSS_BACK",
 };
 
-static_assert(GGML_OP_COUNT == 68, "GGML_OP_COUNT != 68");
+static_assert(GGML_OP_COUNT == 59, "GGML_OP_COUNT != 68");
```
slaren, line 3806: The string needs to be updated as well.

src/ggml.c
```diff
@@ -4958,6 +4961,14 @@ float * ggml_get_data_f32(const struct ggml_tensor * tensor) {
     return (float *)(tensor->data);
 }
 
+enum ggml_unary_op ggml_get_unary_op(const struct ggml_tensor * tensor) {
+    return (enum ggml_unary_op) ggml_get_op_params_i32(tensor, 0);
+}
+
+void ggml_set_unary_op(struct ggml_tensor * tensor, enum ggml_unary_op op) {
+    ggml_set_op_params_i32(tensor, 0, (int32_t) op);
+}
```
slaren, line 4970: A check such as `GGML_ASSERT(tensor->op == GGML_OP_UNARY)` would be good here.

slaren: I am not sure if we really need `ggml_set_unary_op`, though.

slaren:
A function that returns the name of a unary op would be useful. Maybe also one that returns either the name of the op or the name of the unary. Something like this:
```C
if (op == GGML_OP_UNARY) {
    return ggml_unary_op_name(t);
}
return ggml_op_name(t);
```

Ivan-Z:
Thanks for the help @ggerganov @slaren. Do we want to make a similar change for binary ops? There are only a few right now, but might be worth making the change now if ggml intends on supporting more in the future.

ggerganov:
Let's use the unary API for a while and decide later depending on if we like the API or not

PR #407, opened 2023-07-22
Sync metal compile scripts from llama.cpp

li-plus:
Add metal option in CMakeLists.txt here so that we can compile MPS backend depending only on this repo. This is synced from llama.cpp


ggerganov (approved):

ggerganov:
Thanks!

PR #409, opened 2023-07-23
ggml : sync llama.cpp

ggerganov:
- CUDA improvements
- Metal improvements
- `ggml` refactorings

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -1,4 +1,15 @@
 #/bin/bash
+#
+# sample usage:
+#
+# mkdir tmp
+#
+# # CPU-only build
+# bash ./ci/run.sh ./tmp/results ./tmp/mnt
+#
+# # with CUDA support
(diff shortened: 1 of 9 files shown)


slaren:
I think that `ggml_graph_export` is broken after the change to `op_params`, the field isn't exported or imported now.

ggerganov:
Ah yes, I forgot about this.

> I think I could export it as a 32 byte hex string, does that sound good?

The `ggml_graph_export_node()` is named misleading - it just prints stuff to `stdout`. It does not actually export data.

We just need to write the binary data for each tensor (leaf and node). Maybe right after the `name`.
You can push it here and in the meantime I will add a CI test for this functionality

slaren:
Ah, I see, I thought I would have to export it as ASCII, that will be easier. I don't have collaborator status in this repo to push the changes here, though.

PR #410, opened 2023-07-23
Fix gpt-neox docs

tombailey:
The files/directories referenced by the docs are kebab-case not snake_case.


ggerganov (approved):
TIL "kebab-case" 😄

PR #418, opened 2023-07-26
Add SAM decoder & output masks as png

YavorGIvanov:
With this additions, the SAM decoder and therefore the whole SAM ggml implementation for point input is almost finished. The things left are:

- [x] - Implement the output_upscaling part for which we need ggml version of PyTorch nn.ConvTranspose2d() (ref: https://github.com/PABannier/encodec.cpp/blob/main/ggml.c#L12389-L12497)
- [x] - Implement the final output of the model

ЕDIT: Also there is a difference of part of the elements of some decoder results compared to the PyTorch model. The difference seems to be due to accumulating floating point error or due to the linear interpolation result numeric difference when preprocessing the image in the beggining.

### DIFF
--- a/examples/sam/convert-pth-to-ggml.py
+++ b/examples/sam/convert-pth-to-ggml.py
@@ -71,11 +71,6 @@ for k, v in model.items():
     name = k
     shape = v.shape
 
-    # TODO: export only the Encoder -- after it works we will export the other stuff
-    if name[:13] != "image_encoder" and \
-       name[:14] != "prompt_encoder":
-        continue
-
     if name[:19] == "prompt_encoder.mask":
         continue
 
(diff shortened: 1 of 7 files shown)


YavorGIvanov:
The initial version is finished as the GGML version outputs masks close to the ones the PyTorch model outputs for hardcoded point inputs. For the point harcoded in the code (414.375, 162.796875) and this image:
 
img

the PyTorch model outputs this mask:


and the GGML version outputs this:
mask_out_2

There are multiple things to be finished that I will begin doing now. Here are some with high priority:
- Trace where the difference in output masks comes from. This will be done by going through the inference tensors step by step and comparing to the PyTorch version 
- Filter masks based on stability score and based on boxes, which touch crop boundaries
- Optimize the GGML implementation by removing not needed ggml_cont operations/reshapes/transposes and so on
- Add support for user input

ggerganov (approved):
:llama:

Issue #294, opened 2023-06-25
ggml : improve custom operator support


ggerganov:
We have rudimentary support for adding F32 custom `ggml` operators in user space via:

https://github.com/ggerganov/ggml/blob/a1d0ea7c2abd44f56822ffdfcfe0a0fcf7170885/include/ggml/ggml.h#L1173-L1240

The API has to be improved to support:

- multi-threading
- non-F32 types

We also need to provide a few examples of how to define custom operators.

This functionality will significantly improve adoption of `ggml` in third-party projects since currently, they need to rely that certain operators first get implemented into `ggml` before being able to use them

slaren:
Looks like the `custom1-3` versions already support any data type, as they pass the tensor to the function, and the types are never checked. So we could simply remove the `f32` suffix from the names.

To support multi-threading, we could add a `n_tasks` parameter to the op, and `ith`, `nth` parameters to the callback, such as:
```C
typedef void (*ggml_custom1_op_t)(struct ggml_tensor * dst , const struct ggml_tensor * a, int ith, int nth); 

GGML_API struct ggml_tensor * ggml_map_custom1( 
        struct ggml_context          * ctx, 
        struct ggml_tensor           * a, 
               int                     n_tasks, // clamped to the number of threads
               ggml_custom1_op_t       fun); 
```

Would a test work as an example?

ggerganov:
> and ith, nth parameters to the callback

Maybe we can pass the entire `struct ggml_compute_params` to the callback?

slaren:
My view on that is that we have are already exposing way too many details of the internals of ggml in the public API (ie. `ggml.h`), and that's already slowing us down because it makes changing certain aspects of ggml really hard because it is too easy to break user code. Some of that user code is in our own examples. At some point we are going to have to define a public API and commit to supporting it in the long term (or at least until the next major release), but until that happens we shouldn't make the problem harder for ourselves.

In this case, `ggml_compute_params` has the `ggml_task_type` and the work buffer. I could see the task type being useful to allow users to have their own `INIT` and `FINALIZE` passes, but I am not sure that it is worth exposing that complexity to the user. The work buffer shouldn't be exposed to the user, it would require adding another interface to allow the users to specify how much work buffer they need, and they can simply allocate their own buffers if they need to anyway.

We should probably add a `void *` parameter that is passed to the callback, so that the users can use it to pass any custom data that they may need in the callback:
```C
[...]

PR #422, opened 2023-07-30
ggml : add ggml_map_custom1-3

slaren:
Fixes https://github.com/ggerganov/ggml/issues/294

- Adds `ggml_map_custom1-3` and `_inplace` versions
- Marks `ggml_map_unary_f32`, `ggml_map_binary_f32`, `ggml_map_custom1-3_f32` as deprecated
- Adds `tests/test-customop.c` to test these functions and hopefully also serve as an example

Additional changes:
- Removes `GGML_TYPE_F32` type checks from `ggml_map_custom1-3_f32`
- Makes `ggml_are_same_shape` public
- Adds `.editorconfig` from `llama.cpp`
- Adds macro `GGML_DEPRECATED` to `ggml.h` (from `llama.h`)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# https://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file, utf-8 charset
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+
+[prompts/*.txt]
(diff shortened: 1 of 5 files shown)


slaren:
Possible issue: the data type of the destination tensor is copied from the data type of the first parameter. In some cases, users may want to use a different data type.

ggerganov (approved):

ggerganov:
Sure, adding type argument sounds good. But even as it is it should work for most use cases

slaren:
I guess it is better to test this for a while and make sure that it works for the users. Adding another version of the ops that adds a parameter to choose the output type would be trivial and can be done later, and I think that would be better than adding even more parameter to these functions.

PR #433, opened 2023-08-04
ggml: Backport ggml-alloc from llama.cpp

smspillaz:
Backports a113689571420fb4d6540f1a324d12965781356a from llama.cpp


smspillaz:
OK this is ready now. Last two commits fix a bug in the allocator that I noticed - I'll submit it upstream too.

smspillaz:
I incorporated the upstream `>=` fix

slaren (approved):

PR #438, opened 2023-08-07
ggml : sync llama.cpp (memory allocator + cuda  + metal)

ggerganov:


src/ggml.c
```diff
@@ -4110,7 +4110,7 @@ size_t ggml_nbytes(const struct ggml_tensor * tensor) {
     //
     // is enough, but just in case, adding the second part
 
-    return MAX(tensor->ne[3]*tensor->nb[3], (ggml_nelements(tensor)*GGML_TYPE_SIZE[tensor->type])/GGML_BLCK_SIZE[tensor->type]);
+    return GGML_PAD(MAX(tensor->ne[3]*tensor->nb[3], (ggml_nelements(tensor)*GGML_TYPE_SIZE[tensor->type])/GGML_BLCK_SIZE[tensor->type]), GGML_MEM_ALIGN);
```
ggerganov, line 4113: @slaren 

What do you think about this change?

Without this, the `mnist` example fails because when we export the graph, we compute an incorrect size of the tensor data here:

https://github.com/ggerganov/ggml/blob/a30107764ca5544e3a1ead4b318e06d83ed5b14c/src/ggml.c#L16837-L16842

slaren: It should be fine currently, I am a bit concerned that this may cause out of bounds accesses when using `ggml_nbytes` to copy all the data of a tensor. Shouldn't be an issue currently, but it may introduce bugs in the future. If we need to guarantee that the full `ggml_nbytes` of a tensor data are accessible, then maybe `ggml_new_tensor_impl` should pad the `data_size` explicitly rather than depending on `ggml_new_object` to pad the whole object with the assumption that both `GGML_OBJECT_SIZE` and `GGML_TENSOR_SIZE` are padded to the same alignment.

Issue #441, opened 2023-08-09
Support for xPos positional embedding


jploski:
[xPos](https://arxiv.org/pdf/2212.10554.pdf) is an improved version of the original RoPE from the [RoFormer](https://arxiv.org/pdf/2104.09864.pdf) paper (i.e. a modification of ggml_rope with !is_neox flag). I'm unaware of published models using it yet, but it is important because it's the positional embedding employed by the [RetNet](https://arxiv.org/abs/2307.08621) paper (which offers support for O(1) inference, i.e. independent from context length, with non-degraded quality; a potential superior replacement for attention with kv caches).

For a quick Python comparison of the original RoPE, GPT-NeoX RoPE and xPos RoPE see: https://github.com/jploski/RotaryEmbedding

Based on the above I will create a pull request with an initial (non-CUDA) implementation.

jploski:
PR: https://github.com/ggerganov/ggml/pull/442

YavorGIvanov:
Isn't this done ? Please, reopen if I am wrong.

PR #442, opened 2023-08-09
ggml : implementation of xPos RoPE (#441); also extends ggml_rope_bac…

jploski:
This is my implementation of xPos RoPE - based on an implementation provided by https://github.com/syncdoth/RetNet (which differs slightly from Microsoft's torchscale implementation to support iterative/recurrent use).

Note that I decided to extend the existing !is_neox implementation by adding the required parameters, but I did not add a new flag for it (the combination of !is_neox and xpos_scale > 0 is used to distinguish between original RoPE and xPos; for the latter an additional API function ggml_rope_xpos_inplace is provided).

I also extended ggml_rope_back with xPos-specific parameters as well as with the two freq parameters that were already added to the forward version of ggml_rope_custom, but were missing in the backprop.

Code review and help with the missing CUDA port appreciated.

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1207,6 +1207,15 @@ extern "C" {
             float                 freq_base,
             float                 freq_scale);
 
+    // xPos RoPE, in-place, returns view(a)
+    GGML_API struct ggml_tensor * ggml_rope_xpos_inplace(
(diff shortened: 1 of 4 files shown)


ggerganov (approved):
Thanks!

Issue #436, opened 2023-08-06
Cannot load teknium/Replit-v2-CodeInstruct-3B


shrikrishnaholla:
This issue is on similar lines as https://github.com/ggerganov/ggml/issues/248 , but is regarding replit-v2 models, not replit-v1

I am using ggml@ a30107764ca5544e3a1ead4b318e06d83ed5b14c and am having trouble loading [`teknium/Replit-v2-CodeInstruct-3B`](https://huggingface.co/teknium/Replit-v2-CodeInstruct-3B)

I used `examples/replit/convert-h5-to-ggml.py` to convert to ggml `f32`.
Also created both a `q4_1` as well as `q8_0` quantized versions using `replit-quantize`.

However, when trying to load either f32, q4_1 or q8_0 versions of the models with replit
(e.g., `./bin/replit -m Replit-v2-CodeInstruct-3B-f32.bin -p "def hello_world():"`)
I get:

```
replit_model_load: unknown tensor 'transformer.blocks.0.norm_1.weight' in model file
```

Any ideas?

klosax:
The model loads fine for me. The named tensor should be recognized and loaded. Did you get any compilation warnings?

klosax:
> Could this have anything to do with it?

I guess not if the model file was converted successfully.

Any compilation warnings when compling the inference binary?

shrikrishnaholla:
Nothing stood out to me in particular...

```bash
cmake .. && make -j4 replit replit-quantize
-- CMAKE_SYSTEM_PROCESSOR: x86_64
-- x86 detected
-- Linux detected
-- x86 detected
-- Linux detected
-- Configuring done (0.1s)
-- Generating done (0.3s)
-- Build files have been written to: ~/ggml/build
[ 25%] Building CXX object examples/CMakeFiles/common.dir/common.cpp.o
[ 25%] Building C object src/CMakeFiles/ggml.dir/ggml.c.o
In file included from /usr/include/string.h:535,
                 from ~/ggml/src/ggml.c:21:
In function ‘memcpy’,
    inlined from ‘ggml_set_op_params’ at ~/ggml/src/ggml.c:4642:5,
    inlined from ‘ggml_conv_1d’ at ~/ggml/src/ggml.c:6883:5:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:10: warning: ‘__builtin_memcpy’ offset [0, 11] is out of the bounds [0, 0] [-Warray-bounds]
[...]

PR #443, opened 2023-08-10
Fix examples tensor name access

klosax:
Ref https://github.com/ggerganov/ggml/issues/436

Fixes access of `std::string` that do not work in gcc 11.


slaren (approved):

Green-Sky (approved):
good pr

sidenote: .data() and .c_str() are defined to return the exact same for std::string

PR #458, opened 2023-08-18
fix starcoder/convert-hf-to-ggml.py to handle current-folder output file - as in README example

phlogisticfugu:
fix from bug introduced in https://github.com/ggerganov/ggml/pull/363

for when there is no folder specified in the output file, as is the default

this makes the example code in the README work as-is.  otherwise, there is an error when the script attempts to create folders for an empty string


phlogisticfugu:
to clarify, it is the [Quickstart example](https://github.com/ggerganov/ggml/tree/master/examples/starcoder#quick-start) that is currently broken

klosax (approved):

PR #464, opened 2023-08-19
Added MPT as a library example

OlegJakushkin:
Having some commercially sellable network use as a library example is important. A library that could be used in and from other languages. This PR focuses on adding such an example.

## Changes to be made:
 - [x] Add a C++ wrapper class based on MPT example
 - [x] Main updates to cede are prints turned into C++ and redirected from the console to the API class  
 - [x] Check that it builds for Windows, Mac, Linux
 - [x] Added SWIG .i files and generated wrapper examples for direct building ad CICD

#### What types of changes does your code introduce?
- [x] New feature (a non-breaking change that adds functionality)

## Remarks
Tested on CSharp


monatis:
We already have the [MPT example](https://github.com/ggerganov/ggml/tree/master/examples/mpt) and I believe a library wrapper around it is out of scope of this repo. Examples are meant to serve simple usage examples of the GGML library for inference of various ML models and anything more complex than this should leave in user repos.

Green-Sky:
+ since llama.cpp is supposed to gain mpt support and is the better inference library, it would more sense there.

PR #485, opened 2023-08-27
Add MNIST example with CNN

rgerganov:
Add one more implementation for MNIST which uses Conv2D layers, ref: https://keras.io/examples/vision/mnist_convnet/
It achieves ~99% accuracy on the MNIST test set and also performs better for user inputs.

### DIFF
--- a/examples/mnist/main-cnn.cpp
+++ b/examples/mnist/main-cnn.cpp
@@ -1,0 +1,169 @@
+#include "ggml/ggml.h"
+
+#include "common.h"
+
+#include <cmath>
+#include <cstdio>
+#include <cstring>
+#include <ctime>
+#include <fstream>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+#if defined(_MSC_VER)
+#pragma warning(disable: 4244 4267) // possible loss of data
+#endif
+
+struct mnist_model {
+    struct ggml_tensor * conv2d_1_kernel;
+    struct ggml_tensor * conv2d_1_bias;
+    struct ggml_tensor * conv2d_2_kernel;
+    struct ggml_tensor * conv2d_2_bias;
+    struct ggml_tensor * dense_weight;
+    struct ggml_tensor * dense_bias;
+    struct ggml_context * ctx;
+};
+
+bool mnist_model_load(const std::string & fname, mnist_model & model) {
+    struct gguf_init_params params = {
+        /*.no_alloc   =*/ false,
+        /*.ctx        =*/ &model.ctx,
+    };
+    gguf_context * ctx = gguf_init_from_file(fname.c_str(), params);
+    if (!ctx) {
+        fprintf(stderr, "%s: gguf_init_from_file() failed\n", __func__);
+        return false;
+    }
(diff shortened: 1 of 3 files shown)


rgerganov:
Right now I am using raw files for the model. I need to come up with a Python script which saves them to GGUF. Hence this is a draft.

examples/mnist/main-cnn.cpp
```diff
@@ -0,0 +1,250 @@
...
+    const int ne = bias->ne[2];
+    std::vector<float> bias_f32(ne);
+    fin.read(reinterpret_cast<char *>(bias_f32.data()), ne*sizeof(float));
+    fin.close();
+    // TODO: is there a better way to convert 1d bias to 4d bias?
+    for (int i = 0; i < ne; i++) {
+        for (int j = 0; j < ne0*ne1; j++) {
+            ggml_set_f32_1d(bias, i*ne0*ne1 + j, bias_f32[i]);
+        }
+    }
```
ggerganov, line 62: We probably need to add 4d overloads for `ggml_set_...` but for now this should be fine

ggerganov (approved):

ggerganov:
> Right now I am using raw files for the model. I need to come up with a Python script which saves them to GGUF. Hence this is a draft.

Adding GGUF conversion could be a useful example, but it's also OK to keep reading `.raw` files.

rgerganov:
Two observations on this simple example:
 * the `conv2d` operator in GGML expects the same kernel shape as the Conv2D layer in Keras (e.g. (3, 3, 1, 32) for the first one), but its dimensions are reordered. So I am using `np.moveaxis()` in the python code before exporting the convolution kernels.
 * The python `gguf` module is saving the tensor shape in reverse order. Maybe this is useful for the big models like llama but here it is not needed. I am using the `raw_shape` parameter as a workaround here.

ggerganov:
I tested this and everything works correctly. Here is the graph plot:

mnist-cnn dot

> the conv2d operator in GGML expects the same kernel shape as the Conv2D layer in Keras (e.g. (3, 3, 1, 32) for the first one), but its dimensions are reordered. So I am using np.moveaxis() in the python code before exporting the convolution kernels.

Yes, it's always difficult to match the shapes during such conversions. I guess there are conventions in standard frameworks but `ggml` is not following them and this creates difficulties. Having such kind of examples is very useful because it is easy to debug and understand where is the mismatch

> The python gguf module is saving the tensor shape in reverse order. Maybe this is useful for the big models like llama but here it is not needed. I am using the raw_shape parameter as a workaround here.

It's not related to LLMs. The convention for referencing the axes in the tensor is different between Python and `ggml`:

```python
# example 4d tensor in Python
print(x.shape)
(32, 1, 3, 3)

print(x.shape[0], kernel1.shape[1], kernel1.shape[2], kernel1.shape[3])
32 1 3 3
```

```c
// same tensor in ggml
printf("(%d, %d, %d, %d)\n"), x->ne[0], x->ne[1], x->ne[2], x->ne[3]);
(3, 3, 1, 32)
```

So `gguf.py` reverses the dims order to match the `ggml` order.

Shall we merge this or add a README.md first?

rgerganov:
I will update the README and the web app in a follow-up PR

PR #491, opened 2023-08-28
ggml : fix ggml_nbytes with views, add bounds checks to views

slaren:
Changes the implementation of `ggml_nbytes` to generally work with any layout, and adds checks to `ggml_view_1d-4d` to ensure that the view doesn't cause out-of-bounds access on the parent.

I am not entirely sure about this, it may be OK to allow unchecked views, for example when creating a view of a view such as the first view is smaller than the original tensor, and the second view is larger than the first view, but still smaller than the original tensor. It seems a bit convoluted, though. We would be able to check these things more accurately if we had a `view_source` pointer in `ggml_tensor`, and that could also solve some problems with `ggml-alloc`.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -4102,16 +4102,13 @@ int64_t ggml_nrows(const struct ggml_tensor * tensor) {
 }
 
 size_t ggml_nbytes(const struct ggml_tensor * tensor) {
-    static_assert(GGML_MAX_DIMS == 4, "GGML_MAX_DIMS is not 4 - update this function");
+    size_t nbytes = tensor->ne[0]*tensor->nb[0]/ggml_blck_size(tensor->type);
 
-    // this should handle cases where the tensor is not contiguous in memory
-    // probaby just:
-    //
-    //     return tensor->ne[3]*tensor->nb[3]
-    //
-    // is enough, but just in case, adding the second part
+    for (int i = 1; i < GGML_MAX_DIMS; ++i) {
+        nbytes += (tensor->ne[i] - 1)*tensor->nb[i];
+    }
 
-    return MAX(tensor->ne[3]*tensor->nb[3], (ggml_nelements(tensor)*ggml_type_size(tensor->type))/ggml_blck_size(tensor->type));
+    return nbytes;
 }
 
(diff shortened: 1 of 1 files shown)


ggerganov:
>  We would be able to check these things more accurately if we had a view_source pointer in ggml_tensor, and that could also solve some problems with ggml-alloc.

Since `ggml-alloc` is looking like a great improvement and is here to stay, we can extend `ggml_tensor` with new stuff that would help `ggml-alloc`

slaren:
I think this could also be solved by changing `ggml_view_tensor` to return a `GGML_OP_VIEW`, with the parent in `src[0]`. @ggerganov is there any reason why `ggml_view_tensor` returns a `GGML_OP_NONE`?

ggerganov:
`ggml_view_tensor` is used mostly internally by each inplace operator in `ggml.c` like this:

https://github.com/ggerganov/ggml/blob/69bf842b39e6a53218eda35852a91d395d357e18/src/ggml.c#L5280-L5287

So even if it returned `GGML_OP_VIEW` it would be immediately overwritten.

The `ggml_view_Nd` ops already return `GGML_OP_VIEW`:

https://github.com/ggerganov/ggml/blob/69bf842b39e6a53218eda35852a91d395d357e18/src/ggml.c#L6335-L6354

But in any case, I don't see a problem to change `ggml_view_tensor` to return `GGML_OP_VIEW`. It just won't make a difference in the current usages

ggerganov:
> Anyway, all of these issues could be solved by adding a view_src to ggml_tensor

Let's do it.

> Either way, I think we need to do this or remove all the manual _inplace operations.

Initially, I thought that the `_inplace` ops would be used only internally in `ggml.c` and should not be exposed through the API, since at the very least, when the graph uses one of them, it breaks the backward pass. At some point, I decided to drop this restriction and probably this wasn't a good choice.

The `_inplace` API allows for some extra optimizations for forward-only use-cases which are currently 99% of `ggml` applications, so in this regard it kind of makes sense. But even so, if we extend `ggml_tensor` with `view_src`, we should still try to eventually deprecate the `_inplace` ops to simplify the API.

PR #514, opened 2023-09-08
METAL in CI

dc-dc-dc:
leveraging macos-13 runners you can get access to the Metal framework


ggerganov (approved):
Hm, this is really cool - we should add this to https://github.com/ggerganov/llama.cpp

invisiblepancake (commented):
Dep for METAL in CI language

PR #528, opened 2023-09-16
Fix batch for ggml_conv_2d

skirodev:
This PR is to fix batch inference for ggml_conv_2d.

Here are the results of benchmark testing using [clip.cpp](https://github.com/monatis/clip.cpp) and the ggml_laion_clip-vit-b-32-laion2b-s34b-b79k f16 model on the Imagenet1k dataset.

Original code:
`| total                | 0.0805 | 0.1185 |`

Fixed code:
`| total                | 0.3137 | 0.4582 |`

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -13607,20 +13607,22 @@ static void ggml_compute_forward_conv_2d_f16_f32(
         {
             ggml_fp16_t * const wdata = (ggml_fp16_t *) params->wdata + 0;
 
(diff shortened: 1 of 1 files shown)


monatis (approved):

monatis:
Interesting. Batched Conv2D was implemented by me just for clip.cpp. It was tested with CIFAR10 dataset, and results were reasonable. Maybe we had a regression later on.

ggerganov (approved):

PR #547, opened 2023-10-03
ggml backends interface v1

slaren:
Initial version of the common backends interface.

Supports CPU and CUDA with full offloading, but not partial offloading or fallback to the CPU backend for unimplemented ops.

Modifies the gpt-2 example to use the backends interface. If built with CUDA support and executed with `-ngl 1`, the CUDA backend is used.


ggerganov:
> many of the changes in this PR are due to a sync with llama.cpp

I will do a sync of `ggml` with `llama.cpp` later tonight so we can focus on the relevant changes

ggerganov:
@slaren Let's rebase on `master` so we can see the relevant diff

src/ggml-cuda.h
```diff
@@ -41,6 +43,10 @@ GGML_API bool   ggml_cuda_compute_forward(struct ggml_compute_params * params, s
 GGML_API int    ggml_cuda_get_device_count(void);
 GGML_API void   ggml_cuda_get_device_description(int device, char * description, size_t description_size);
 
+// backend API
+GGML_API ggml_backend_t ggml_backend_cuda_init(void); // TODO: take a list of devices to use
```
iboB, line 47: Hey, it will be nice if besides devices, one could provide stream handles and cublas handles here. This will allow plugging ggml in existing code. 

I currently have hacked `ggml_init_cublas` as it is in `master` to allow exactly this. Thus I created a pytorch plugin which uses the cuda instance from pytorch and now I directly assign pytorch tensor pointers to ggml tensors.

I was planning to make my hack more presentable, but this PR is basically the way to go. 

The same goes for other backends, btw. Introducing ggml to existing codebases would be easiest from the inside out as a plugin: first doing some leaves of the computation, then more and more towards the root.

iboB, line 47: Hey, it will be nice if besides devices, one could provide stream handles and cublas handles here. This will allow plugging ggml in existing code.

I currently have hacked ggml_init_cublas as it is in master to allow exactly this. Thus I created a pytorch plugin which uses the cuda instance from pytorch and now I directly assign pytorch tensor pointers to ggml tensors.

I was planning to make my hack more presentable, but this PR is basically the way to go.

iboB (changes requested):
Doing something similar as the comment in `ggml-cuda.h` for all other backends, should be the goal IMO. 

Introducing ggml to existing codebases would be easiest from the inside out as a plugin: first doing some leaves of the computation, then more and more towards the root.

src/ggml-cuda.h
slaren: In the future, support for the old API will be removed, and then all the globals will be moved into the backend context. At that point we could consider how to add this functionality, but I think it is too early for that.

iboB: What's wrong with using the globals? 

If one wants to add a plugin, it is very unlikely for them to need multiple backend instances. And even if they do, this can be marked as "not supported yet".

If one wants to use the old behavior, they won't provide handles and it will be just the same.

No existing functionality is going to break if this feature is added with the PR. On the contrary: plugins like mine will be possible with vanilla ggml

slaren: There are a few problems with this:
- The backends interface is still not finished, many things are going to change in the next weeks
- I don't really like the idea exposing internals of the CUDA backend, especially at a point where things are going to change significantly
- This looks like an extremely niche application, so I am not convinced that it needs to be merged. The patch can be maintained in the relevant repository.
- I do not understand what are your requirements, so I cannot really do this myself in the first place

ggerganov:
Need to log off earlier today - will review tomorrow.
Is this branch in a state where I can attempt to add a Metal backend?

slaren:
The branch should be stable, I think this is already good enough for a v1.

It should be easier to implement than the last time. There is a pointer to the buffer in the tensors, and that should help with backends that cannot abstract a buffer+offset into a pointer, like Metal. But there are very likely going to be some rough edges. So I would say to not do it yet, unless you are willing to look into the internals and change anything that may be required.

src/ggml-cuda.h
iboB: I wouldn't say that it's a niche application. Consider an existing inference codebase which doesn't use ggml. Be it pytorch-based, pure cuda, or tensorflow. If one wants to migrate it to ggml for some reason, what can they do? 

They can completely rewrite the inference in ggml which is quite error prone. Or they can implement parts of it in ggml. Leaves of the computation graph and leave the rest as is. Gradually moving to entirely ggml-based inference while having a working version at each step.

This is exactly my use case. I'm moving a pytorch-based application to ggml.

I think this can be quite beneficial for the library and people who're looking into evaluate it, without having to go all the way.

examples/gpt-2/main.cpp
```diff
@@ -299,14 +330,25 @@ bool gpt2_model_load(const std::string & fname, gpt2_model & model, gpt_vocab &
         const size_t memory_size = ggml_nbytes(model.memory_k) + ggml_nbytes(model.memory_v);
 
         printf("%s: memory size = %8.2f MB, n_mem = %d\n", __func__, memory_size/1024.0/1024.0, n_mem);
+
+        // allocate buffer and tensors
+        model.buffer_kv = ggml_backend_alloc_buffer(model.backend, memory_size + 256);
+        ggml_allocr * alloc = ggml_allocr_new_from_buffer(model.buffer_kv);
+        ggml_allocr_alloc(alloc, model.memory_k);
+        ggml_allocr_alloc(alloc, model.memory_v);
+        ggml_allocr_free(alloc);
```
ggerganov, line 339: What is the purpose of this `ggml_allocr` here?

slaren: To allocate the tensors in the backend buffer. It is overkill, and we probably should add a simpler allocator that cannot free tensors, but since this is only done during initialization it doesn't seem like a big deal to use the big allocator.

examples/gpt-2/main.cpp
```diff
@@ -360,11 +403,19 @@ bool gpt2_model_load(const std::string & fname, gpt2_model & model, gpt_vocab &
                 return false;
             }
 
-            fin.read(reinterpret_cast<char *>(tensor->data), ggml_nbytes(tensor));
+            // read into a temporary buffer first, then copy to the tensor
+            // TODO: read directly into the tensor if the backend is CPU
```
YavorGIvanov, line 407: Isn't this a simple if ? :)  And is there a simple way to check whether a backend is cpu or gpu ones it is created

slaren: The problem is that there isn't an easy way to tell if  backend is the CPU backend. The only way to do this at the moment is by checking the name of the backend (as is done in `gpt2_eval` to set the number of threads), but I don't really like that. I am thinking about adding a backend id enum, but that doesn't seem great either because it will need to be modified every time a new backend is added.

examples/gpt-2/main.cpp
```diff
@@ -703,11 +760,10 @@ bool gpt2_eval(
     ggml_allocr_alloc_graph(allocr, gf);
 
     // run the computation
-    struct ggml_cplan plan = ggml_graph_plan(gf, n_threads);
-    static std::vector<uint8_t> work_buffer;
-    work_buffer.resize(plan.work_size);
-    plan.work_data = work_buffer.data();
-    ggml_graph_compute(gf, &plan);
+    if (strcmp(ggml_backend_name(model.backend), "CPU") == 0) {
```
YavorGIvanov, line 763: Ah guess this is a way for figuring out whether a backed is CPU (my prev comment). Can we add a helper function returning the type of backend ?

ggerganov, line 763: minor : this should be more strongly-typed, such as `ggml_backend_is_cpu()`

ggerganov (approved):

examples/gpt-2/main.cpp
slaren: Yeah I think that's better than adding an enum. If you need to do a backend-specific operation, then using a backend-specific function to check if it is the correct backend is fine.

examples/gpt-2/main.cpp
```diff
@@ -334,13 +334,21 @@ bool gpt2_model_load(const std::string & fname, gpt2_model & model, gpt_vocab &
...
-        // allocate buffer and tensors
+        // create a backend buffer (can be in host or device memory)
         model.buffer_kv = ggml_backend_alloc_buffer(model.backend, memory_size + 256);
 
-        ggml_allocr * alloc = ggml_allocr_new_from_buffer(model.buffer_kv);
-        ggml_allocr_alloc(alloc, model.memory_k);
-        ggml_allocr_alloc(alloc, model.memory_v);
-        ggml_allocr_free(alloc);
+        // allocate the tensors into the backend buffer
+        // TODO: better API for this
```
slaren, line 341: I am not sure what a better API would look like, unless we integrate the backends interface into ggml contexts. We could use a more lightweight allocator, but the API is going to be mostly the same.

examples/gpt-2/main.cpp
```diff
@@ -770,25 +842,27 @@ int main(int argc, char ** argv) {
...
+        size_t align = ggml_backend_get_alignment(model.backend);
+        allocr = ggml_allocr_new_measure(align);
 
         // create the worst case graph for memory usage estimation
         int n_tokens = std::min(model.hparams.n_ctx, params.n_batch);
         int n_past = model.hparams.n_ctx - n_tokens;
         struct ggml_cgraph * gf = gpt2_graph(model, allocr, n_past, std::vector<gpt_vocab::id>(n_tokens, 0));
 
         // compute the required memory
-        size_t mem_size = ggml_allocr_alloc_graph(allocr, gf) + GGML_MEM_ALIGN;
```
YavorGIvanov, line 786: I tested doing this CPU-only for sam.cpp and I had to keep this addition of GGML_MEM_ALIGN (I replaced it with "align" from above, but whatever). Otherwise I got an assert later here:
```
            fprintf(stderr, "%s: not enough space in the buffer (needed %zu, largest block available %zu)\n",
                    __func__, size, max_avail);
            GGML_ASSERT(!"not enough space in the buffer");
```
The difference in sizes was either 64 or 48 bytes and the alignment is 64 bytes, so I decided that I need this addition here in order for sam.cpp to work. Initially I thought you remove it here, because you handle it internally in the ggml now, but I had the described above problem.

YavorGIvanov: Here is my experiment: https://github.com/YavorGIvanov/sam.cpp/pull/22

slaren: Should be fixed now, I forgot to add the alignment to the size. The backends should always return aligned buffers, or with enough space to align them, so this shouldn't be necessary anymore.

src/ggml-cuda.h
slaren: Would a function to change the main stream work? That would allow you to queue the ggml-cuda computations on the pytorch stream, which I may improve performance due to reduced synchronizaiton. Is there any reason to re-use the pytorch cuBLAS handle? Does each cuBLAS handle use a large amount of GPU resources?

examples/gpt-2/main.cpp
YavorGIvanov: Great. I will remove and verify it tomorrow, but from your change it seems that indeed it should be resolved :)

examples/gpt-2/main.cpp
```diff
@@ -703,11 +796,16 @@ bool gpt2_eval(
...
-    ggml_graph_compute(gf, &plan);
+    if (ggml_backend_is_cpu(model.backend)) {
+        ggml_backend_cpu_set_n_threads(model.backend, n_threads);
+    }
+#ifdef GGML_USE_METAL
+    // TODO: not great - what should we do?
+    if (ggml_backend_is_metal(model.backend)) {
+        ggml_backend_metal_set_n_threads(model.backend, n_threads);
+    }
+#endif
```
ggerganov, line 807: @slaren Any ideas how to avoid this `ifdef`.  It's not terrible, but maybe try to minimize such branching from the start

ggerganov (approved):
Let's merge this - I think it is a good addition to the framework

examples/gpt-2/main.cpp
slaren, line 807: What is the issue here? There are always going to be some backend-specific functions, I think this is unavoidable.

slaren: In the long term, I would like to remove all `#ifdefs` entirely and just make the backend init functions return `NULL` if the backend is not available. That should allow using a single build for all backends. But the checks before calling a backend-specific function will always be required.

ggerganov: We could extend the backend API with `set_n_threads()` and backends like CUDA would be just noops.
This specific code would become a one line without `#ifdef`

```c++
ggml_backend_set_n_threads(model.backend, n_threads);
```

Just an idea - we can just leave it as it is

slaren: We can't do this for every backend-specific function. I am not sure that calling this function `ggml_backend_metal_set_n_threads` is very honest here. If it was renamed to `ggml_backend_metal_set_n_cb`, the problem would become more obvious.

src/ggml-cuda.cu
```diff
@@ -6585,7 +6732,8 @@ static void ggml_cuda_op_mul_mat(
         if (convert_src1_to_q8_1) {
             src1_ddq[id] = (char *) ggml_cuda_pool_malloc(nrows1*src1_padded_col_size*q8_1_ts/q8_1_bs, &src1_asq[id]);
 
-            if (split && src1_on_device && src1_is_contiguous) {
+            // FIXME: why split only? src1 never gets quantized, breaks ggml-backend/GPT-2
+            if (/*split &&*/ src1_on_device && src1_is_contiguous) {
```
slaren, line 6736: @JohannesGaessler does this fix look ok? Without it, when src0 is `GGML_BACKEND_GPU` rather than split, src1 is never quantized.

JohannesGaessler: It should be fine. This is not a case that I previously considered for the implementation (because for LLaMA the backend for quantized matrices is always `GGML_BACKEND_GPU_SPLIT`) and I don't remember this being a necessity.

examples/gpt-2/main.cpp
ggerganov: Yes, it's better to have the correct name

PR #550, opened 2023-10-04
Added CMake export generation and install.

alud0s3:
Enhanced install command usage via GNUInstallDirs.

Added helper module GGMLExtraDependencies.cmake that:
- Automatically adds third party libraries (e.g. CUDA) into generated exports in addition to including and linking to them during the build process.
- Wraps libraries that can not be found via find_package nicely into CMake targets so that paths etc. can be resolved correctly inside the exported package config file.

Implemented CMake package export configuration generation and installation using GGMLExtraDependencies.cmake that allows using GGML in other projects via find_package(). It is implemented in a way that also allows using the GGML library directly from its build directory without the need to install it first (allows faster development of projects that are strongly coupled with the development of the GGML library).

Please note that I only checked building with OpenCL/CLBlast since I only have an older AMD GPU available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     set(GGML_STANDALONE ON)
     include(cmake/GitVars.cmake)
     include(cmake/BuildTypes.cmake)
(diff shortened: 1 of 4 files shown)


dg0yt:
Is this solved?

ggerganov:
It needs to be updated to latest version. An likely simplified into separate parts - it's hard to review such large change.

PR #552, opened 2023-10-05
ggml-backend : metal

ggerganov:
This is probably not the best way to implement this, but at least it works

### DIFF
--- a/src/ggml-metal.m
+++ b/src/ggml-metal.m
@@ -151,8 +151,6 @@ static void ggml_metal_log(enum ggml_log_level level, const char* format, ...){
     }
 }
 
-
-
 struct ggml_metal_context * ggml_metal_init(int n_cb) {
     GGML_METAL_LOG_INFO("%s: allocating\n", __func__);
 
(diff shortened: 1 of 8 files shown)


slaren:
Other than some leftovers from the CPU backend, like the work buffers, plans and `n_threads`, looks good.

src/ggml-metal.m
```diff
@@ -1371,3 +1369,189 @@ void ggml_metal_graph_compute(
...
+    void * data = ggml_metal_host_malloc(size);
+
+    ggml_metal_add_buffer(metal_ctx->ctx, "backend", data, size, 0);
+
+    return ggml_backend_buffer_init(backend, metal_backend_buffer_i, data, size);
+}
+
+static size_t ggml_backend_metal_get_alignment(ggml_backend_t backend) {
+    // TODO: get page size ??
+    //return sysconf(_SC_PAGESIZE);
```
slaren, line 1431: This is the tensor alignment only. I don't think that tensors need to be aligned to pages.

ggerganov: Ok, I'll clean-up this tomorrow and merge and then we can merge the main PR

PR #557, opened 2023-10-07
fix MSVC build issues

slaren:
Fixes:
- Remove `struct` literal in C++ code in `ggml-cuda`
- `interface` in `ggml-backend` is a reserved word in MSVC, renamed to `iface`
- Some `ggml.h` functions were missing `GGML_API`, and were not being exported in dynamic library builds
- `whisper-cpp` link failed due to being built as a dynamic library without exported symbols, changed to static library
- Issues with multiple GPU in the `ggml-backend` implementation of `ggml-cuda`

### DIFF
--- a/examples/whisper/CMakeLists.txt
+++ b/examples/whisper/CMakeLists.txt
@@ -1,7 +1,7 @@
 #
 # whisper
 
-add_library(whisper-cpp
+add_library(whisper-cpp STATIC
     whisper.cpp
     )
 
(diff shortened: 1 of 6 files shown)


slaren:
@ggerganov does the change to the whisper.cpp `CMakeLists.txt` look good? Maybe I should define `WHISPER_SHARED` instead?

ggerganov (approved):

ggerganov:
> Maybe I should define WHISPER_SHARED instead?

No need I think - this should be fine

PR #558, opened 2023-10-08
sync : llama.cpp (Metal + OpenCL + minor alibi)

ggerganov:
Bringing Metal stuff back to ggml, so I can move it to `whisper.cpp`


slaren (approved):
Looks good, but I can't really comment on the OpenCL or Metal changes.

Issue #559, opened 2023-10-08
ggml : replace conv stage_0 and stage_1 with im2col and mul_mat


ggerganov:
Currently, we implement `ggml_conv_1d` and `ggml_conv_2d` as a sequence of 2 internal ops: `stage_0` and `stage_1`

For more context see the discussion: https://github.com/ggerganov/ggml/pull/483

We should instead introduce `ggml_im2col` and reuse the `ggml_mul_mat` implementation both on the CPU and the GPU.

FSSRepo:
I will do it. @ggerganov

ggerganov:
We should probably start with some unit tests of the current implementation so that we can easily prove the new version is correct

FSSRepo:
> We should probably start with some unit tests of the current implementation so that we can easily prove the new version is correct

That would be great to speed up my work. Right now, I'm making a lot of changes blindly without testing since I want to finish quickly.

audiovention:
I see that this issue is almost closed with the proposed changes almost ready in a pull request. 
I had a different idea on how to restructure the two stages and wanted to share. 
IMHO the im2col step should write in the scratch memory. As I'm currently working on a project that involves quite a lot of conv1d steps, and having each one create an internal tensor for the im2col output completely blows up my memory usage, while there's no good reason for this memory to all exist at the same time.

slaren:
> having each one create an internal tensor for the im2col output completely blows up my memory usage

Does this happen while using `ggml-alloc`?

audiovention:
No ggml-alloc

slaren:
With `ggml-alloc`, the memory of the temporary tensors should be automatically reused. The memory usage should be very similar compared to using the work buffer. If it doesn't, that most likely would be a bug. You can find an example of how to use it in the gpt-2 example, or [here](https://github.com/ggerganov/ggml/blob/ggml-backend-v2/examples/gpt-2/main-alloc.cpp) without all the `ggml-backend` stuff.

audiovention:
oh ok, still very new to ggml, figuring out things
does this apply to back-propagation as well?

slaren:
I think that using `ggml-alloc` for training is possible, but currently requires some hacks to avoid freeing/overwriting some tensors. You could look into the `finetune` example in llama.cpp to see how it handles this.

ggerganov:
Completed via #564

PR #564, opened 2023-10-09
ggml: replace conv 1D - 2D stage_0 and stage_1 with im2col and mul_mat

FSSRepo:
Fix this #559

This pull request aims to add `ggml_im2col` operator, which I have no idea what it is (I'm serious). I only know that it converts images into columns, eliminating the stage 0 of conv1d and conv2d, and refactors `ggml_mul_mat` to support float16 source data types, presumably to reduce memory usage, and remove the stage 1 of conv1d and conv2d. Additionally, I want to implement CUDA kernels for these operations.

Conv1D = im2col (1D) + Matrix Multiplication
```c++
        struct ggml_context * ctx,
        struct ggml_tensor  * a,
        struct ggml_tensor  * b,
        int                   s0,
        int                   p0,
        int                   d0) {
    struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, 0, p0, 0, d0, 0, false);
    result = ggml_mul_mat(ctx, a, result);
    return result;
}
```

Conv2D = im2col (2D) + Matrix Multiplication
```c++
struct ggml_tensor * ggml_conv_2d(
    struct ggml_context * ctx,
    struct ggml_tensor  * a,
    struct ggml_tensor  * b,
    int                  s0,
    int                  s1,
    int                  p0,
    int                  p1,
    int                  d0,
    int                  d1) {

    struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, s1, p0, p1, d0, d1, true); // [N, OH, OW, IC * KH * KW]
    result = ggml_mul_mat(ctx, a, result);

    return result;

}
```

Where:
```c++
GGML_API struct ggml_tensor * ggml_im2col(
            struct ggml_context * ctx,
            struct ggml_tensor  * a,
            struct ggml_tensor  * b,
            int                  s0,
            int                  s1,
            int                  p0,
            int                  p1,
            int                  d0,
            int                  d1,
            bool             is_2D);
```

Here's the plan to finish this PR. I'll be working on this as soon as possible since I won't have more time due to my studies:

- [x] remove `ggml_conv_1d_stage_0` and `ggml_conv_1d_stage_1`.
- [x] remove `ggml_conv_2d_stage_0` and `ggml_conv_2d_stage_1`.
- [x] add `ggml_im2col` for 1D and 2D, I am going to investigate the difference in order to refactor the code better and avoid duplication.
- [x] add `ggml_compute_forward_mul_mat_f16_f32` for support float 16 matrix multiplication, this reduces memory.
- [x] Research how stage 1 (General Matrix Multiply) of Conv1D differs from Conv2D, to see if it involves the same matrix multiplication or not.
- [x] Add the CUDA kernels, I already have them done, I just want to finish the previous ones.
- [x] Perform final tests on the conv1d and conv2d functions and im2col, optimize results.
- [x] reuse `ggml_mul_mat`, and remove `gemm_fp16_out_fp32`. I'm testing the differences.
- [x] Refactor the code (remove unused comments and the GGML_OP_CONV_1D and GGML_OP_CONV_2D remainder implementations that will no longer be used)."
```c++
        GGML_OP_ALIBI,
        GGML_OP_CLAMP,
        GGML_OP_CONV_1D, // this
        GGML_OP_CONV_TRANSPOSE_1D,
        GGML_OP_CONV_2D, // this
        GGML_OP_IM2COL,
```

Suggestions and feedback are welcome.


FSSRepo:
Added some 'unit' tests:

master:
```bash
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
main: compute buffer size: 0.00 MB
ggml_conv2d (64): PASS
```

PR:
```bash
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
GGML_ASSERT: D:\proyectos\ggml\src\ggml.c:6425: ggml_can_mul_mat(a, b)
```

😂

FSSRepo:
Fixed:

```bash
D:\proyectos\ggml\build\bin\Release>test-conv1d.exe
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
main: compute buffer size: 0.00 MB
ggml_conv1d (15): PASS
D:\proyectos\ggml\build\bin\Release>test-conv2d.exe
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
main: compute buffer size: 0.00 MB
ggml_conv2d (64): PASS
```

FSSRepo:
Final test with new data and results:

image

src/ggml.c
```diff
@@ -6431,7 +6422,11 @@ struct ggml_tensor * ggml_mul_mat(
         struct ggml_context * ctx,
         struct ggml_tensor  * a,
         struct ggml_tensor  * b) {
-    GGML_ASSERT(ggml_can_mul_mat(a, b));
+    // hack to admit GEMM custom operator
+    bool mult_mat_conv1d = (a->ne[0] * a->ne[1]) == b->ne[0];
+    bool mult_mat_conv2d = (a->ne[0] * a->ne[1] * a->ne[2]) == b->ne[0];
```
ggerganov, line 6427: You should be able to avoid this hack by using `ggml_reshape` before calling `ggml_mul_mat` in `ggml_conv_xd`

src/ggml.c
```diff
@@ -11801,7 +11696,9 @@ static bool ggml_compute_forward_mul_mat_use_blas(
 }
 #endif
 
-static void ggml_compute_forward_mul_mat(
+// legacy multiplication matrix just float 32 data type
+
+static void ggml_compute_forward_mul_mat_f32_f32(
```
ggerganov, line 11701: I don't get why this was changed?
This function does all types of matrix multiplications (F32, F16, Q) - it's not just F32 x F32

src/ggml.c
```diff
@@ -12012,6 +11909,116 @@ static void ggml_compute_forward_mul_mat(
...
 }
 
+// GEMM
+// TODO: compare gemm op with the current implementation of mul_mat
+
+static void ggml_compute_forward_mul_mat_f16_f32(
+        const struct ggml_compute_params * params,
+        const struct ggml_tensor * src0,
+        const struct ggml_tensor * src1,
+              struct ggml_tensor * dst) {
```
ggerganov, line 11919: This seems unnecessary - try to remove this and just reuse the original function

src/ggml.c
```diff
@@ -14549,62 +14329,6 @@ static void ggml_compute_forward_conv_2d_stage_0_f32(
...
-    for (int i = 0; i < N; i++) {
-        ggml_fp16_t * A = (ggml_fp16_t *)src0->data; // [m, k]
-        ggml_fp16_t * B = (ggml_fp16_t *)src1->data + i * m * k; // [n, k]
-        float * C = (float *)dst->data + i * m * n; // [m, n]
-
-        gemm_f16_out_f32(m, n, k, A, B, C, ith, nth);
-    }
-}
-
 static void ggml_compute_forward_conv_2d_f16_f32(
```
ggerganov, line 14332: This should be removed, correct?

ggerganov (commented):
Overall, I think this is great - I'm surprised you pulled it so quickly!

However, there are some things in `ggml.c` that should be done in a simpler way. See my. comments.

src/ggml.c
FSSRepo: The current conv1d and conv2d need fp16 x fp16 = fp32 for reduce memory usage and precision.

FSSRepo:
@ggerganov @slaren I need some feedback about ggml_mul_mat

In the following mult of matrices with the shape (3,2) x (2, 3):

$$
\begin{bmatrix}
2 & 8 \\
5 & 1 \\
4 & 2 \\
\end{bmatrix}
\times
\begin{bmatrix}
10 & 9 & 5 \\
5 & 9 & 4 \\
\end{bmatrix}
\=
\begin{bmatrix}
60 & 90 & 42 \\
55 & 54 & 29 \\
50 & 54 & 28 \\
\end{bmatrix}
$$

To resolve it with `gemm_f16_out_f32` I need to transpose B (2, 3) -> (3, 2) but `ggml_mul_mat` fail

```bash
D:\proyectos\cpp-projects\ggml-test\build\Release>mult-mat-test.exe
Transposed B:
[10.0 5.0
9.0 9.0
5.0 4.0]
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 1084 bytes
Performing ggml_gemm test:
60.0 90.0 42.0
55.0 54.0 29.0
50.0 54.0 28.0
gemm_mult: PASSED
main: compute buffer size: 0.0625 KB

Performing ggml_mul_mat test:
105.0 48.0 -123612456.0
78.0 37.0 -54790864.0
-1873442825329845586624512.0 -0.0 0.0
ggml_mul_mat (6): FAILED
```

Am I making a mistake? look the test `test-mul-mat` that I added.

ggerganov:
In `ggml`, `ne[0]` is always the number of columns in the tensor.
Also, `ggml_mul_mat` returns transposed result

FSSRepo:
> In `ggml`, `ne[0]` is always the number of columns in the tensor. Also, `ggml_mul_mat` returns transposed result

FSSRepo:
### Conv2d GEMM:

Matrix A: [3, 3, 10, 10]
im2col: [90, 8, 6, 1]

m: 10, n: 48, k: 90

Where:
Matrix A (m, k) = [10, 90] --> (10, 10 * 3 * 3)
Matrix B (im2col) No Transposed (n, k) = [48, 90] --> (1 * 6 * 8, 10 * 3 * 3)
Result Con2D (m, n) = [10, 48] --> (1 * 10, 6 * 8)

### ggml_mul_mat

**NEED**: reshape `ne` updating the byte offsets `nb`.

Remember (columns, rows) in `ggml_mul_mat`
matrix A = [90, 10]
Matrix B = [90, 48] (no require transpose)
Result Transposed = [48, 10]

Result Conv2D = Tranpose([48, 10]) = [10, 48]

FSSRepo:
Now conv2d is this to reuse `ggml_mul_mat`.

```c++
struct ggml_tensor* im2col_res = ggml_im2col(ctx0, model.a, model.b, s0, s1, p0, p1, d0, d1, true);
    struct ggml_tensor* conv2d_res = ggml_cont(ctx0, ggml_transpose(ctx0,  ggml_mul_mat(ctx0, ggml_cont(ctx0, ggml_reshape_2d(ctx0, model.a, (model.a->ne[0] * model.a->ne[1] * model.a->ne[2]),  model.a->ne[3])),ggml_cont(ctx0, ggml_reshape_2d(ctx0, im2col_res, im2col_res->ne[0],  (im2col_res->ne[3] * im2col_res->ne[2] * im2col_res->ne[1]))))));
```

FSSRepo:
Error:

CPU Backend:
```bash
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
main: compute buffer size: 0.02 MB
GGML_ASSERT: D:\proyectos\ggml\src\ggml.c:11730: nb10 == sizeof(float)
```

GPU Backend:

```bash
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
load_model: using CUDA backend
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3050 Laptop GPU, compute capability 8.6
main: compute buffer size: 0.02 MB
ggml_cuda_cpy: unsupported type combination (f16 to f16)
GGML_ASSERT: D:\proyectos\ggml\src\ggml-cuda.cu:7168: false
```

FSSRepo:
@slaren Does the CUDA implementation of ggml_mul_mat support data sources of type fp16?

I tried to fix this `GGML_ASSERT: D:\proyectos\ggml\src\ggml-cuda.cu:7168: false` implementing 'ggml_cpy_f16_f16_cuda', but now fails `cuBLAS error 7 at D:\proyectos\ggml\src\ggml-cuda.cu:6337: an unsupported value or parameter was passed to the function`.

```c++
cublasGemmEx(g_cublas_handles[id], CUBLAS_OP_T, CUBLAS_OP_N,
                    row_diff, src1_ncols, ne10,
                    &alpha_f16, src0_ptr, CUDA_R_16F, ne00,
                                src1_ptr, CUDA_R_16F, ne10,
                    &beta_f16,   dst_f16, CUDA_R_16F, ldc,
                    CUBLAS_COMPUTE_16F,
                    CUBLAS_GEMM_DEFAULT_TENSOR_OP));
```

slaren:
It should work, but maybe there is a bug with fp16 `src1`. Uncommenting this line may show more details about the error:
https://github.com/ggerganov/ggml/blob/f04b02bc9f57c80bf93af0dbe947f314b9a65fc5/src/ggml-cuda.cu#L5681

FSSRepo:
@slaren 

```bash
I! cuBLAS (v12.0) function cublasStatus_t __cdecl cublasSetStream_v2(struct cublasContext *,struct CUstream_st *) called:
i!  handle: type=cublasHandle_t; val=POINTER (IN HEX:0x0000015CD6DC1730)
i!  streamId: type=SOME TYPE; val=POINTER (IN HEX:0x0000015CBBBD36B0)
i! Time: 2023-10-14T19:38:20 elapsed from start 0.016667 minutes or 1.000000 seconds
i!Process=12976; Thread=18488; GPU=0; Handle=POINTER (IN HEX:0x0000015CD6DC1730); StreamId=POINTER (IN HEX:0x0000000000000000) (defaultStream); MathMode=CUBLAS_TF32_TENSOR_OP_MATH
i! COMPILED WITH: Microsoft Visual Studio / 192628806.0
I! cuBLAS (v12.0) function cublasStatus_t __cdecl cublasGemmEx(struct cublasContext *,cublasOperation_t,cublasOperation_t,int,int,int,const void *,const void *,enum cudaDataType_t,int,const void *,enum cudaDataType_t,int,const void *,void *,enum cudaDataType_t,int,cublasComputeType_t,cublasGemmAlgo_t) called:
i!  handle: type=cublasHandle_t; val=POINTER (IN HEX:0x0000015CD6DC1730)
i!  transa: type=cublasOperation_t; val=CUBLAS_OP_T(1)
i!  transb: type=cublasOperation_t; val=CUBLAS_OP_N(0)
i!  m: type=int; val=10
i!  n: type=int; val=8
i!  k: type=int; val=30
i!  alpha: type=void; val=POINTER (IN HEX:0x0000000F12CFEE98)
i!  A: type=void; val=POINTER (IN HEX:0x000000071A820E00)
i!  Atype: type=cudaDataType_t; val=CUDA_R_16F(2)
i!  lda: type=int; val=30
i!  B: type=void; val=NULL_PTR(POINTER (IN HEX:0x0000000000000000))
i!  Btype: type=cudaDataType_t; val=CUDA_R_16F(2)
i!  ldb: type=int; val=30
i!  beta: type=void; val=POINTER (IN HEX:0x0000000F12CFEF28)
i!  C: type=void; val=POINTER (IN HEX:0x000000071A821600)
i!  Ctype: type=cudaDataType_t; val=CUDA_R_16F(2)
i!  ldc: type=int; val=10
i!  computeType: type=cublasComputeType_t; val=CUBLAS_COMPUTE_16F(64)
i!  algo: type=SOME TYPE; val=CUBLAS_GEMM_DEFAULT_TENSOR_OP(99)
i! Time: 2023-10-14T19:38:20 elapsed from start 0.016667 minutes or 1.000000 seconds
i!Process=12976; Thread=18488; GPU=0; Handle=POINTER (IN HEX:0x0000015CD6DC1730); StreamId=POINTER (IN HEX:0x0000015CBBBD36B0); MathMode=CUBLAS_TF32_TENSOR_OP_MATH
i! COMPILED WITH: Microsoft Visual Studio / 192628806.0

cuBLAS error 7 at D:\proyectos\ggml\src\ggml-cuda.cu:6337: an unsupported value or parameter was passed to the function
current device: 0
```

slaren:
Try this change:
```diff
diff --git a/ggml-cuda.cu b/ggml-cuda.cu
index 654d363..96e63e0 100644
--- a/ggml-cuda.cu
+++ b/ggml-cuda.cu
@@ -6295,7 +6295,7 @@ inline void ggml_cuda_op_mul_mat_cublas(
             src1_as_f16 = (half *) ggml_cuda_pool_malloc(ne * sizeof(half), &src1_as);
             to_fp16_cuda(src1_ddf_i, src1_as_f16, ne, stream);
         }
-        const half * src1_ptr = src1->type == GGML_TYPE_F16 ? (const half *) src1_ddq_i : src1_as_f16;
+        const half * src1_ptr = src1->type == GGML_TYPE_F16 ? (const half *) src1_ddf_i : src1_as_f16;

         size_t dst_as = 0;
         half * dst_f16 = (half *) ggml_cuda_pool_malloc(row_diff*src1_ncols * sizeof(half), &dst_as);
```

FSSRepo:
> Try this change:

Works!!!!!

```bash
D:\proyectos\ggml\build\bin\Release>test-conv1d.exe
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
load_model: using CUDA backend
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3050 Laptop GPU, compute capability 8.6
main: compute buffer size: 0.00 MB

Performing test:
ggml_im2col (240): PASSED
ggml_conv1d (80): PASSED

D:\proyectos\ggml\build\bin\Release>test-conv2d.exe
load_model: ggml tensor size    = 320 bytes
load_model: backend buffer size = 0.00 MB
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce RTX 3050 Laptop GPU, compute capability 8.6
main: compute buffer size: 0.02 MB
Res [90, 8, 6, 1]

Performing test:
ggml_im2col (4320): PASSED
ggml_conv2d (480): PASSED

D:\proyectos\ggml\build\bin\Release>
```

FSSRepo:
@slaren ould you explain a bit what the `ggml_cont` function is used for and why it is sometimes necessary when performing matrix multiplications?

FSSRepo:
@ggerganov  I'm almost done with this pull request; I just need to remove the old functions that are no longer referenced for external use. I refer `ggml_compute_forward_conv_2d_f16_f32` and `ggml_compute_forward_conv_1d_f16_f32` due to the insertion of `ggml_im2col` and the reuse of `ggml_mul_mat`.

slaren:
`ggml_cont` is used for making a copy of a tensor that is contiguous in memory. Some operations such as `ggml_transpose` and views, only create a view of the original tensor with the strides adjusted, but as a result the elements of the tensor are no longer contiguous in memory. Many operations, including matrix multiplication, can only operate on tensors that are contiguous on memory, so in some cases after creating a view it is necessary to use `ggml_cont` to use these ops.

FSSRepo:
> `ggml_cont` is used for making a copy of a tensor that is contiguous in memory. Some operations such as `ggml_transpose` and views, only create a view of the original tensor with the strides adjusted, but as a result the elements of the tensor are no longer contiguous in memory. Many operations, including matrix multiplication, can only operate on tensors that are contiguous on memory, so in some cases after creating a view it is necessary to use `ggml_cont` to use these ops.

When I use ggml_reshape for adapt the tensor to ggml_mul_mat I should to use 'ggml_cont' or no?

slaren:
I think so, you should get an assert failure if `ggml_cont` is required and you don't use it.

src/ggml.c
```diff
@@ -7758,11 +7652,14 @@ struct ggml_tensor * ggml_conv_2d(
...
-    struct ggml_tensor * result = ggml_conv_2d_stage_0(ctx, a, b, s0, s1, p0, p1, d0, d1); // [N, OH, OW, IC * KH * KW]
-    result = ggml_conv_2d_stage_1(ctx, a, result);
-
+    struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, s1, p0, p1, d0, d1, true); // [N, OH, OW, IC * KH * KW]
+    result = ggml_reshape_4d(ctx,
+            ggml_cont(ctx, ggml_transpose(ctx,
+            ggml_mul_mat(ctx,
+            ggml_cont(ctx, ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1] * a->ne[2]),  a->ne[3])),
+            ggml_cont(ctx, ggml_reshape_2d(ctx, result, result->ne[0],  (result->ne[3] * result->ne[2] * result->ne[1])))))),
+            result->ne[1], result->ne[2], a->ne[3], result->ne[3]); // [N, OC, OH, OW] = [OC, IC * KH * KW] x [N*OH*OW, IC * KH * KW]
```
ggerganov, line 7661: Do we actually need all these `ggml_cont`?

If a tensor is already contiguous, after `ggml_reshape` you don't need to use `ggml_cont`.
This should eliminate the inner 2 `ggml_cont`

The outer one seems unnecessary - we just need to declare that the `ggml_conv` op returns a transposed result and it will be up to the user to transpose it if they need to.

FSSRepo:
@ggerganov This PR is ready to merge.

FSSRepo:
@PABannier what is `ggml_conv_transpose_2d`, why don't reuse `ggml_conv2d` or are totally different operations?

PABannier:
@FSSRepo Thanks for this great effort! Indeed, `ggml_conv_transpose_<1d|2d>` and `ggml_conv_<1d|2d>` are two different operations. Actually, one is reverting the other. A convolution reduces the spatial dimension and usually increases the channel dimension, while the transposed convolution increases the spatial dimension but reduces the channel dimension.

You can learn more about them here: [convolution](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html) and [transpose convolution](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose1d.html) .

src/ggml.c
```diff
@@ -7570,44 +7481,15 @@ GGML_API struct ggml_tensor * ggml_conv_1d(
...
         int                   p0,
         int                   d0) {
-    struct ggml_tensor * result = ggml_conv_1d_stage_0(ctx, a, b, s0, p0, d0);
-    result = ggml_conv_1d_stage_1(ctx, a, result);
+    struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, 0, p0, 0, d0, 0, false); // [N, OH, OW, IC * KH * KW]
+    result = ggml_reshape_3d(ctx, ggml_cont(ctx, ggml_transpose(ctx,
+            ggml_mul_mat(ctx,
+            ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1]),  a->ne[2]),
+            ggml_reshape_2d(ctx, result, result->ne[0],  (result->ne[2] * result->ne[1]))))),
+            result->ne[1],  a->ne[2], result->ne[2]);
```
ggerganov, line 7489:
```suggestion
    struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, 0, p0, 0, d0, 0, false); // [N, OH, OW, IC * KH * KW]
    result = ggml_reshape_3d(ctx,
            ggml_mul_mat(ctx,
            ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1]),  a->ne[2]),
            ggml_reshape_2d(ctx, result, result->ne[0],  (result->ne[2] * result->ne[1]))))),
            a->ne[2], result->ne[1], result->ne[2]);
```

I still think we don't need this `ggml_cont()` and `ggml_transpose()` here.
Have to take a deeper look to see if I'm missing something

FSSRepo:
As a final opinion on the topic, I would expect to see what people who truly need these features have to say about whether the current method of returning is suitable or the one you propose, as long as they don't have to use 'ggml_cont' and 'ggml_transpose.' If, in any way, they had to do it when operating with the regular version of 'ggml_conv,' then the ideal scenario would be to return the matrix without transposing it.

ggerganov:
The `whisper` example produces wrong results on this branch:

https://github.com/ggerganov/ggml/blob/97809b39bce234178cd143ec443dfee3b594b59a/ci/run.sh#L199-L219

It uses `ggml_conv_1d_ph()` so likely the problem is there:

https://github.com/ggerganov/ggml/blob/97809b39bce234178cd143ec443dfee3b594b59a/examples/whisper/whisper.cpp#L1542-L1561

In `ggml`, we have the following identity:

```c
z0 = ggml_mul_mat(ctx, x, y);
z1 = ggml_conv(ctx, ggml_transpose(ctx, ggml_mul_mat(ctx, y, x)));

z0 == z1
```

The `ggml_cont` operation is very expensive, so it definitely should not be implicitly applied.
The conv operators have to be implemented to output the correct results without `ggml_cont` and `ggml_transpose`.

FSSRepo:
> The `ggml_cont` operation is very expensive, so it definitely should not be implicitly applied.
> The conv operators have to be implemented to output the correct results without `ggml_cont` and `ggml_transpose`.
> 

I will see if `ggml_cont` can be omitted in both conv1d and conv2d, but ggml_transpose definitely needs to be included for it to produce the same results as the master branch implementation. I will also try to identify the issue with examples that require these operations. This is quite a complex task, and I'm not very familiar with how to implement the functionality of conv1d and conv2d in PyTorch.

ggerganov:
@FSSRepo I think you can leave the branch as it is and I will try to fix it

FSSRepo:
> @FSSRepo I think you can leave the branch as it is and I will try to fix it

Ok, i won't push commit more at this point

FSSRepo:
@ggerganov I don't get some error in the lastest commit.

ggerganov:
The computation was failing when using BLAS - on the CPU it supports only F32. I fixed that in 53f805e

I now applied the identity transformation that I told you about earlier:

```diff
diff --git a/src/ggml.c b/src/ggml.c
index 927f03a..b33a741 100644
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -7482,10 +7482,10 @@ GGML_API struct ggml_tensor * ggml_conv_1d(
         int                   p0,
         int                   d0) {
     struct ggml_tensor * result = ggml_im2col(ctx, a, b, s0, 0, p0, 0, d0, 0, false); // [N, OL, IC * K]
-    result = ggml_reshape_3d(ctx, ggml_cont(ctx, ggml_transpose(ctx,
+    result = ggml_reshape_3d(ctx,
             ggml_mul_mat(ctx,
-            ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1]),  a->ne[2]), // [OC，IC, K] => [OC, IC * K]
-            ggml_reshape_2d(ctx, result, result->ne[0],  (result->ne[2] * result->ne[1]))))), // [N, OL, IC * K] => [N*OL, IC * K]
+            ggml_reshape_2d(ctx, result, result->ne[0],  (result->ne[2] * result->ne[1])), // [N, OL, IC * K] => [N*OL, IC * K]
+            ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1]),  a->ne[2])), // [OC，IC, K] => [OC, IC * K]
             result->ne[1],  a->ne[2], result->ne[2]); // [N, OC, OL]
     return result;
 }
```

This again produces correct results for `test-conv-1d` and for `whisper`.
And we no longer need `ggml_cont` and `ggml_transpose`.

FSSRepo:
In Conv2D

you can replace `(a->ne[0] * a->ne[1] * a->ne[2])` -> `result->ne[0]`, I forgot that.

```c++
ggml_reshape_2d(ctx, a, result->ne[0],  a->ne[3])), // [OC，IC, KH, KW] => [OC, IC * KH * KW]
```

The same

```c++
ggml_reshape_2d(ctx, a, (a->ne[0] * a->ne[1] * a->ne[2]),  a->ne[3])), // [OC，IC, KH, KW] => [OC, IC * KH * KW]
```

FSSRepo:
The identity that you refer is just swap the matrix operands? 😂 Good that, I didn't get it.

FSSRepo:
> Will get back to this PR soon, sorry for delay

Don't worry. The next week I will add cuda kernels for `ggml_concat`, `ggml_upscale` and `ggml_group_norm`.

src/ggml-cuda.cu
```diff
@@ -4723,6 +4736,25 @@ static __global__ void clamp_f32(const float * x, float * dst, const float min,
     dst[i] = x[i] < min ? min : (x[i] > max ? max : x[i]);
 }
 
+static  __global__ void im2col_f32_f16(
+        const float * x, half * dst,
+        int ofs0, int ofs1, int IW, int IH, int CHW,
+        int s0, int s1, int p0, int p1, int d0, int d1) {
+    const int iiw = blockIdx.z * s0 + threadIdx.z * d0 - p0;
+	const int iih = blockIdx.y * s1 + threadIdx.y * d1 - p1;
```
slaren, line 4744: There is a tab here.
```suggestion
    const int iih = blockIdx.y * s1 + threadIdx.y * d1 - p1;
```

PR #580, opened 2023-10-13
Parallelize unary tanh on cpu, generalize ADD to allow more shapes

audiovention:
I'm working on a project which needed those operations. 
tanh was parallelized in the same manner as other unary ops 

ADD is generalized to allow for ggml_can_repeat constraint, instead of the ggml_can_repeat_rows
This was done adding two extra branches in the function, one of them is likely very slow and handles the most general case. The second is particularly optimized for my project's need (adding MxN and 1xP tensors) and uses ggml_vec_add1_f32.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -5558,8 +5558,8 @@ static struct ggml_tensor * ggml_add_impl(
         struct ggml_tensor * b,
         bool inplace) {
     // TODO: support less-strict constraint
-    //       GGML_ASSERT(ggml_can_repeat(b, a));
-    GGML_ASSERT(ggml_can_repeat_rows(b, a));
+    GGML_ASSERT(ggml_can_repeat(b, a));
+    // GGML_ASSERT(ggml_can_repeat_rows(b, a));
 
     bool is_node = false;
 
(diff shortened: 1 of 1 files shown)


src/ggml.c
```diff
@@ -9297,65 +9297,114 @@ static void ggml_compute_forward_add_f32(
...
+                const int64_t i12 = i02 % ne12;
+                const int64_t i11 = i01 % ne11;
+                const int64_t i10 = i00 % ne10;
+
+                float * dst_ptr  = (float *) ((char *) dst->data  + i03*nb3  + i02*nb2  + i01*nb1  + i00*nb0 );
+                float * src0_ptr = (float *) ((char *) src0->data + i03*nb03 + i02*nb02 + i01*nb01 + i00*nb00);
+                float * src1_ptr = (float *) ((char *) src1->data + i13*nb13 + i12*nb12 + i11*nb11 + i10*nb10);
+
+                *dst_ptr = *src0_ptr + *src1_ptr;
             }
```
ggerganov, line 9402: This branch will probably be very slow - most of the computation will probably go into computing the indices

ggerganov (approved):

PR #581, opened 2023-10-15
[experimental] ggml C++ bindings

slaren:
Adds an example with C++ bindings, and the gpt-2 example adapted to use these bindings.

At this point this mostly an experiment to see how ggml C++ bindings could look like. I am not sure if this is something that we may want at all.

The advantage over the C API would be automatic resource freeing, and maybe a nicer syntax in some cases.


Green-Sky (commented):
can we make c++ headers be `.hpp`

slaren:
I think using `.h` or `.hpp` is mostly a matter a matter of style. Generally I prefer to use `.h` for C++ too, but I don't really care either way.

ggerganov (commented):
The graph building code looks very clean.
The context stack should probably be `thread_local` so that we can build graphs in parallel threads.

I am also not sure if we need the C++ bindings. Some things indeed get much easier to implement in C++. Maybe if we limit the bindings scope to some very basic things like auto deallocation of resources and scoped-contexts it wouldn't be a problem to maintain them.

Issue #567, opened 2023-10-10
ggml : remove GGML_MAX_NODES limit


ggerganov:
We should be able to create graphs with large number of nodes. One application is support for LSTM networks: 

https://github.com/ggerganov/ggml/issues/467

The exact approach for implementing this is not currently clear, but some ideas and discussion is available in the linked issue above

datduonguva:
my approach for this would be to create the graph for just 1 cell. For each time step, will simply update the input and states, call the ggml_graph_compute_with_ctx and get the output. That way, the graph's size doesn't grow with the number of timestep.

Example:

https://github.com/datduonguva/ggml-experiments/blob/master/rnn_text_gen/rnn_text_generation.cpp

ggerganov:
> my approach for this would be to create the graph for just 1 cell.

cc @PABannier

PABannier:
Yes we discussed that in #467 ! But for large sequences, isn't the overhead of building the graph per cell slowing this down dramatically?

@datduonguva Have you tried by building a general graph by setting `GGML_MAX_NODES` to 100,000 for instance?

ggerganov:
@PABannier They build the graph only once and then for each iteration, they put different input in the input tensor and read the output from the output tensor and so on. I don't know if this is valid for all LSTM models, but this specific one has a fixed graph that does not change it's shape with time.

slaren:
> Deprecate `ggml_build_forward (struct ggml_tensor * tensor);`

Deprecate usually means that there would be a transition period where both APIs would work, but I am not sure that would be possible here. Do you mean actually mean that, or just removing `ggml_build_forward` entirely?

>I think leafs always have to be "viewed in full", i.e only the nodes/grads can be split?

For the purposes of `ggml-backend`, I don't care about leafs at all. I don't think that backends ever use them while computing a graph either. So I would be fine with just setting it to `NULL`. I don't know in what cases a list of leaf is required.

>The hash table can be reused by the views I think? 

The hashtable is only used to avoid revisiting nodes while expanding the graph, and views shouldn't be allowed to be expanded since they would just overwrite the parent's memory. I think it would be ok to just set the hash table to `NULL`.

>Probably OK to return graph views on the stack?

It's probably OK, but I doesn't really solve anything and it may reduce options in the future. Ie. we may need to change something that would require dynamic memory allocation even on views. So I think the safer route would be to also allocate views in a context.

[...]

PR #586, opened 2023-10-18
[wip] ggml-backend v2 : add ggml_backend_sched

### LINKED ISSUE #549: ggml : expose hash table API from ggml.c and reuse in ggml-alloc
#548 adds hash table implementation:

https://github.com/ggerganov/ggml/blob/a094d9c2db1d4cb64bbea3655032523dbc3c75c4/src/ggml.c#L16926-L16985

It should be exposed through the API and reused in `ggml-alloc`

slaren:
Adds support for computing graphs using multiple backends through `ggml_backend_sched`.

Currently, this allows partially offloading models to the GPU, and establishes the framework necessary for more advanced uses, such as:
- Automatic fallback to the CPU backend for ops unimplemented in the GPU backends
- Using any combination of backends, for example some layers could be offloaded with CUDA and others with OpenCL
- Using multiple CUDA or OpenCL devices by splitting the computation at a layer or tensor level

Basic functionality should be working now, but still needs some work, especially in the graph splitting logic.

Fixes #549 #567


slaren:
Currently, `ggml_backend_sched` allocates a graph for every backend split and copies the range of operations corresponding to the split from the original graph. This not only is not very efficient, but also wastes a large amount of memory due to the size of the `ggml_cgraph`. `ggml_backend_sched` objects are 160MB due to this.

I can see two ways to address this:
- Modify the `graph_compute` functions to take the indices of a range of nodes to evaluate
- Modify `ggml_cgraph` for dynamic allocation, then we could reuse the nodes pointer from the original graph

The second option is probably the right one, but current code would need to be updated to use `ggml_new_graph` or similar. We would also need to the decide on the implementation of dynamically sized `ggml_cgraph`, which is not very clear at the moment.

ggerganov:
Let's discuss how to implement the second option - I've sketched some rough notes here: https://github.com/ggerganov/ggml/issues/567#issuecomment-1771203788

include/ggml/ggml.h
```diff
@@ -531,36 +533,35 @@ extern "C" {
         int n_threads;
 
         // the `n_tasks` of nodes, 1:1 mapping to cgraph nodes
-        int n_tasks[GGML_MAX_NODES];
+        int n_tasks[GGML_DEFAULT_GRAPH_SIZE]; // FIXME: need to allocate ggml_cplan dynamically?
```
slaren, line 536: The array of `n_tasks` in `ggml_cplan` would need to be dynamically allocated to account for varying graph sizes. Maybe we could just remove it and replace it with a lookup table per OP?

ggerganov: The unary ops might be problematic because some of them are parallel and some of the are not.

Can we remove `n_tasks` all together as a member, and instead convert it to a function:

```c
int ggml_cplan_n_tasks(struct ggml_tensor * node);
```

Don't think we need a storage for `n_tasks` - can be computed on the fly?

src/ggml.c
```diff
@@ -20747,14 +20890,11 @@ enum ggml_opt_result ggml_opt_resume(
...
-
-    struct ggml_cgraph * gf = (struct ggml_cgraph *) gfbuf->data;
-    struct ggml_cgraph * gb = (struct ggml_cgraph *) gbbuf->data;
+    struct ggml_cgraph * gf;
+    struct ggml_cgraph * gb;
 
-    *gf = ggml_build_forward (f);
-    *gb = ggml_build_backward(ctx, gf, true);
+    gf = ggml_build_forward_ctx (ctx, f, GGML_DEFAULT_GRAPH_SIZE); // TODO: make graph size configurable
+    gb = ggml_build_backward_ctx(ctx, gf, true);
```
slaren, line 20897: The `ggml_opt` interface may also need to be changed to take a graph size.

ggerganov: Ok, I will take a look tomorrow if you haven't updated it yet

ggerganov: Think we can simply extend `ggml_opt_params` with a new parameter:

```c
size_t size; // max graph size (both for forward and backward graphs)
```

The backward graph should always be bigger than the forward, so the user has to determine the size needed to store it.

Regarding how the user would determine the necessary graph size (not just for `ggml_opt` but in general), I guess it will be via trial and error. It should be fine, since we don't expect graphs to change dynamically so it is going to be a one-time effort to find the max size (given that the default value is not enough). We only have to make sure to report an error that the graph size has been exceeded when this happens.

slaren: I have been thinking about how to remove the graph sizes entirely and make this automatic, but I don't think that this can be done as long as graphs are allocated in a context, because otherwise we would just be moving the problem to the context allocation (ie. even if the user doesn't need to specify the size of the graph, they would still need to calculate the size to determine how much memory to allocate for the context). So if we wanted to do this, we would need to allocate the graphs in the heap with `malloc`/etc, and to avoid reallocations between evaluations, the graph object would need to be reused between evaluations.

We should probably do this at some point, while I agree that it is not a huge problem to find the size of a graph, it shouldn't be necessary, and generally it feels a bit archaic to work with fixed size containers.

examples/gpt-2/main.cpp
```diff
@@ -872,30 +951,34 @@ int main(int argc, char ** argv) {
...
 
         // create the worst case graph for memory usage estimation
         int n_tokens = std::min(model.hparams.n_ctx, params.n_batch);
         int n_past = model.hparams.n_ctx - n_tokens;
-        struct ggml_cgraph * gf = gpt2_graph(model, allocr, n_past, std::vector<gpt_vocab::id>(n_tokens, 0));
+        struct ggml_cgraph * gf = gpt2_graph(model, n_past, std::vector<gpt_vocab::id>(n_tokens, 0));
 
-        // compute the required memory
-        size_t mem_size = ggml_allocr_alloc_graph(allocr, gf);
+        ggml_backend_sched_init_measure(sched, gf);
```
ggerganov, line 966: This also works, correct? Just for my understanding.

```diff
diff --git a/examples/gpt-2/main.cpp b/examples/gpt-2/main.cpp
index 1a24925..a62a2b9 100644
--- a/examples/gpt-2/main.cpp
+++ b/examples/gpt-2/main.cpp
@@ -956,13 +956,13 @@ int main(int argc, char ** argv) {
     ggml_backend_sched_t sched;
     {
         // initialize the scheduler
-        sched = ggml_backend_sched_new(model.backends.data(), model.backends.size());
-
         // create the worst case graph for memory usage estimation
         int n_tokens = std::min(model.hparams.n_ctx, params.n_batch);
         int n_past = model.hparams.n_ctx - n_tokens;
         struct ggml_cgraph * gf = gpt2_graph(model, n_past, std::vector<gpt_vocab::id>(n_tokens, 0));
 
+        sched = ggml_backend_sched_new(model.backends.data(), model.backends.size());
+
         ggml_backend_sched_init_measure(sched, gf);
 
```

slaren: Yes, that should work since it pre-allocates the inputs from a different buffer, so the `gpt2_graph` doesn't need the sched object.

However, the previous pattern of allocating the inputs manually by calling `ggml_allocr_alloc` is still supported by using `ggml_backend_sched_get_tallocr` to obtain the allocators for each backend. In that case, the sched object would need to be created before the graph.

include/ggml/ggml-backend.h
```diff
@@ -133,11 +65,72 @@ extern "C" {
...
 
+
+    //
+    // Backend scheduler
+    //
+
+    // The backend scheduler allows for multiple backends to be used together
+    // Handles compute buffer allocation, assignment of tensors to backends, and copying of tensors between backends
+    // The backends are selected based on:
+    // - the backend that supports the operation
```
ggerganov, line 81: > - the backend that supports the operation

Can you explain how this part works? How does the scheduler know which ops are supported

slaren: This isn't implemented yet (this is the automatic fallback to the CPU backend), but it will do this by calling the `supports_op` function of the backend interface.

ggerganov:
Wow, this looks very powerful. I'm still going through the code, but overall I really like what I see.

> Using multiple CUDA or OpenCL devices by splitting the computation at a layer or tensor level

I understand the multi-device layer split, but how do you imagine tensor level splits to work?

I see that https://github.com/ggerganov/ggml/issues/549 https://github.com/ggerganov/ggml/issues/567 are resolved which is great. What is the next thing that should be resolved for the backend interface to function better / be easier to use?

Shall we look into merging and using this in `llama.cpp` soon?

slaren:
> I understand the multi-device layer split, but how do you imagine tensor level splits to work?

I think that we are talking about the same thing here. What I mean by "tensor level" is that each tensor can be individually assigned to a different backend or device. What I think you mean here is what I would call splitting at the row level, which is what `ggml-cuda` does currently. I don't think that this can be realistically done with a generic backend interface because the synchronization needs to be implemented at a lower level than what `ggml-backend` supports.

> Shall we look into merging and using this in `llama.cpp` soon?

It shouldn't take long, but there are still a few steps missing before this can be used in `llama.cpp`, at least:
- Automatic fallback to CPU support
- `ggml-opencl` support
- Improved graph splitting logic
- Support for the current row-level multi GPU implementation of `ggml-cuda` in `ggml-backend` (note this will be specific to `ggml-cuda`, not generic for any backend)

I would also like to implement a backend registry so that applications can query the list of available backends in a generic way. Ideally this would allow building llama.cpp with both CUDA and OpenCL and check availability at run time, and even mix CUDA and OpenCL devices.

ggerganov:
The refactoring in https://github.com/ggerganov/llama.cpp/pull/3837 should still be helpful, correct? At least it moves the CUDA specifics outside the build functions which is inline with this interface. The `cb()` stuff and name lookups can easily be removed when we no longer need them.

slaren:
I think most of it will still be useful. Ideally, the graph splitting logic should be good enough that we won't need to assign tensors to a specific backend (and I think that's possible to implement, but it isn't quite there yet, the splits that it generates are not optimal). If necessary, it is still possible to assign manually tensors to a specific backend with `ggml_backend_sched_set_node_backend`.

ggerganov (approved):

ggerganov:
@slaren Planning to do one more sync between `ggml` and `llama.cpp`. We don't expect any issues with the new dynamic graphs, correct? I'll just apply the same changes as in this PR

ggerganov:
I think only the graph copying in the training examples is left to update in the sync PR, but I'm not sure what it refers to.
Can you point me to it?

slaren:
To perform a deep copy of a graph we need to use now `ggml_graph_cpy`, as the operator = now only copies the pointers. I think these need to be updated to use `ggml_graph_cpy`:

https://github.com/ggerganov/llama.cpp/blob/16e819d53ce5bb7025a545bd45b6404c16f3d432/examples/finetune/finetune.cpp#L775
https://github.com/ggerganov/llama.cpp/blob/16e819d53ce5bb7025a545bd45b6404c16f3d432/examples/train-text-from-scratch/train-text-from-scratch.cpp#L439C8-L439C8

Issue #589, opened 2023-10-21
test-conv-transpose fails when building with sanitizers enabled


slaren:
Not sure if this is something specific to my machine or a compiler bug, but I noticed that when building with sanitizers `test-conv-transpose` fails.

<details>
<summary>With sanitizers</summary>

 ```bash
$ rm -fr build; cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug -DGGML_SANITIZE
_ADDRESS=ON -DGGML_SANITIZE_UNDEFINED=ON
-- The C compiler identification is GNU 12.3.0
-- The CXX compiler identification is GNU 12.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Git: /usr/bin/git (found version "2.39.2")
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- CMAKE_SYSTEM_PROCESSOR: x86_64
-- x86 detected
-- Linux detected
-- x86 detected
-- Linux detected
-- Configuring done
-- Generating done
-- Build files have been written to: /home/diego/code/ggml/build

$ cmake --build build -t test-conv-transpose && build/bin/test-conv-transpose
[ 50%] Building C object src/CMakeFiles/ggml.dir/ggml-backend.c.o
[ 50%] Building C object src/CMakeFiles/ggml.dir/ggml-alloc.c.o
[ 50%] Building C object src/CMakeFiles/ggml.dir/ggml.c.o
[ 66%] Linking C shared library libggml.so
[ 66%] Built target ggml
[ 83%] Building C object tests/CMakeFiles/test-conv-transpose.dir/test-conv-transpose.c.o
[100%] Linking C executable ../bin/test-conv-transpose
[100%] Built target test-conv-transpose
GGML_ASSERT: /home/diego/code/ggml/tests/test-conv-transpose.c:55: expected == actual
[...]

PR #591, opened 2023-10-23
`fix`: memset dst to 0 in `ggml_conv_transpose_1d`

PABannier:
I came across a nasty bug in the Encodec.cpp library using `ggml_conv_transpose_1d`.

If the `dst->data` is not `memset` to 0, pre-existing values can be stored there, resulting in the wrong output.
This is due to the fact that values are accumulated into `dst->data`: `dst_data[i10*s0 + i00] += v;`.

This PR introduces two fixes: 
- [x] `memset` of `dst->data` to 0
- [x] Fix wrong indexation of `wdata` buffer in `ggml_compute_forward_conv_transpose_1d_f32`.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -14850,6 +14850,8 @@ static void ggml_compute_forward_conv_transpose_2d(
             }
         }
 
+        memset(dst->data, 0, ggml_nbytes(dst));
+
         return;
     }
 
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -14315,6 +14315,9 @@ static void ggml_compute_forward_conv_transpose_1d_f16_f32(
             }
         }
 
+        // need to zero dst since we are accumulating into it
+        memset(dst->data, 0, ggml_nbytes(dst));
+
         return;
     }
 
@@ -14405,6 +14408,9 @@ static void ggml_compute_forward_conv_transpose_1d_f32(
                 }
             }
         }
+
+        // need to zero dst since we are accumulating into it
+        memset(dst->data, 0, ggml_nbytes(dst));
 
         return;
     }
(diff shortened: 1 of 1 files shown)


slaren:
https://github.com/ggerganov/ggml/issues/589 may be related, but the `conv_transpose_2d` still fails. Does it need the same fix?

PABannier:
@slaren Very possible. I can push a fix for ggml_conv_transpose_2d if needed.

ggerganov (approved):

ggerganov:
Thanks for fixing this - let's fix `ggml_conv_transpose_2d` dst memset in this PR as well

PABannier:
@slaren @YavorGIvanov FYI, I just pushed a fix for `conv_transpose_2d` as well.

YavorGIvanov:
> @YavorGIvanov I wonder if this is also the cause of the issues in `sam.cpp` with `ggml-alloc` that required some of the tensors to be pre-allocated.

Probably it is indeed the reason it happened. I will update to latest ggml after this fix is merged and will verify over in the [Sam.cpp](https://github.com/YavorGIvanov/sam.cpp) repo by removing them

slaren:
I have verified that this fixes #589.

YavorGIvanov:
Yup. This was the reason I "needed"the two allocs after conv_transpose_2d op in Sam.cpp.
I update to latest ggml + removed them and everything works fine - https://github.com/YavorGIvanov/sam.cpp/commit/81002818eb0e2cb3b9a523286b067f80f8424431

FSSRepo:
@slaren In stable diffusion, something similar also happened when using the im2col function because it doesn't usually traverse the entire destination array. It is assumed that dst will always have values of 0, but I've encountered non-random values (they are always the same, and I'm not sure why this happens; it's possible that a kernel accesses beyond the allowed address and writes extra data). When I use `cudaMemset`, conv2d gives me the correct values.

PR #607, opened 2023-11-13
sync : llama.cpp (CUDA ReLU, CPU-only with CUDA, bloom fix, etc)

ggerganov:
Also reverts the CUDA memory pull stuff, which was synced last time, but was later reverted in `llama.cpp`

copilot:all

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1371,8 +1371,13 @@ extern "C" {
             int                   n_dims,
             int                   mode,
             int                   n_ctx,
+            int                   n_orig_ctx,
             float                 freq_base,
             float                 freq_scale,
+            float                 ext_factor,
+            float                 attn_factor,
+            float                 beta_fast,
+            float                 beta_slow,
             float                 xpos_base,
             bool                  xpos_down);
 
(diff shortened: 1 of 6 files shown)


slaren (approved):

FSSRepo:
@slaren I think `ggml_init_cublas()` should be removed from `ggml_init(...)` by now.  I was creating a model converter for stable-diffusion.cpp. The only thing I need is the gguf module, but it requires me to create a `ggml_context` to add tensors to the `gguf_context`. However, when it compiles, `ggml_init` calls `ggml_init_cublas` unnecessarily, leading to the error mentioned in [this comment](https://github.com/leejet/stable-diffusion.cpp/pull/75#issuecomment-1806832790).

ggerganov:
With this change, you can now run with `CUDA_VISIBLE_DEVICES=-1` and `ggml_init_cublas()` should do nothing in this case

slaren:
Once we move llama.cpp to ggml-backend, we will be able to remove the ggml-cuda calls from ggml.c and make the backend more self-contained, but as it is now it is required to support the "old" way to use ggml-cuda. But as @ggerganov says it, it should be possible to use a CUDA build for CPU-only after https://github.com/ggerganov/llama.cpp/pull/3946

Issue #578, opened 2023-10-13
ggml : improve memory allocation for weights and similar lists of tensors


slaren:
There are several patterns used to allocate memory for a list of fixed size tensor, such as model weights:

- Manually calculating the number of elements of each tensor and adding it all up
- Creating the tensors in a `no-alloc` context, adding them to list or map, or obtaining them by name from a `ggml_context` with `ggml_get_tensor`, summing their sizes and finally allocating them (the last one is $O(N^2)$ )
- Creating the tensors in a `no-alloc` context, allocate the weights manually with `ggml-alloc`, first with a measure allocator and then again with the exact memory requirements (current llama.cpp `finetune`)
- Creating the tensors in a `no-alloc` context, then enumerating the tensors in the context and summing their sizes (new `finetune` in https://github.com/ggerganov/llama.cpp/pull/3605)
- Create a `ggml_context` with a lot of memory and hope for the best

This becomes significantly more complicated when the weights have to be split between different backends (current llama.cpp and `ggml-backend` wip).

For something so basic, this is a lot more complicated than it should, and we should have a normalized way to do this. At the most basic level, it could be simply a function to automatically allocate all the tensors created in a `no-alloc` context with the exact memory requirements. Support for multiple backends will be more complicated.

This could also be useful for debugging operations in compute contexts, where it might be desirable to allocate memory for every tensor in the graph to be able to inspect the results of each op later.

ggerganov:
Yes, we should consolidate the different ways of allocating memory. 

> At the most basic level, it could be simply a function to automatically allocate all the tensors created in a no-alloc context with the exact memory requirements. 

Either this, or even just a function that returns the required memory for a context by doing a similar loop as the one in https://github.com/ggerganov/llama.cpp/pull/3605 would be helpful.

slaren:
I consider this fixed with `ggml_backend_alloc_ctx_tensors`.

PR #620, opened 2023-11-23
ggml-backend update: buffer types, backend registry, graph compare, tests

slaren:
- [x] Buffer types
- [x] Backend registry
- [x] Graph copy & compare between backends
- [x] `ggml_backend_alloc_ctx_tensors`
- [x] `ggml_unary_op_name` and `ggml_op_desc`
- [x] Backend buffer tests
- [x] Backend op tests
- [x] Update Metal backend
- [x] CUDA multiple device support

## Buffer types
- Buffers are no longer tied to a backend instance
- A buffer may be used in multiple backends. In systems with unified memory or integrated GPUs, this can allow fallback to CPU without copies
- Backends may add different types of buffers, even for different backends. For example, ggml-cuda adds a host pinned buffer that can be used with the CPU backend for faster transfer of inputs and outputs between system and device memory.

## Backend registry

Allows enumerating and initializing the available backends.
```C
// enumerate backends
for (size_t i = 0; i < ggml_backend_reg_get_count(); i++) {
    printf("Backend %zu/%zu (%s)\n", i + 1, ggml_backend_reg_get_count(), ggml_backend_reg_get_name(i));
    ggml_backend_buffer_t buf = ggml_backend_buft_alloc_buffer(ggml_backend_reg_get_default_buffer_type(i), size);
    ggml_backend_t backend = ggml_backend_reg_init_backend(i, NULL);
    // ...
}

// initialize a backend by name
ggml_backend_t backend = ggml_backend_reg_init_backend_from_str("CUDA0"); // cuda device 0
```

## Graph copy & compare between backends

Copy a graph to a different backend and evaluate it on both one op at a time. A callback can be used to compare the results of each operation.

```C++
// graph is allocated on the CUDA backend: copy and compare with CPU backend
auto eval_callback = [](int index, struct ggml_tensor * t1, struct ggml_tensor * t2, void * user_data) {
    if (t1->type == GGML_TYPE_F32) {
        std::vector<float> v1(GGML_PAD(ggml_nbytes(t1), sizeof(float)) / sizeof(float));
        std::vector<float> v2(GGML_PAD(ggml_nbytes(t2), sizeof(float)) / sizeof(float));

        ggml_backend_tensor_get(t1, v1.data(), 0, ggml_nbytes(t1));
        ggml_backend_tensor_get(t2, v2.data(), 0, ggml_nbytes(t2));

        for (size_t i = 0; i < v1.size(); ++i) {
            if (fabsf(v1[i] - v2[i]) > 1e0) {
                printf("[%d] %s [%s]: mismatch at index %zu: %f != %f\n", index, t1->name, ggml_op_desc(t1), i, v1[i], v2[i]);
                return false; // stop graph eval on first error
            }
        }
    }
    return true;
};

ggml_backend_compare_graph_backend(model.backend, backend_cpu, gf, eval_callback, nullptr);
```

## `ggml_backend_alloc_ctx_tensors`

Allocate all the tensors in a context and a backend buffer in one step. Ref: #578 

```C++
model.ln_f_g = ggml_new_tensor_1d(ctx, GGML_TYPE_F32, n_embd);
model.ln_f_b = ggml_new_tensor_1d(ctx, GGML_TYPE_F32, n_embd);

model.wte     = ggml_new_tensor_2d(ctx, wtype,         n_embd, n_vocab);
model.wpe     = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, n_embd, n_ctx);
model.lm_head = ggml_new_tensor_2d(ctx, wtype,         n_embd, n_vocab);
// etc

// allocate the model tensors in a backend buffer
model.buffer_w = ggml_backend_alloc_ctx_tensors(ctx, model.backend);
```

## `ggml_unary_op_name` and `ggml_op_desc`

`ggml_op_desc` can be used as a replacement of `ggml_op_name` that also returns the name of the unary op when the op is `GGML_OP_UNARY`.

## CUDA multiple device support

`ggml_backend_cuda_init` takes a `int device` parameter that specifies the CUDA device to use. To use the default device, pass `0`. Each device is registered as a different backend in the backend registry, with names `CUDA0` for device 0, `CUDA1` for device 1 and so on.

## Backend op tests

Each op implemented in the backends is tested against the CPU backend. Tensors are initialized with random data, and the result of the op is compared using a normalized MSE. New ops implemented in the backends should add a test in `test-backend-ops.cpp`. Only F16/F32 tests for now, quantized types are not yet supported in the test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
 
     - name: Configure CMake
       working-directory: ./build
-      run: cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DGGML_TEST_COVERAGE=ON -DGGML_METAL=ON ..
+      run: cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DGGML_TEST_COVERAGE=ON ..
 
     - name: Build
       working-directory: ./build
@@ -67,7 +67,7 @@ jobs:
 
     - name: Test
       working-directory: ./build
-      run: GGML_METAL_PATH_RESOURCES=./bin ctest --verbose --timeout 900
+      run: ctest --verbose --timeout 900
 
     - name: Test Coverage
       working-directory: ./build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Dependencies
       run: |
         wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | sudo tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null
-        echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list  
+        echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | sudo tee /etc/apt/sources.list.d/oneAPI.list
         sudo apt-get update
         sudo apt-get install -y --no-install-recommends llvm intel-oneapi-runtime-opencl intel-oneapi-runtime-compilers libclblast-dev
     - name: Create Build Environment
@@ -67,7 +67,7 @@ jobs:
 
     - name: Test
       working-directory: ./build
-      run: ctest --verbose --timeout 900
+      run: GGML_METAL_PATH_RESOURCES=./bin ctest --verbose --timeout 900
 
     - name: Test Coverage
       working-directory: ./build
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -94,6 +94,10 @@ function gg_run_ctest_debug {
     (time cmake -DCMAKE_BUILD_TYPE=Debug ${CMAKE_EXTRA} ..     ) 2>&1 | tee -a $OUT/${ci}-cmake.log
     (time make -j                                              ) 2>&1 | tee -a $OUT/${ci}-make.log
 
+    if [ ! -z ${GG_BUILD_METAL} ]; then
+        export GGML_METAL_PATH_RESOURCES="$(pwd)/bin"
+    fi
+
     (time ctest --output-on-failure -E test-opt ) 2>&1 | tee -a $OUT/${ci}-ctest.log
 
     set +e
@@ -121,6 +125,10 @@ function gg_run_ctest_release {
 
     (time cmake -DCMAKE_BUILD_TYPE=Release ${CMAKE_EXTRA} ..   ) 2>&1 | tee -a $OUT/${ci}-cmake.log
     (time make -j                                              ) 2>&1 | tee -a $OUT/${ci}-make.log
+
+    if [ ! -z ${GG_BUILD_METAL} ]; then
+        export GGML_METAL_PATH_RESOURCES="$(pwd)/bin"
+    fi
 
     if [ -z $GG_BUILD_LOW_PERF ]; then
         (time ctest --output-on-failure ) 2>&1 | tee -a $OUT/${ci}-ctest.log
--- a/examples/gpt-2/main-backend.cpp
+++ b/examples/gpt-2/main-backend.cpp
@@ -177,47 +177,6 @@ bool gpt2_model_load(const std::string & fname, gpt2_model & model, gpt_vocab & 
 
     auto & ctx = model.ctx;
 
-    size_t buffer_size = 0;
-
-    {
-        const auto & hparams = model.hparams;
-
-        const int n_embd  = hparams.n_embd;
-        const int n_layer = hparams.n_layer;
-        const int n_ctx   = hparams.n_ctx;
-        const int n_vocab = hparams.n_vocab;
-
-        buffer_size += n_embd*ggml_type_sizef(GGML_TYPE_F32); // ln_f_g
-        buffer_size += n_embd*ggml_type_sizef(GGML_TYPE_F32); // ln_f_b
-
-        buffer_size += n_vocab*n_embd*ggml_type_sizef(wtype);         // wte
-        buffer_size +=   n_ctx*n_embd*ggml_type_sizef(GGML_TYPE_F32); // wpe
-        buffer_size += n_vocab*n_embd*ggml_type_sizef(wtype);         // lm_head
-
-        buffer_size += n_layer*(n_embd*ggml_type_sizef(GGML_TYPE_F32)); // ln_1_g
-        buffer_size += n_layer*(n_embd*ggml_type_sizef(GGML_TYPE_F32)); // ln_1_b
-
-        buffer_size += n_layer*(n_embd*ggml_type_sizef(GGML_TYPE_F32)); // ln_2_g
-        buffer_size += n_layer*(n_embd*ggml_type_sizef(GGML_TYPE_F32)); // ln_2_b
-
-        buffer_size += n_layer*(3*n_embd*n_embd*ggml_type_sizef(wtype));         // c_attn_attn_w
-        buffer_size += n_layer*(       3*n_embd*ggml_type_sizef(GGML_TYPE_F32)); // c_attn_attn_b
-
-        buffer_size += n_layer*(n_embd*n_embd*ggml_type_sizef(wtype));           // c_attn_proj_w
-        buffer_size += n_layer*(       n_embd*ggml_type_sizef(GGML_TYPE_F32));   // c_attn_proj_b
-
-        buffer_size += n_layer*(4*n_embd*n_embd*ggml_type_sizef(wtype));         // c_mlp_fc_w
-        buffer_size += n_layer*(       4*n_embd*ggml_type_sizef(GGML_TYPE_F32)); // c_mlp_fc_b
-
-        buffer_size += n_layer*(4*n_embd*n_embd*ggml_type_sizef(wtype));         // c_mlp_proj_w
-        buffer_size += n_layer*(         n_embd*ggml_type_sizef(GGML_TYPE_F32)); // c_mlp_proj_b
-
-        buffer_size += (6 + 12*n_layer)*128; // alignment overhead
-
-        printf("%s: ggml tensor size    = %d bytes\n", __func__, (int) sizeof(ggml_tensor));
-        printf("%s: backend buffer size = %6.2f MB\n", __func__, buffer_size/(1024.0*1024.0));
-    }
-
     // create the ggml context
     {
         size_t n_tensors = 2 + 6 + 12*model.hparams.n_layer;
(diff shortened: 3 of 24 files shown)


FSSRepo:
Nice upate, Normally, one has to create the graph at every moment before computation; it cannot be reused, and simply assigning IDs to input tensors wouldn't suffice.

```c++
         ggml_allocr_reset(compute_alloc);

        struct ggml_cgraph * gf = build_graph(z, decode_graph);
        ggml_allocr_alloc_graph(compute_alloc, gf);

        if (ggml_backend_is_cpu(backend)) {
            ggml_backend_cpu_set_n_threads(backend, n_threads);
        }

        ggml_backend_graph_compute(backend, gf);

        ggml_backend_tensor_get(gf->nodes[gf->n_nodes - 1], work_result->data, 0, ggml_nbytes(work_result));
```

```c++
ggml_allocr_reset(compute_alloc);
ggml_set_input(gf, Z_INPUT, z);

ggml_allocr_alloc_graph(compute_alloc, gf);

 if (ggml_backend_is_cpu(backend)) {
          ggml_backend_cpu_set_n_threads(backend, n_threads);
}

 // gf is created once, it can be the graph used for measure buffer size
ggml_backend_graph_compute(backend, gf);

ggml_backend_tensor_get(gf->nodes[gf->n_nodes - 1], work_result->data, 0, ggml_nbytes(work_result));

```

slaren:
>Nice upate, Normally, one has to create the graph at every moment before computation; it cannot be reused, and simply assigning IDs to input tensors wouldn't suffice.

I think that what you mean here is that graphs used for measure cannot be re-used for computation, which is not very good for graphs that only need to be evaluated once. I think we could solve this by assigning offsets rather than absolute addresses to tensors (ie. in the `data` member), then it would be possible to replace the buffer used to allocate the graph without having to update every tensor. However, this would break compatibility with the CPU backend, and it would require significant changes to `ggml.c`, which is something that we have tried to avoid for now. Maybe in the future we can re-evaluate this, but it is not going to happen soon.

>ggml_backend_alloc_ctx_tensors(ctx, model.backend); Skip already allocated tensors, for example when assigning a constant tensor such as the attention query scale.

I will do this.

>ggml_backend_compare_graph_backend wonderful function, thank you, It will be very useful for debugging dequantization kernel issues since, for some reason, in stable diffusion, quantizations work fine for me in the CPU backend but not in CUDA.

The goal is indeed to help debugging issues with the backends. I already found some differences in the GELU op between the CPU and CUDA backends while testing this with the gpt-2 example, and I am sure that there are more cases. I will try to add automated tests for all the ops.

>If possible, it could be beneficial for parameter tensors to be dynamically passed in a backend (dynamic offloading) as needed during computation.

I don't really understand what you mean by this. If you prefer to use Spanish, you can send me a message to the email address in my github profile.

FSSRepo:
> some differences in the GELU op

comment `#define GGML_GELU_FP16` in ggml.c and try again

slaren:
@ggerganov is the github CI capable of using the Metal backend? I suspect that it is failing somewhere during `ggml_metal_init`.
```
19: Test command: /Users/runner/work/ggml/ggml/build/bin/test-backend-buffer
19: Working Directory: /Users/runner/work/ggml/ggml/build/tests
19: Environment variables: 
19:  LLVM_PROFILE_FILE=test-backend-buffer.profraw
19: Test timeout computed to be: 900
19: ggml_backend_register: registered backend CPU
19: ggml_backend_register: registered backend Metal
19: ggml_metal_init: allocating
19: ggml_metal_init: found device: Apple Paravirtual device
19: ggml_metal_init: picking default device: Apple Paravirtual device
19: ggml_metal_init: default.metallib not found, loading from source
19: ggml_metal_init: error: could not use bundle path to find ggml-metal.metal, falling back to trying cwd
19: ggml_metal_init: loading 'ggml-metal.metal'
19/19 Test #19: test-backend-buffer ..............***Exception: SegFault  0.04 sec
Errors while running CTest
```

ggerganov:
Hm, seems it does not work, but it does not print any meaningful error.
Probably we should just disable it for now and use `ggml-ci` for `test-backend-buffer`?

slaren:
I am not sure what is happening either, it seems that the log is missing the output from `stdout` too. Let's disable it for now then, thanks for looking into it.

slaren:
@FSSRepo can you help me define a few test cases for im2col? I need different values of parameters to test, currently it is tested with these:

```C++
// GGML_OP_IM2COL
struct test_im2col : public test_case {
    const ggml_type type_a;
    const ggml_type type_b;
    const std::array<int64_t, 4> ne_a;
    const std::array<int64_t, 4> ne_b;
    const int s0;
    const int s1;
    const int p0;
    const int p1;
    const int d0;
    const int d1;
    const bool is_2D;

    std::string vars() override {
        return VARS_TO_STR11(type_a, type_b, ne_a, ne_b, s0, s1, p0, p1, d0, d1, is_2D);
    }

    test_im2col(ggml_type type_a = GGML_TYPE_F16, ggml_type type_b = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_a = {10, 10, 10, 10},
            std::array<int64_t, 4> ne_b = {10, 10, 10, 10},
            int s0 = 1, int s1 = 1,
            int p0 = 0, int p1 = 0,
            int d0 = 1, int d1 = 1,
            bool is_2D = false)
        : type_a(type_a), type_b(type_b), ne_a(ne_a), ne_b(ne_b), s0(s0), s1(s1), p0(p0), p1(p1), d0(d0), d1(d1), is_2D(is_2D) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * a = ggml_new_tensor(ctx, type_a, 4, ne_a.data());
        ggml_tensor * b = ggml_new_tensor(ctx, type_b, 4, ne_b.data());
        ggml_tensor * out = ggml_im2col(ctx, a, b, s0, s1, p0, p1, d0, d1, is_2D);
        return out;
    }
};
```

FSSRepo:
> can you help me define a few test cases for im2col?

Sure:

```c++
// GGML_OP_IM2COL
struct test_im2col : public test_case {
    const ggml_type type_input;
    const ggml_type type_kernel;
    const std::array<int64_t, 4> ne_input;
    const std::array<int64_t, 4> ne_kernel;
    // stride
    const int s0;
    const int s1;
    // padding
    const int p0;
    const int p1;
    // dilatation
    const int d0;
    const int d1;
    // mode
    const bool is_2D;

    std::string vars() override {
        return VARS_TO_STR11(type_input, type_kernel, ne_input, ne_kernel, s0, s1, p0, p1, d0, d1, is_2D);
    }

    test_im2col(ggml_type type_input = GGML_TYPE_F16, ggml_type type_b = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_input = {10, 10, 3, 1}, // [input_width, input_height, input_channels, 1]
            std::array<int64_t, 4> ne_kernel = {3, 3, 8, 1}, // [kernel_width, kernel_height, output_channels, 1]
            int s0 = 1, int s1 = 1,
            int p0 = 0, int p1 = 0,
            int d0 = 1, int d1 = 1,
            bool is_2D = false)
        : type_a(type_input), type_b(type_kernel), ne_input(ne_input), ne_kernel(ne_kernel), s0(s0), s1(s1), p0(p0), p1(p1), d0(d0), d1(d1), is_2D(is_2D) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * input = ggml_new_tensor(ctx, type_input, 4, ne_input.data());
        ggml_tensor * kernel = ggml_new_tensor(ctx, type_kernel, 4, ne_kernel.data());
        ggml_tensor * out = ggml_im2col(ctx, input, kernel, s0, s1, p0, p1, d0, d1, is_2D);
        return out;
    }
};
```

slaren:
These tensor dimensions don't work because it fails an assert in `ggml_im2col`. However these are just the default parameters, we can add as many test cases with different values as needed.

FSSRepo:
My mistake, I gave you the wrong order of the tensors

FSSRepo:
Fixed:

```c++
// GGML_OP_IM2COL
struct test_im2col : public test_case {
    const ggml_type type_input;
    const ggml_type type_kernel;
    const std::array<int64_t, 4> ne_input;
    const std::array<int64_t, 4> ne_kernel;
    // stride
    const int s0;
    const int s1;
    // padding
    const int p0;
    const int p1;
    // dilatation
    const int d0;
    const int d1;
    // mode
    const bool is_2D;

    std::string vars() override {
        return VARS_TO_STR11(type_input, type_kernel, ne_input, ne_kernel, s0, s1, p0, p1, d0, d1, is_2D);
    }

    test_im2col(ggml_type type_input = GGML_TYPE_F16, ggml_type type_kernel = GGML_TYPE_F32,
            std::array<int64_t, 4> ne_input = {10, 10, 3, 1}, // [input_width, input_height, input_channels, 1]
            std::array<int64_t, 4> ne_kernel = {3, 3, 3, 1}, // [kernel_width, kernel_height, input_channels, 1]
            int s0 = 1, int s1 = 1,
            int p0 = 1, int p1 = 1,
            int d0 = 1, int d1 = 1,
            bool is_2D = true)
        : type_input(type_input), type_kernel(type_kernel), ne_input(ne_input), ne_kernel(ne_kernel), s0(s0), s1(s1), p0(p0), p1(p1), d0(d0), d1(d1), is_2D(is_2D) {}

    ggml_tensor * build_graph(ggml_context * ctx) override {
        ggml_tensor * input = ggml_new_tensor(ctx, type_input, 4, ne_input.data());
        ggml_tensor * kernel = ggml_new_tensor(ctx, type_kernel, 4, ne_kernel.data());
        ggml_tensor * out = ggml_im2col(ctx, kernel, input, s0, s1, p0, p1, d0, d1, is_2D);
        return out;
    }
};
```

slaren:
I had to swap the kernel and input types, but otherwise it works.

ggerganov:
Will take a detailed look tomorrow. Great work on the tests - I was just thinking about something like this. Much needed.

ggerganov:
Regarding multiple CUDA devices: is it fully functional at this point? Is there going to be a conflict around the shared memory pool if it is used by more than one device or thanks to the lock everything should be good?

With this infrastructure, if let's say I wanted to implement an alternative backend for an existing GPU (for example, `ggml-cuda-2.cu` or `ggml-metal-2.m`), I can just register it as a new backend and both implementations should be usable together in a single binary, correct?

ggerganov:
How come none of the `MUL_MAT` tests pass for the Metal backend:

```
Error: MUL_MAT: NMSE = 5.200507
  MUL_MAT(type_a=f32,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[1,1]): FAIL
Error: MUL_MAT: NMSE = 5.312463
  MUL_MAT(type_a=f32,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[2,1]): FAIL
Error: MUL_MAT: NMSE = 7.459601
  MUL_MAT(type_a=f32,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[1,2]): FAIL
Error: MUL_MAT: NMSE = 7.485240
  MUL_MAT(type_a=f32,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[2,2]): FAIL
Error: MUL_MAT: NMSE = 5.304665
  MUL_MAT(type_a=f16,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[1,1]): FAIL
Error: MUL_MAT: NMSE = 5.417071
  MUL_MAT(type_a=f16,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[2,1]): FAIL
Error: MUL_MAT: NMSE = 7.481416
  MUL_MAT(type_a=f16,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[1,2]): FAIL
Error: MUL_MAT: NMSE = 7.437984
  MUL_MAT(type_a=f16,type_b=f32,m=32,n=32,k=32,bs=[10,10],nr=[2,2]): FAIL
```

ggerganov (approved):
Great stuff! Merge at will

PR #625, opened 2023-11-30
ggml : full broadcast in mul, add, div + ggml_mul_mat_id, ggml_argsort, ggml_top_k

slaren:
Ref: #621

The CPU implementation should be done, I will add CUDA and Metal next.

PR #629, opened 2023-12-04
false alarm of error check in `ggml_new_tensor_impl()`

foldl:
In the checking the condition `data_size + view_offs <= ggml_nbytes(view_src)`, `nb[]` is not considered. So the assertion is triggered when, for example, broadcasting a vector into a matrix:

```c
// bias is a vector
// output is a matrix
ggml_tensor *bcast_bias = ggml_view_2d(ctx, bias, output->ne[0], output->ne[1], 0, 0);
```

Maybe this error checking can be removed?

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -2488,8 +2488,6 @@ static struct ggml_tensor * ggml_new_tensor_impl(
         data_size *= ne[i];
     }
 
-    GGML_ASSERT(view_src == NULL || data_size + view_offs <= ggml_nbytes(view_src));
-
     void * data = view_src != NULL ? view_src->data : NULL;
     if (data != NULL) {
         data = (char *) data + view_offs;


foldl:
I am not sure about the meaning of "utilize the broadcast functionality of the ggml operators". Does it mean a dedicated broadcast function, or some operators will do the broadcast internally?

In my use case, `bcast_bias` adds to another matrix `output`. As far as I can tell, `ggml_add_inplace`
requires two inputs to be `ggml_are_same_shape()` and does not broadcast one of its operand automatically.

https://github.com/foldl/chatllm.cpp/blob/b769e17fe1b8a432a6973e0e6051b49d5812d492/layers.cpp#L35C1-L37C1

By the way, it does work without this `GGML_ASSERT` in `ggml_new_tensor_impl()`.

ggerganov:
The `ggml_add()` and `ggml_add_inplace()` operators will automatically broadcast the second operand `b` into the fist `a`. This should work too, without changes in `ggml`:

```c++
            // old
            ggml_tensor *bcast_bias = ggml_view_2d(ctx->gctx.get(), bias, output->ne[0], output->ne[1], 0, 0);
            output = ggml_add_inplace(ctx->gctx.get(), output, bcast_bias);

            // new
            output = ggml_add_inplace(ctx->gctx.get(), output, bias);
```

The `ggml_are_same_shape()` requirement is only necessary when doing backwards pass, which is not the case during inference.

On current `master`, we require that the `b` tensor has the same row shape as `a` - the rest of the dimensions will be automatically broadcasted internally in the implementation:

https://github.com/ggerganov/ggml/blob/a399be09771d9ac56fad7f81dae40df391286a62/src/ggml.c#L3181-L3210

After we merge #625, then `ggml_add` will also broadcast across the row dimension (but this does not seem relevant to your case).

Let us know if the above change works for you or there are still any issues.

foldl:
@ggerganov Yes, it works for me. It's brilliant.

Back to `ggml_view_2d`, this PR is not needed any more, although I think this error checking could be improved, or just _removed_, :)

PR #634, opened 2023-12-05
test-backend-ops : initialize ggml_argsort test with unique values

slaren:
In rare (but still surprisingly common) cases, the randomly initialized tensor could contain rows with repeated values, which would result in multiple valid sorts, and that was causing the test to fail sometimes because the backend sort would not be exactly the same as the CPU sort. I guess there aren't that many 32-bit floats.


ggerganov (approved):
Huh, surprising

PR #640, opened 2023-12-07
sync : llama.cpp (fused soft max, gpu cpy ops, etc.)

ggerganov:
ggml-ci

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -4884,7 +4884,17 @@ struct ggml_tensor * ggml_diag_mask_zero_inplace(
 static struct ggml_tensor * ggml_soft_max_impl(
         struct ggml_context * ctx,
         struct ggml_tensor  * a,
+        struct ggml_tensor  * mask,
+        float                 scale,
         bool                  inplace) {
+    GGML_ASSERT(ggml_is_contiguous(a));
+    if (mask) {
+        GGML_ASSERT(ggml_is_contiguous(mask));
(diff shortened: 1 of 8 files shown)


examples/whisper/whisper.cpp
```diff
@@ -1078,6 +1078,11 @@ static ggml_backend_t whisper_backend_init(const whisper_context_params & params
         if (!backend_gpu) {
             WHISPER_LOG_ERROR("%s: ggml_backend_metal_init() failed\n", __func__);
         }
+        if (!ggml_backend_metal_supports_family(backend_gpu, 7)) {
```
slaren, line 1081: This may crash if `ggml_backend_metal_init` fails and `backend_gpu` is `NULL`.

src/ggml.c
```diff
@@ -8446,7 +8469,7 @@ static void ggml_compute_forward_concat_f32(
     GGML_ASSERT(nb10 == sizeof(float));
 
     for (int i3 = 0; i3 < ne3; i3++) {
-        for (int i2 = ith; i2 < ne2; i2++) {
+        for (int i2 = ith; i2 < ne2; i2 += nth) {
```
ggerganov, line 8472: @FSSRepo FYI there was a bug in `ggml_concat`

FSSRepo: This already solves it anyway.

ggerganov: Yes - waiting for CI to pass and will merge

PR #650, opened 2023-12-14
Completing default values for enumerations.

GitEventhandler:
The values of an enum are often used to define constants in a program, which will remain unchanged throughout the program's entire lifecycle. Therefore, manually defining the specific values of an enum will make the code easier to read, porting and prevent accidental modification that could lead to compatibility issues later on.


GitEventhandler:
The purpose of doing this is: when ggml needs to be ported to other programming languages environments, such as ggml-python, programmers do not need to guess the specific values of the enumeration. For example, the following code comes from [ggml-python](https://github.com/abetlen/ggml-python):
```python
# enum ggml_object_type {
#     GGML_OBJECT_TENSOR,
#     # If an enumeration value is inserted here, ggml-python will need to be midified
#     GGML_OBJECT_GRAPH,
#     GGML_OBJECT_WORK_BUFFER
# };
GGML_OBJECT_TENSOR = 0
GGML_OBJECT_GRAPH = 1
GGML_OBJECT_WORK_BUFFER = 2
```

gwenzek:
this would be useful when debugging python code calling GGML.

PR #654, opened 2023-12-14
ggml: ggml_conv_2d with winograd convolution implementation + ggml_pad symmetric padding

FSSRepo:
Currently, im2col and GEMM are used for performing convolutions, which perform well in terms of computation and memory usage for small kernels and not-so-large inputs. However, there are cases where the ggml_conv_2d operation cannot handle very large input sizes because im2col significantly increases memory usage. This is where the use case for the Winograd convolution method comes in. It supports only 3x3 kernels, which are the most common, and reduces the number of operations and memory usage by 40%.

This implementation still requires improving the distribution of computation among threads.

Tasks:

- [ ] Efficiently implement the CPU backend. Currently, kernel processing is done in a single thread; the ideal scenario is to use multithreading. However, it would be necessary to split the Winograd operation into two stages `GGML_OP_WINOGRAD_STAGE0` process kernel reduction, `GGML_OP_WINOGRAD_STAGE1` apply processed kernel to image.
- [ ] Implement the CUDA and Metal kernels and evaluate the speed compared to im2col.


FSSRepo:
@ggerganov To avoid the need for 2 stages in this operation, `GGML_TASK_INIT` should be multithreaded, or there should be something equivalent to syncthreads to perform the two operations (process kernel + apply kernel) in multiple threads without a race condition.

ggerganov:
The CPU performance is likely poor because it does not utilize SIMD. One would normally expect that such type of computations will be efficiently vectorized by the compiler, but my experience is that this is not the case.

If you measure x2 times performance drop with this implementation, I can imaging that it will be faster if it is vectorized properly. However, I'm not sure if it is worth doing it on the CPU because there will be a lot of extra code for the various instructions sets (AVX, ARM, etc.)

Let's see if you observe the performance gain with CUDA and can decide how to proceed after that.

FSSRepo:
This technique definitely does not improve the performance of `ggml_conv_2d`. It's a pity since it allows for convolutions with 60% less memory usage. I'm not sure what the approach would be to optimize it and achieve decent performance. Increasing kernel threads to perform more operations did not improve anything.

## Test with UNet

im2col

```bash
[       ADD] - 22.238001 ms - 419 - 0.053074 ms
[       MUL] - 9.046000 ms - 125 - 0.072368 ms
[    CONCAT] - 2.272000 ms - 12 - 0.189333 ms
[      NORM] - 2.518000 ms - 48 - 0.052458 ms
[GROUP_NORM] - 5.532000 ms - 61 - 0.090689 ms
[   MUL_MAT] - 120.073997 ms - 329 - 0.364967 ms
[     SCALE] - 2.232000 ms - 32 - 0.069750 ms
[      CONT] - 10.939000 ms - 128 - 0.085461 ms
[  SOFT_MAX] - 51.769001 ms - 32 - 1.617781 ms
[    IM2COL] - 65.295998 ms - 97 - 0.673155 ms
[   UPSCALE] - 0.566000 ms - 3 - 0.188667 ms
[     UNARY] - 4.599000 ms - 84 - 0.054750 ms
Total Time: 297.080994 ms
```

winograd

```bash
[       ADD] - 25.370001 ms - 419 - 0.060549 ms
[       MUL] - 9.018000 ms - 125 - 0.072144 ms
[    CONCAT] - 2.445000 ms - 12 - 0.203750 ms
[      NORM] - 2.670000 ms - 48 - 0.055625 ms
[GROUP_NORM] - 5.533000 ms - 61 - 0.090705 ms
[   MUL_MAT] - 101.792000 ms - 281 - 0.362249 ms
[     SCALE] - 2.480000 ms - 32 - 0.077500 ms
[      CONT] - 12.264000 ms - 128 - 0.095812 ms
[  SOFT_MAX] - 58.417999 ms - 32 - 1.825562 ms
[    IM2COL] - 11.237000 ms - 49 - 0.229327 ms
[   UPSCALE] - 0.599000 ms - 3 - 0.199667 ms
[       PAD] - 6.317000 ms - 48 - 0.131604 ms
[WINOGRAD_STAGE0] - 99.783997 ms - 48 - 2.078833 ms
[WINOGRAD_STAGE1] - 9505.807617 ms - 48 - 198.037659 ms
[     UNARY] - 4.805000 ms - 84 - 0.057202 ms
```

PR #662, opened 2023-12-22
sync : llama.cpp (ggml_scale, ggml_row_size, ggml_mul_mat_set_prec)

ggerganov:
- llama.cpp: https://github.com/ggerganov/llama.cpp/pull/4591
- whisper.cpp: https://github.com/ggerganov/whisper.cpp/pull/1677


ggerganov:
The PRs should be ready to review - will merge later today

slaren (approved):

Issue #668, opened 2023-12-28
test-conv1d and test-conv2d failed on GPUs with computation capability <= 6.1


bssrdf:
Just to report right now two test cases `test-conv1d` and `test-conv2d` cannot pass if running on GPU backend and its computation capability <= 6.1. On a GTX 1070 with CUDA version 12.1, it gave 
ggml_im2col (4320): PASSED
ggml_conv2d (480): FAILED

This might be related to this section of the code
https://github.com/ggerganov/ggml/blob/3d57e767653eeaf7b3cc311bdc4ff24771be1ee7/src/ggml-cuda.cu#L7618-L7639
Here `src1` is not converted to FP32 while `src0` is.

PR #669, opened 2023-12-28
Fix mul-mat error for older GPUs

bssrdf:
This PR fixed [issue 668](https://github.com/ggerganov/ggml/issues/668). The two test cases `test-conv1d` and `test-conv2d` passed with this PR on a GTX 1070 with CUDA v12.1. It also fixed [problems](https://github.com/leejet/stable-diffusion.cpp/issues/95) downstream in other projects which use `ggml`.


src/ggml-cuda.cu
```diff
@@ -7615,16 +7615,25 @@ static void ggml_cuda_op_mul_mat_cublas(
         const to_fp32_cuda_t to_fp32_cuda = ggml_get_to_fp32_cuda(GGML_TYPE_F16);
         to_fp32_cuda(dst_f16.get(), dst_dd_i, row_diff*src1_ncols, stream);
     }
-    else {
+    else {
```
Green-Sky, line 7618: please fix this trailing whitespace

Green-Sky (changes requested):

src/ggml-cuda.cu
```diff
@@ -7634,7 +7643,7 @@ static void ggml_cuda_op_mul_mat_cublas(
             cublasSgemm(g_cublas_handles[id], CUBLAS_OP_T, CUBLAS_OP_N,
                     row_diff, src1_ncols, ne10,
                     &alpha, src0_ddf_i, ne00,
-                            src1_ddf_i, ne10,
+                            src1_ddf1_i, ne10,
                     &beta,  dst_dd_i,   ldc));
```
Cyberhan123, line 7647:
```suggestion
CUBLAS_CHECK(
            cublasGemmEx(g_cublas_handles[id], CUBLAS_OP_T, CUBLAS_OP_N,
                    row_diff, src1_ncols, ne10,
                    &alpha, src0_ddf_i, CUDA_R_32F, ne00,
                                src1_ddf1_i, CUDA_R_32F, ne10,
                    &beta,   dst_ddf1, CUDA_R_32F, ldc,
                    CUBLAS_COMPUTE_32F,
                    CUBLAS_GEMM_DEFAULT_TENSOR_OP));
```

Cyberhan123: @Green-Sky Can you test the modifications I suggested above? I am not a professional CUDA developer, but after my testing, such modifications will greatly reduce the probability of bad images. However, I am not sure whether there is a compatibility problem with cublasGemmEx.

Green-Sky: where does the `dst_ddf1` come frome?
```
/build/xxx-source/ggml/src/ggml-cuda.cu(7456): error: identifier "dst_ddf1" is undefined
```

Cyberhan123: Sorry it‘s `dst_dd_i `

slaren: I don't think this is necessary.

Green-Sky: i cant really see any significant correlation to the issues i observe. Since they look like synchronization issues, a slightly different invocation can accidentally make them go away or less likely.

Cyberhan123: You can use stable-diffusion.cpp
 Will `save_tensor_to_file` in [sample]( https://github.com/FSSRepo/stable-diffusion.cpp/blob/421e39b0f0aa30dcc35151f69028a7b224943196/stable-diffusion.cpp#L6258) export the tensor? The export fails and succeeds in the case of the same seed. Yes, then export the failed and successful ones at [decode](https://github.com/FSSRepo/stable-diffusion.cpp/blob/421e39b0f0aa30dcc35151f69028a7b224943196/stable-diffusion.cpp#L6272).

Cyberhan123: When I tested, the tensor in the sample and the failed tensor were equal.

Cyberhan123: But in the case of decoding, they have a large deviation. After I added the correction of cublasGemmEx, the deviation became smaller, but I am more confused as to why such a deviation occurs.

bssrdf: Thanks, @slaren, for the style fix and other updates. 
@Cyberhan123, I'll leave it to @slaren and others to decide whether replacing `cublasSgemm` with `cublasGemmEx` is a good idea. I am not an export on CUDA/CUBlas.

FSSRepo:
@bssrdf i am going to review and test this fix on my GPU, even though it works fine for me, just to ensure there is no impact on the current performance.

bssrdf:
> @bssrdf i am going to review and test this fix on my GPU, even though it works fine for me, just to ensure there is no impact on the current performance.

@FSSRepo, sounds good. Thanks for the test.

ggerganov (approved):
Thank you for digging into this and resolving the issues. I should have used more `GGML_ASSERT`s to avoid such kind of issues - will try to do so in the future

PR #673, opened 2023-12-29
cuda : ggml_mul_mat assert for padded src1

ggerganov:
Currently, the padded matrix multiplications in `whisper.cpp` are silently failing with CUDA:

https://github.com/ggerganov/ggml/blob/dbd02958fa4f46898f68ca29c27ddcdc58a06f98/examples/whisper/whisper.cpp#L224-L230

The reason is that the `to_fp16_cuda` and `to_fp32_cuda` calls assume no padding of the data. We can either assert that the data is not padded, or over-allocate a buffer accounting for the padding. The latter produces correct results, but is sub-optimal.

Drafting this PR to brainstorm some potential solutions

### DIFF
--- a/src/ggml-cuda.cu
+++ b/src/ggml-cuda.cu
@@ -8529,6 +8529,7 @@ static void ggml_cuda_mul_mat_mat_batched_cublas(const ggml_tensor * src0, const
     if (src1->type != GGML_TYPE_F16) {
         const to_fp16_cuda_t to_fp16_cuda = ggml_get_to_fp16_cuda(src1->type);
         const int64_t ne_src1 = ggml_nelements(src1);
+        GGML_ASSERT(ne_src1 == ggml_nbytes(src1)/ggml_type_size(src1->type));
         src1_f16_alloc.alloc(ne_src1);
         GGML_ASSERT(to_fp16_cuda != nullptr);
         to_fp16_cuda(src1_ddf, src1_f16_alloc.get(), ne_src1, main_stream);


slaren:
Is there any benefit to using `ggml_mul_mat_pad` with CUDA (and specially with cuBLAS)? This seems to me like it should be a implementation detail of the matrix multiplication algorithm.

ggerganov:
Can't test the `ggml_mul_mat_pad` with CUDA because it would need some extra changes to `to_fp16_cuda`. Likely there will be no benefit in this case because cuBLAS will probably make the necessary optimizations internally. For Metal it helps because that mat-mat kernel requires dim 0 to be multiple of 32

But regardless, the issue remains if users try to perform matrix multiplications with padded views

slaren:
I think that in general there is very poor support for any kind of non-contiguous tensors in ggml-cuda. We should definitely add more checks and improve support, but I am afraid that this is just the tip of the iceberg. We could extend `test-backend-ops` to automatically run tests with non-contiguous tensors for all the operations, so that we could have a better overall picture of what needs to be fixed.

slaren:
At the moment I am working on moving `llama.cpp` fully to ggml-backend with `ggml_backend_sched` for partial offloading, and bringing everything up to feature parity with the old CUDA API. After that, we can remove a lot of code, globals, and generally do a lot of refactoring of the CUDA backend.

FSSRepo:
I think the metal backend could benefit from expanding matrix multiplication support and significantly improving performance in some cases such as [comment](https://github.com/leejet/stable-diffusion.cpp/issues/108#issuecomment-1853903260).

PR #677, opened 2024-01-03
sync : llama.cpp

ggerganov:
Can't decide if we should be squashing these sync PRs or merge the commits as is.
On one hand we preserve a more detailed history, but on the other hand each commit is replicated across all synced repos.


slaren:
I think it is worth it to keep the commits. Currently, when looking at the last commit that changed a line, to understand why that change is there, many times we only get "sync". It would be useful to have the correct history.

src/ggml-cuda.cu
```diff
@@ -10032,14 +10039,22 @@ static bool ggml_backend_cuda_supports_op(ggml_backend_t backend, const ggml_ten
...
             } break;
+        case GGML_OP_DUP:
+        case GGML_OP_REPEAT:
+        case GGML_OP_CONCAT:
+            {
+                ggml_type src0_type = op->src[0]->type;
+                if (src0_type != GGML_TYPE_I32 && src0_type != GGML_TYPE_I16) {
+                    return true;
+                }
+                return false;
```
slaren, line 10050:
```suggestion
                return src0_type != GGML_TYPE_I32 && src0_type != GGML_TYPE_I16;
```

slaren (approved):

PR #679, opened 2024-01-04
fix : cuda order of synchronization when setting a buffer

Green-Sky:
we need to wait **after** the set to wait for all data to arrive.

see here https://github.com/leejet/stable-diffusion.cpp/pull/129#issuecomment-1873844723
and here https://github.com/leejet/stable-diffusion.cpp/pull/129#issuecomment-1873978488


Green-Sky:
sd.cpp without this: 
output_3

slaren:
I have added a fix for this in https://github.com/ggerganov/llama.cpp/pull/4766. I think there are still some missing synchronizations in this implementation. Since sd.cpp already has a workaround, I think we can wait until the next sync with llama.cpp.

slaren (approved):

PR #682, opened 2024-01-05
change GGML_MAX_NAME to 128

leejet:
Names of certain tensors, like those in stable-diffusion, exceed a length of 64 characters. I hope the upstream ggml can merge this pr, so I don't have to maintain a separate fork.

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -218,7 +218,7 @@
 #define GGML_MAX_PARAMS         2048
 #define GGML_MAX_CONTEXTS       64
 #define GGML_MAX_SRC            10
-#define GGML_MAX_NAME           64
+#define GGML_MAX_NAME           128
 #define GGML_MAX_OP_PARAMS      64
 #define GGML_DEFAULT_N_THREADS  4
 #define GGML_DEFAULT_GRAPH_SIZE 2048
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -218,7 +218,9 @@
 #define GGML_MAX_PARAMS         2048
 #define GGML_MAX_CONTEXTS       64
 #define GGML_MAX_SRC            10
-#define GGML_MAX_NAME           128
+#ifndef GGML_MAX_NAME
+#define GGML_MAX_NAME           64
+#endif
 #define GGML_MAX_OP_PARAMS      64
 #define GGML_DEFAULT_N_THREADS  4
 #define GGML_DEFAULT_GRAPH_SIZE 2048


Cyberhan123:
@slaren @ggerganov  Can we merge this as well?

slaren (approved):

slaren:
I hope we can find a better solution in the future, `ggml_tensor` is becoming absurdly large.

Cyberhan123:
I think if I define it in cmake, will it work?
add_definitions(-DGGML_MAX_NAME=128)

ggerganov (commented):
Maybe projects using long tensor names should implement their own name shorteners to fit into `ggml` limit. Can be done with basic string  replacements:

```c
// set 
ggml_set_name(t, name_pack("some.very.long.tensor.name"));

// get
name_unpack(ggml_get_name(t));
```

Or name maps as in `llama.cpp`:

https://github.com/ggerganov/llama.cpp/blob/90fd43c7eb3d6d69326c26c02fea5cd0c6859a6e/llama.cpp#L369-L425

leejet:
I think the best approach might be to place the tensor name and some variable-length fields into dynamically allocated memory.

ggerganov:
> I think the best approach might be to place the tensor name and some variable-length fields into dynamically allocated memory.

Hm, not obvious how to do that.

Are my suggestions above not applicable in `sd.cpp`?

leejet:
> Are my suggestions above not applicable in sd.cpp?

Your suggestion is feasible, but I'm more inclined to increase the tensor name size. This way, I can maintain consistency between the tensor names in sd.cpp and the original stable diffusion code, aiding in code comprehension and debugging.

Or could the code be modified like this? This way, I can independently increase the tensor name size in the sd.cpp code without affecting other repositories using ggml."

```c
#ifndef GGML_MAX_NAME
#define GGML_MAX_NAME           64
#endif
```

ggerganov:
Yes - let's do the `-DGGML_MAX_NAME` thing. Could you update the PR?

leejet:
I've updated the PR.

ggerganov (approved):

WhyNotHugo:
When the system libggml is built with `-DGGML_MAX_NAME=128`, the headers still declare GGML_MAX_NAME=64, so there's a mismatch when other programs link to libggml (e.g.: stable-diffusion.cpp using `-DSD_USE_SYSTEM_GGML=ON`).

Cyberhan123:
@WhyNotHugo Because this is a compile-time macro substitution, this code will be embedded into the ggml library, therefore you need to recompile ggml.

WhyNotHugo:
Recompiling won't change anything; libggml is properly compiled with -DGGML_MAX_NAME=128, but its header libraries don't reflect this, and program which link to it have no way of ascertaining the value of `GGML_MAX_NAME`

Green-Sky:
> Recompiling won't change anything; libggml is properly compiled with -DGGML_MAX_NAME=128, but its header libraries don't reflect this, and program which link to it have no way of ascertaining the value of `GGML_MAX_NAME`

`-DGGML_MAX_NAME` is not taken into account by the cmake.

WhyNotHugo:
> -DGGML_MAX_NAME is not taken into account by the cmake.

It isn't? Then how do we build libggml with GGML_MAX_NAME=128?

PR #683, opened 2024-01-05
fix ROCm on Windows

Cyberhan123:
There is no need for the m link library in the Windows environment. If we try to link, an error will occur.
I can make sure it's correct because at [rwkv](https://github.com/saharNooby/rwkv.cpp/blob/d37568443493b540204bc93008d7820b1eefb74a/CMakeLists.txt#L422), that's what I do. We can also check in the rwkv release that the built library is correct.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -343,11 +343,7 @@ target_include_directories(${TARGET} PUBLIC
     ${GGML_EXTRA_INCS}
     )
 
-if (WIN32)
-    target_link_libraries(${TARGET} PUBLIC ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
-else()
-    target_link_libraries(${TARGET} PUBLIC m ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
-endif()
+target_link_libraries(${TARGET} PUBLIC ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
 
 if (BUILD_SHARED_LIBS)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -343,6 +343,11 @@ target_include_directories(${TARGET} PUBLIC
     ${GGML_EXTRA_INCS}
     )
 
+find_library(MATH_LIBRARY m)
+if (MATH_LIBRARY)
+    target_link_libraries(${TARGET} PUBLIC ${MATH_LIBRARY})
+endif()
+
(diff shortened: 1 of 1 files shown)


slaren:
Is there any reason to not add the library to the ggml-rocm target instead?

Cyberhan123:
I also searched on google and I found similar questions: https://github.com/conda-forge/fftw-feedstock/pull/75

slaren:
I have been checking how this is done in `llama.cpp`, but from what I can tell `llama.cpp` doesn't link to `m` at all, or at least not explicitly. So maybe this isn't necessary? @ggerganov do you know why ggml links to `m`?

Otherwise, if I understand the issue, under Windows there is no need to link to the `m` library, and that applies both to MSVC and clang. So maybe checking for `WIN32` instead of `MSVC` would also work:
```cmake
if (WIN32)
    target_link_libraries(${TARGET} PUBLIC ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
else()
    target_link_libraries(${TARGET} PUBLIC m ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
endif()
```

ggerganov:
I have forgotten why this `if` exists. Let's change it to never link `m` and merge it. If something breaks we will bring it back

Cyberhan123:
I also added some settings, which are the same as llama.cpp.

src/CMakeLists.txt
```diff
@@ -343,7 +343,7 @@ target_include_directories(${TARGET} PUBLIC
     ${GGML_EXTRA_INCS}
     )
 
-if (MSVC)
+if (WIN32)
     target_link_libraries(${TARGET} PUBLIC ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
 else()
     target_link_libraries(${TARGET} PUBLIC m ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
```
slaren, line 349: Let's remove the `m` lib as @ggerganov said.

```suggestion
target_link_libraries(${TARGET} PUBLIC ${GGML_EXTRA_LIBS} ${CMAKE_THREAD_LIBS_INIT})
```

Cyberhan123:
ok

Cyberhan123:
It looks like ci failed

slaren:
I have changed it to `find_library` as suggested in https://cliutils.gitlab.io/modern-cmake/chapters/features/small.html
Can you test if this still works with hip?

slaren:
I suspect the reason why it is not required to link with it in `llama.cpp` is because there are always C++ sources in the target.

Cyberhan123:
it is working

slaren (approved):

PR #685, opened 2024-01-07
added cuda float16->float32 upcasting to ggml_cuda_cpy

balisujohn:
Really basic addition. This seems to work in tortoise.cpp, but I haven't tested it after copying it here other thank to see that it compiles. It wasn't obvious to me if there are tests for cuda ops in ggml where I could add a test for this behavior. Curious to hear if I should add a test somewhere and If I should lint etc..

### DIFF
--- a/src/ggml-cuda.cu
+++ b/src/ggml-cuda.cu
@@ -4830,6 +4830,13 @@ static __device__ void cpy_1_f16_f16(const char * cxi, char * cdsti) {
     *dsti = *xi;
 }
 
+static __device__ void cpy_1_f16_f32(const char * cxi, char * cdsti) {
+    const half * xi = (const half *) cxi;
+    float * dsti = (float *) cdsti;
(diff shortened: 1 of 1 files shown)


slaren:
You can add a test to `test-backend-ops` by adding a test case. Something like this should work:
```C++
test_cases.emplace_back(new test_cpy(GGML_TYPE_F16, GGML_TYPE_F32, {256, 10, 10, 1}));
```
`ggml_backend_cuda_supports_op` also needs to be updated to return `true` for this case, otherwise the test will be skipped as "not supported"

It seems that #686 also contains these changes, should we focus on that instead?

balisujohn:
Yeah that works for me, since these are both improvements to `ggml_cuda_cpy`. I'll be away from my graphics card for a few days but once I'm back I'll look into adding a test for this and the 4d copy behavior.

PR #690, opened 2024-01-10
metal : wrap each operation in debug group

jmousseau:
The screenshot below shows a Metal debug capture of `gpt-2-backend` with the addition of debug groups.

ggml-metal-debug-group

### DIFF
--- a/src/ggml-metal.m
+++ b/src/ggml-metal.m
@@ -1067,6 +1067,8 @@ bool ggml_metal_graph_compute(
                     GGML_ASSERT(!"unsupported op");
                 }
 
+                [encoder pushDebugGroup:[NSString stringWithCString:ggml_op_desc(dst)]];
+
                 const int64_t  ne00 = src0 ? src0->ne[0] : 0;
                 const int64_t  ne01 = src0 ? src0->ne[1] : 0;
                 const int64_t  ne02 = src0 ? src0->ne[2] : 0;
@@ -2423,6 +2425,8 @@ bool ggml_metal_graph_compute(
                             GGML_ASSERT(false);
                         }
                 }
+
+                [encoder popDebugGroup];
             }
 
             if (encoder != nil) {


ggerganov (approved):
Hm this is very interesting and potentially useful! If you know more ways to improve the metal code for debugging and profiling purposes (e.g. using instruments) - please share.

P.S. This does not have effect on performance in Release builds - correct?

jmousseau:
>  If you know more ways to improve the metal code for debugging and profiling purposes (e.g. using instruments) - please share.

There are some next steps I'd like to explore:

- Add a `ggml_backend_metal_capture_next_compute` that would insert the necessary capture boundaries.  
- Consolidating constant kernel arguments into a single struct (https://github.com/ggerganov/llama.cpp/issues/3229)

> P.S. This does not have effect on performance in Release builds - correct?

As far as I know, performance should be unaffected in release builds. In my testing locally, I didn't see any changing in the timings. Would wrapping the capture and debug logic in `GGML_METAL_NDEBUG` be preferable?

ggerganov:
In `llama.cpp` we've decided to guard the debug calls with `GGML_METAL_NDEBUG`:

https://github.com/ggerganov/llama.cpp/commit/2a7c94db5fb67b2f8882d2d16a11bf5d8d12d397

Will sync the changes here soon

ggerganov:
@jmousseau How do you create these Metal debug captures that you've shown in the screenshot? I'm not very familiar with Xcode - been trying to figure it, but no luck so far. Would appreciate if you can share some step-by-step instructions

jmousseau:
@ggerganov Here are the steps I use, starting with Xcode project generation.

```bash
cmake -DGGML_METAL=ON -DBUILD_SHARED_LIBS=Off -G Xcode ..
```

```bash
open ggml.xcodeproj
```

 Select the `gpt-2-backend` scheme at the top, right of the git info.

xcode-select-scheme

Again click on the `gpt-2-backend` scheme, and choose `Edit Scheme...`. Configure the desired launch arguments and environment variables.

xcode-scheme-run-configuration


---

Traditionally, the easiest way to produce a Metal capture is with the "M" button above the debug console as shown below.

metal-capture-button

However, this will only capture GPU work enqueued *after* the button is pressed. For traditional graphics programs, this isn't a problem as the next rendered frame will be captured. In our case, the work will be queued (command buffers and encoders created) before you're able to initiate the capture. 

Therefore, setting up the capture boundary programmatically necessary. In `main-backend.cpp`, you'll want to initiate a capture before `ggml_backend_graph_compute` is called (requires https://github.com/ggerganov/ggml/pull/694).

```c
if (ggml_backend_is_metal(model.backend)) {
    ggml_backend_metal_capture_next_compute(model.backend);
}
```

Run the program by pressing the play button. Once the GPU work completes, Xcode will automatically open the [Metal debugger](https://developer.apple.com/documentation/xcode/metal-debugger).

PR #691, opened 2024-01-10
ggml: add ggml_is_4d()

ide-rea:


slaren:
This is not the way these functions are meant to be used, `ggml_is_4d` would just return true because every tensor is a 4d tensor.

ide-rea:
There has a  ` ggml_is_3d()` function, does you mean  3d tensor is a special 4d tensor?
Thanks!

PR #694, opened 2024-01-11
metal : add debug capture backend function

jmousseau:
`ggml_backend_metal_capture_next_compute` will capture all command buffers the next time `ggml_backend_graph_compute` is called, simplifying the Metal debugger workflow.

```c
// Example usage
if (ggml_backend_is_metal(model.backend)) {
    ggml_backend_metal_capture_next_compute(model.backend);
}
```

### DIFF
--- a/src/ggml-metal.h
+++ b/src/ggml-metal.h
@@ -109,6 +109,9 @@ GGML_API ggml_backend_buffer_type_t ggml_backend_metal_buffer_type(void);
 // ref: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
 GGML_API bool ggml_backend_metal_supports_family(ggml_backend_t backend, int family);
 
+// capture all command buffers committed the next time `ggml_backend_graph_compute` is called
+GGML_API void ggml_backend_metal_capture_next_compute(ggml_backend_t backend);
+
 #ifdef __cplusplus
 }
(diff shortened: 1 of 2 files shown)


src/ggml-metal.h
```diff
@@ -109,6 +109,9 @@ GGML_API ggml_backend_buffer_type_t ggml_backend_metal_buffer_type(void);
 // ref: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
 GGML_API bool ggml_backend_metal_supports_family(ggml_backend_t backend, int family);
 
+// capture all command buffers committed the next time `ggml_backend_graph_compute` is called
+GGML_API void ggml_backend_metal_capture_next_compute(ggml_backend_t backend);
```
ggerganov, line 113: @slaren Is it a good idea to extend the backend API with these type of calls or should we try to avoid it. In this case, it could remain a compile-time option (https://github.com/ggerganov/llama.cpp/commit/2a7c94db5fb67b2f8882d2d16a11bf5d8d12d397), but it might be useful to control at runtime in some cases

slaren: Generally I don't see an issue with adding more backend-specific functions, as long as they are not required for normal usage.

ggerganov (approved):

PR #696, opened 2024-01-12
sync : llama.cpp (ggml-backend)

ggerganov:
We should resolve the build so we can merge the latest changes from `llama.cpp` here


slaren:
I have fixed the issue with gpt-2, but the `starcoder-mmap.cpp` example uses the old ggml-cuda API, which is no longer supported.

ggerganov:
I think it's time to remove this example - will do so now

PR #709, opened 2024-01-27
gguf : add input validation, prevent integer overflows

ggerganov:
Input validation when reading GGUF files

Without these checks it is possible for crafted GGUF files to cause integer overflows, followed by heap overflows.

Thanks to Databricks team for responsibly reporting these

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -19363,9 +19363,9 @@ struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_p
 
         // sanity-checks to prevent from integer/buffer overflows
 
-        ok = ok && (ctx->header.n_tensors < SIZE_MAX/sizeof(struct gguf_tensor_info));
-        ok = ok && (ctx->header.n_tensors < SIZE_MAX/ggml_tensor_overhead());
-        ok = ok && (ctx->header.n_kv      < SIZE_MAX/sizeof(struct gguf_kv));
+        ok = ok && (ctx->header.n_tensors < (SIZE_MAX/2)/sizeof(struct gguf_tensor_info));
+        ok = ok && (ctx->header.n_tensors < (SIZE_MAX/2)/ggml_tensor_overhead());
+        ok = ok && (ctx->header.n_kv      < (SIZE_MAX/2)/sizeof(struct gguf_kv));
 
         if (!ok) {
             fprintf(stderr, "%s: failed to read header\n", __func__);
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -19348,10 +19348,9 @@ struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_p
         }
 
         // sanity-checks to prevent from integer/buffer overflows
-        // - sizeof(struct gguf_tensor_info)*n_tensors < SIZE_MAX
-        // - sizeof(struct gguf_kv)*n_kv               < SIZE_MAX
 
         ok = ok && (ctx->header.n_tensors < SIZE_MAX/sizeof(struct gguf_tensor_info));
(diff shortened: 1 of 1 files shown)


slaren:
Not directly related to this, but shouldn't these cases be `default:` instead to detect invalid types?

https://github.com/ggerganov/ggml/blob/866cc68b7b65da1f67f65720435240b86fc1f8c9/src/ggml.c#L19360-L19363

slaren:
I think this may also cause a buffer overflow if `n_dims > 4`:
https://github.com/ggerganov/ggml/blob/866cc68b7b65da1f67f65720435240b86fc1f8c9/src/ggml.c#L19391-L19394

slaren:
This `type` is not checked and then later is used to call ggml functions such as `ggml_type_name` that may cause an overflow:
https://github.com/ggerganov/ggml/blob/866cc68b7b65da1f67f65720435240b86fc1f8c9/src/ggml.c#L19395

src/ggml.c
```diff
@@ -19355,12 +19389,18 @@ struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_p
...
+            ok = ok && (info->n_dims <= GGML_MAX_DIMS);
+
             for (uint32_t j = 0; j < info->n_dims; ++j) {
                 ok = ok && gguf_fread_el(file, &info->ne[j], sizeof(info->ne[j]), &offset);
             }
+
             ok = ok && gguf_fread_el (file, &info->type,   sizeof(info->type),    &offset);
             ok = ok && gguf_fread_el (file, &info->offset, sizeof(info->offset),  &offset);
 
+            ok = ok && (info->type < GGML_TYPE_COUNT);
```
slaren, line 19402: There is no guarantee that the underlying type of the enum is unsigned, so I think it would be good to also check for `type >= 0`.

src/ggml.c
neilarchibald, line 1: Fixes look pretty good, I would also suggest checking the return of calloc/malloc for error, currently it will return a null pointer and crash.

neilarchibald: I just noticed you're using GGUF_TYPE_SIZE indexing using an arbitrary value read from the file. This could be used to index to a large value elsewhere and cause a wrap again, a check needs to be added when reading in the type value.

slaren:
https://github.com/ggerganov/ggml/blob/fb345bd49e01932314649d38e1e8cf7f5ed1cd6a/src/ggml.c#L19439-L19453

There are a few potential integer overflows here when computing the size, and this size may be later used to allocate a buffer, which then may result in a buffer overflow when reading the data.

slaren:
https://github.com/ggerganov/ggml/blob/fb345bd49e01932314649d38e1e8cf7f5ed1cd6a/src/ggml.c#L19400

We should probably check that the offset and the end of the tensor is within the size of the file.

slaren:
https://github.com/ggerganov/ggml/blob/fb345bd49e01932314649d38e1e8cf7f5ed1cd6a/src/ggml.c#L19399
We should check that the type is valid.

slaren:
https://github.com/ggerganov/ggml/blob/fb345bd49e01932314649d38e1e8cf7f5ed1cd6a/src/ggml.c#L19466-L19469
It seem unlikely to happen in practice, even with `no_alloc` a large enough `n_tensors` could cause an integer overflow. It would probably be good to wrap all the integer operations in a function that checks for overflows.

slaren:
https://github.com/ggerganov/ggml/blob/fb345bd49e01932314649d38e1e8cf7f5ed1cd6a/src/ggml.c#L19288
This is another potential integer overflow that may cause a the buffer to be too small.

ggerganov:
Think I covered all of the listed concerns. Will merge this and if you spot something else, directly open a PR. Thanks

PR #713, opened 2024-01-29
ggml : add simple example

FSSRepo:
This PR only aims to add a new example that simply performs a matrix multiplication, solely for the purpose of demonstrating a basic usage of ggml and backend handling. The code is commented to help understand what each part does.

$$
\begin{bmatrix}
2 & 8 \\
5 & 1 \\
4 & 2 \\
8 & 6 \\
\end{bmatrix}
\times
\begin{bmatrix}
10 & 9 & 5 \\
5 & 9 & 4 \\
\end{bmatrix}
\=
\begin{bmatrix}
60 & 110 & 54 & 29 \\
55 & 90 & 126 & 28 \\
50 & 54 & 42 & 64 \\
\end{bmatrix}
$$


examples/simple/simple.cpp
```diff
@@ -0,0 +1,208 @@
...
+    model.b = ggml_new_tensor_2d(model.ctx, GGML_TYPE_F32, 2, 2);
+
+    // create a backend buffer (backend memory) and alloc the tensors from the context
+    model.buffer = ggml_backend_alloc_ctx_tensors(model.ctx, model.backend);
+
+    // load data from cpu memory to backend buffer
+    if(ggml_backend_buffer_is_host(model.buffer)) {
+        memcpy(model.a->data, a, ggml_nbytes(model.a));
+        memcpy(model.b->data, b, ggml_nbytes(model.b));
+    } else {
```
slaren, line 84: This is only useful to avoid a copy when, for example, reading from a file. `ggml_backend_tensor_set` to a CPU buffer is just a `memcpy`. Even then, it is a minor optimization and shouldn't be in a simple example.

examples/simple/simple.cpp
```diff
@@ -0,0 +1,208 @@
...
+    simple_model model;
+    load_model(model, matrix_A, matrix_B);
+
+    // calculate the temporaly memory required to compute
+    ggml_backend_buffer_t buf_compute;
+    struct ggml_allocr * allocr = NULL;
+
+    {
+        size_t align = ggml_backend_get_alignment(model.backend);
+        allocr = ggml_allocr_new_measure(align);
```
slaren, line 172: It's better to use `ggml_allocr_new_measure_from_backend`, then the alignment and padding required by the backend is correctly taken into account.

examples/simple/simple.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    // release backend memory used for computation
+    ggml_backend_buffer_free(buf_compute);
+
+    // free memory
+    ggml_free(model.ctx);
+
+    // release backend memory and free backend
+    ggml_backend_buffer_free(model.buffer);
+    ggml_backend_free(model.backend);
```
slaren, line 206: The `allocr` should also be freed.

slaren: It's also possible to use `ggml_allocr_new_from_backend` instead of allocating a buffer separately. Then you only have the `allocr` object to free.

examples/simple/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+#
+# gpt-2
```
Green-Sky, line 2: :)

i have seen "gpt-2" so often as a copy paste leftover in different projects now, it basically a meme.

Green-Sky:
will this add to/replace the example in ggml.h itself?

https://github.com/ggerganov/ggml/blob/b2a5c34073e68b841568b2787ed5d15ae6a8d475/include/ggml/ggml.h#L32-L174

slaren:
This is a very useful example to have, good job!

slaren:
> will this add to/replace the example in ggml.h itself?

Not really, it is still possible to use ggml with that API, ggml-backend and ggml-alloc are completely optional. But it may be better to rename this example to `simple-backend` or similar to make that clear.

ggerganov:
Thanks for this @FSSRepo !

It would be useful to change the dimensions of the matrices in order to highlight along which dimension is the reduction. Most people are used to multiplying row by column and might get surprised by the numbers.

To disambiguate, you can use `(2x4) * (2x3) = (4x3)` for example

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    // perform computation
+    struct ggml_tensor * result = compute(model, allocr);
+
+    // create a array to print result
+    float* out_data = new float[ggml_nelements(result)];
+
+    // bring the data from the backend memory
+    ggml_backend_tensor_get(result, out_data, 0, ggml_nbytes(result));
```
slaren, line 190: This is never freed. Since this is C++, I would suggest changing it to a `std::vector` instead:
```suggestion
    std::vector<float> out_data(ggml_nelements(result));

    // bring the data from the backend memory
    ggml_backend_tensor_get(result, out_data.data(), 0, ggml_nbytes(result));
```

examples/simple/simple-ctx.cpp
```diff
@@ -0,0 +1,121 @@
...
+    int n_threads = 1; // number of threads to perform some operations with multi-threading
+
+    ggml_graph_compute_with_ctx(model.ctx, gf, n_threads);
+
+    // in this case, the output tensor is the last one in the graph
+    return gf->nodes[gf->n_nodes - 1];
+}
+
+int main(void)
+{
```
slaren, line 73:
```suggestion
int main(void) {
```

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+#endif
+
+    ggml_backend_graph_compute(model.backend, gf);
+
+    // in this case, the output tensor is the last one in the graph
+    return gf->nodes[gf->n_nodes - 1];
+}
+
+int main(void)
+{
```
slaren, line 141:
```suggestion
int main(void) {
```

FSSRepo:
This is ready

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    // the backend buffer to storage the tensors data of a and b
+    ggml_backend_buffer_t buffer;
+
+    // the context to define the tensor information (dimensions, size, memory address)
+    struct ggml_context * ctx;
+};
+
+// initialize the tensors of the model in this case two matrices 2x2
+void load_model(simple_model & model, float* a, float* b, int rows_A,int cols_A, int rows_B, int cols_B) {
```
ggerganov, line 38:
```suggestion
void load_model(simple_model & model, float * a, float * b, int rows_A, int cols_A, int rows_B, int cols_B) {
```

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+    fprintf(stderr, "%s: using Metal backend\n", __func__);
+    ggml_metal_log_set_callback(ggml_log_callback_default, nullptr);
+    model.backend = ggml_backend_metal_init();
+    if (!model.backend) {
+        fprintf(stderr, "%s: ggml_backend_metal_init() failed\n", __func__);
+    }
+#endif
+
+    // if there aren't GPU Backends fallback to CPU backend
+    if(!model.backend) {
```
ggerganov, line 58:
```suggestion
    if (!model.backend) {
```

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+    printf("mult mat (%d x %d) (transposed result):\n[", result->ne[0], result->ne[1]);
+    for(int j = 0; j < result->ne[1] /* rows */; j++) {
+        if(j > 0) {
+            printf("\n");
+        }
+        for(int i = 0; i < result->ne[0] /* cols */; i++) {
+            printf(" %.2f", out_data[i * result->ne[1] + j]);
+        }
+    }
+    printf(" ]");
```
ggerganov, line 203:
```suggestion
    printf("mul mat (%d x %d) (transposed result):\n[", result->ne[0], result->ne[1]);
    for (int j = 0; j < result->ne[1] /* rows */; j++) {
        if (j > 0) {
            printf("\n");
        }
        for (int i = 0; i < result->ne[0] /* cols */; i++) {
            printf(" %.2f", out_data[i * result->ne[1] + j]);
        }
    }
    printf(" ]\n");
```

examples/simple/simple-ctx.cpp
```diff
@@ -0,0 +1,125 @@
...
+struct simple_model {
+    struct ggml_tensor * a;
+    struct ggml_tensor * b;
+
+    // the context to define the tensor information (dimensions, size, memory data)
+    struct ggml_context * ctx;
+};
+
+// initialize the tensors of the model in this case two matrices 2x2
+void load_model(simple_model & model, float* a, float* b, int rows_A,int cols_A, int rows_B, int cols_B) {
```
ggerganov, line 22:
```suggestion
void load_model(simple_model & model, float * a, float * b, int rows_A, int cols_A, int rows_B, int cols_B) {
```

examples/simple/simple-ctx.cpp
```diff
@@ -0,0 +1,125 @@
...
+    printf("mult mat (%d x %d) (transposed result):\n[", result->ne[0], result->ne[1]);
+    for(int j = 0; j < result->ne[1] /* rows */; j++) {
+        if(j > 0) {
+            printf("\n");
+        }
+        for(int i = 0; i < result->ne[0] /* cols */; i++) {
+            printf(" %.2f", out_data[i * result->ne[1] + j]);
+        }
+    }
+    printf(" ]");
```
ggerganov, line 120:
```suggestion
    printf("mul mat (%d x %d) (transposed result):\n[", result->ne[0], result->ne[1]);
    for (int j = 0; j < result->ne[1] /* rows */; j++) {
        if (j > 0) {
            printf("\n");
        }
        for (int i = 0; i < result->ne[0] /* cols */; i++) {
            printf(" %.2f", out_data[i * result->ne[1] + j]);
        }
    }
    printf(" ]\n");
```

ggerganov (approved):

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+    // build operations nodes
+    ggml_build_forward_expand(gf, result);
+
+    // delete the temporally context used to build the graph
+    ggml_free(ctx0);
+    return gf;
+}
+
+// compute with backend
+struct ggml_tensor* compute(const simple_model & model, ggml_gallocr_t allocr) {
```
slaren, line 113:
```suggestion
struct ggml_tensor * compute(const simple_model & model, ggml_gallocr_t allocr) {
```

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+    simple_model model;
+    load_model(model, matrix_A, matrix_B, rows_A, cols_A, rows_B, cols_B);
+
+    // calculate the temporaly memory required to compute
+    ggml_gallocr_t allocr = NULL;
+
+    {
+        allocr = ggml_gallocr_new(ggml_backend_get_default_buffer_type(model.backend));
+
+        //create the worst case graph for memory usage estimation
```
slaren, line 172:
```suggestion
        // create the worst case graph for memory usage estimation
```

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,215 @@
...
+        /*.no_alloc   =*/ true, // the tensors will be allocated later by ggml_allocr_alloc_graph()
+    };
+
+    // create a temporally context to build the graph
+    struct ggml_context * ctx0 = ggml_init(params0);
+
+    struct ggml_cgraph  * gf = ggml_new_graph(ctx0);
+
+    // result = a*b^T
+    struct ggml_tensor* result = ggml_mul_mat(ctx0, model.a, model.b);
```
slaren, line 102:
```suggestion
    struct ggml_tensor * result = ggml_mul_mat(ctx0, model.a, model.b);
```

slaren (approved):

examples/simple/simple-backend.cpp
```diff
@@ -0,0 +1,216 @@
...
+    struct ggml_cgraph  * gf = ggml_new_graph(ctx0);
+
+    // result = a*b^T
+    struct ggml_tensor * result = ggml_mul_mat(ctx0, model.a, model.b);
+
+    // build operations nodes
+    ggml_build_forward_expand(gf, result);
+
+    // delete the temporally context used to build the graph
+    ggml_free(ctx0);
```
zcbenz, line 108: Since `gf` is allocated in `ctx0`'s buffer, wouldn't using `gf` after freeing `ctx0` be a use-after-free?

slaren: The backing memory for the context is stored in the static `buf` vector. It's not great but it is not UB.

PR #728, opened 2024-02-10
examples : remove old stuff

ggerganov:
Any reason to keep these around?


slaren:
I think it is good to remove these. There is still a mpt test case in the ggml-ci that should also be removed.

Issue #724, opened 2024-02-05
ggml : simplify the ggml_compute_forward_ calls


ggerganov:
There is no need to explicitly list all the source tensors - they can be accessed through `dst` when needed:

https://github.com/ggerganov/ggml/blob/6b14d738d9100c50c199a3b1aaa960f633904476/src/ggml.c#L14783-L14812

ggerganov:
Thanks!

PR #729, opened 2024-02-11
refactored compute forward to not pass in the src tensors each time

siddharthvader:
intended to address issue #724


src/ggml.c
```diff
@@ -2607,7 +2607,7 @@ static struct ggml_tensor * ggml_new_tensor_impl(
         /*.nb           =*/ { 0, 0, 0, 0 },
         /*.op           =*/ GGML_OP_NONE,
         /*.op_params    =*/ { 0 },
-        /*.flags        =*/ 0,
+        /*.is_param     =*/ false,
```
slaren, line 2610: It looks like something went wrong while merging with the latest changes, this and all other uses of `flags` shouldn't be changed.

siddharthvader: hmm got it, will take a look

siddharthvader: okay I should have fixed all the instances where there were merge issues affecting `flags`. the code builds now after running `make -j4 gpt-2-backend gpt-j` so I think it should be in good shape now!

siddharthvader:
looking into why the customop test is failing

slaren (approved):

PR #730, opened 2024-02-12
Add Vulkan to cmake build

abetlen:
- Adds a `GGML_VULKAN` option to cmake to allow building those sources into ggml
- Adds the `ggml_vk_generate_shaders.py` and `ggml-vulkan-shaders.hpp` from `llama.cpp` to `src/`


ggerganov:
We should wait for https://github.com/ggerganov/llama.cpp/issues/5356 to see if they come up with a way to re-organize the Vulkan shaders

abetlen:
Makes sense, no rush for me I was just experimenting with the new backends.

foldl:
https://github.com/ggerganov/llama.cpp/issues/5356 is closed for being inactive.

So, any updates?

abetlen:
@ggerganov maybe for now I can just remove the auto-generated `.hpp` file and update the sync scripts. That should at least allow users to build with Vulkan by first running the python script to generate the shaders. How does that sound?

ggerganov:
@abetlen Ok, let's do that

abetlen:
@ggerganov okay, I've removed the auto-generated file and updated the sync scripts. I also updated the `ggml_vk_generate_shaders.py` file to match the current `ggml-vulkan` files.

ggerganov (approved):

PR #731, opened 2024-02-12
ggml-alloc : allocate all leafs as if they were inputs

slaren:
Effectively, this means that it is no longer necessary to call `ggml_set_input` to flag inputs. All tensors created with `ggml_new_tensor` that end in a graph (either as a source of an operation, or as a leaf by calling `ggml_build_forward_expand`) are allocated at the beginning of the graph as if they were inputs.

### DIFF
--- a/examples/gpt-2/main-alloc.cpp
+++ b/examples/gpt-2/main-alloc.cpp
@@ -407,11 +407,11 @@ struct ggml_cgraph * gpt2_graph(
         /*.no_alloc   =*/ true, // the tensors will be allocated later by ggml_gallocr_alloc_graph()
     };
 
-    struct ggml_context * ctx0 = ggml_init(params);
+    struct ggml_context * ctx = ggml_init(params);
 
-    struct ggml_cgraph  * gf = ggml_new_graph(ctx0);
+    struct ggml_cgraph  * gf = ggml_new_graph(ctx);
 
-    struct ggml_tensor * embd = ggml_new_tensor_1d(ctx0, GGML_TYPE_I32, N);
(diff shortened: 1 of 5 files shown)


ggerganov:
I think `ggml_gallocr_alloc_graph` should also now handle the leafs, not only the nodes?

https://github.com/ggerganov/ggml/blob/40adc8694c16cee780e7a484a3f3c760bf976bfe/src/ggml-alloc.c#L838-L851

slaren:
Ah yes, I missed that. I will fix it.

slaren:
Should be fixed now. I noticed that this change increased the compute buffer size of the gpt-2 example by a factor of 10 (7 MB to 70 MB), due to the `ggml_new_tensor` in a few `ggml_cpy`. Replacing these with `ggml_cont_nd` restores the memory usage back to normal, but this shows the risk of this change. A `ggml_new_tensor` in each layer will cause all of these tensors to be allocated at the start of the graph, and in some cases that can be several times more memory than the whole compute buffer needs.

ggerganov:
Hm, I see. In the `whisper.cpp` PR, I added a workaround that does not look too bad and works with the existing allocator. Wondering if we should proceed with this change

slaren:
I think in the long term this change will cause the least issues to the users and will make ggml easier to use, but as always with (somewhat) breaking changes, there will be issues during the transition.

slaren:
We could add a check in `ggml_cpy`, and if the `dst` tensor is a `GGML_OP_NONE` that is not allocated, then print a warning in debug builds, suggesting changing that to `ggml_cont` or `ggml_cast`.

ggerganov (approved):
Ok, let's merge it - I think it's a good change

PR #743, opened 2024-02-22
Introduce backend GUIDs

uextm:
Initial proposed implementation of backend GUIDs
(Discussed in https://github.com/ggerganov/ggml/pull/741)

Hardcoded CPU backend GUID (for now)
Change ggml_backend_is_cpu logic to use GUID


src/ggml-backend-impl.h
```diff
@@ -76,6 +76,8 @@ extern "C" {
     typedef void * ggml_backend_context_t;
 
     struct ggml_backend_i {
+        ggml_guid_t (*GGML_CALL get_guid)();
+
```
slaren, line 80: Since the UUID is already stored in the `struct ggml_backend`, I don't think this needs to be part of the interface. Either on the interface or on the struct is fine, but putting it on both is redundant.

I cannot see any reason to change the UUID at run time, so I think that it would be better to just keep it as a constant on the struct, to be set by the backends during initialization in the `ggml_backend_xxx_init` functions.

uextm: Alright I removed the interface function and its helper

include/ggml/ggml-backend.h
```diff
@@ -83,6 +82,8 @@ extern "C" {
 
     GGML_API ggml_backend_t ggml_backend_cpu_init(void);
 
+    GGML_API GGML_CALL ggml_guid_t ggml_backend_cpu_guid();
```
slaren, line 85:
```suggestion
    GGML_API GGML_CALL ggml_guid_t ggml_backend_cpu_guid(void);
```

include/ggml/ggml.h
```diff
@@ -672,6 +672,16 @@ extern "C" {
         GGML_NUMA_STRATEGY_COUNT
     };
 
+    //
+    // GUID
+    //
+
+    // GUID types
+    typedef uint8_t ggml_guid [16];
+    typedef ggml_guid* ggml_guid_t;
```
slaren, line 681:
```suggestion
    typedef uint8_t ggml_guid[16];
    typedef ggml_guid * ggml_guid_t;
```

src/ggml-backend.c
```diff
@@ -12,6 +12,7 @@
 
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 
+static ggml_guid CPU_GUID = {0x7c, 0x7d, 0x73, 0x6e, 0xc7, 0xaf, 0x45, 0xd3, 0xad, 0x31, 0x55, 0xd5, 0x57, 0x49, 0xa0, 0x3a };
```
slaren, line 15: This can be moved as a static local to `ggml_backend_cpu_guid()`.

src/ggml.c
```diff
@@ -355,6 +355,11 @@ void ggml_fp32_to_fp16_row(const float * x, ggml_fp16_t * y, int n) {
     }
 }
 
+bool ggml_guid_matches(ggml_guid_t guid_a, ggml_guid_t guid_b)
+{
+    return memcmp(guid_a, guid_b, sizeof (ggml_guid)) == 0;
+}
```
slaren, line 361:
```suggestion
bool ggml_guid_matches(ggml_guid_t guid_a, ggml_guid_t guid_b) {
    return memcmp(guid_a, guid_b, sizeof (ggml_guid)) == 0;
}
```

slaren (commented):
It would be good to keep the helper function `ggml_backend_guid`, just returning the value from the struct instead.

uextm:
> It would be good to keep the helper function `ggml_backend_guid`, just returning the value from the struct instead.

OK adding it back, it just seemed like this was redundant in the same ways the other things you wanted removed were.

I'm just trying to follow the style of the project and your comments; personally I would have done things very differently so I'm going off your guidance.

ggerganov (approved):

slaren (approved):

PR #747, opened 2024-02-25
add some new ops, fix some operators and add batch operations to certain operators.

leejet:
I tested it in the [sd.cpp](https://github.com/leejet/stable-diffusion.cpp) project, and everything works well.


FSSRepo:
@leejet I believe that before merging this, the respective Metal kernels should be created to maintain compatibility, although it doesn't really help much since sd.cpp doesn't benefit from the Metal backend as matrix operations are very slow.

leejet:
OK,  I will find time to add support for Metal backend.

leejet:
@FSSRepo I have implemented Metal backend support for ggml_arange and ggml_timestep_embedding, and tested it with the test-arrange and test-timestep-embedding test cases to ensure it works correctly. Could you find some time to review and merge this PR?

FSSRepo:
Ok, I will test this PR with latest sd.cpp

slaren:
- Does `arange` affect the performance significantly? Are there any downsides to constructing the tensor in a CPU buffer and copying it to the backend as an input?
- Does `timestep_embedding` have any use outside of SD? And again, would it hurt performance significantly if it was implemented on the CPU and copied to the backend as an input?

leejet:
These two operations won't significantly improve performance, but in stable video diffusion, it's necessary to perform these operations multiple times within the UNet network. Additionally, the parameters may change with each execution. If I don't execute these operations when computing the graph, I would have to remember the tensors produced by each operation and their corresponding parameters. Then, during the ggml_gallocr_alloc_graph, I'd have to copy the data, making the process more complex. Implementing this through ggml operators would make the code more elegant.

src/ggml-cuda.cu
```diff
@@ -1042,8 +1051,42 @@ static __global__ void pad_f32(const float * x, float * dst, const int ne0, cons
...
+    int half = dim / 2;
+    if (j >= half) {
+        return;
+    }
+
+    float timestep = timesteps[i];
+    float freq = (float)exp(-logf(max_period) * j / half);
+    float arg = timestep * freq;
+    embed_data[j] = cos(arg);
+    embed_data[j + half] = sin(arg);
```
slaren, line 1083: This is using double precision functions, which can be very slow on some devices. Using `expf`, `cosf` and `sinf` instead should improve performance.

src/ggml-cuda.cu
```diff
@@ -10458,6 +10553,45 @@ static void ggml_cuda_pad(const ggml_tensor * src0, const ggml_tensor * src1, gg
...
+    cuda_pool_alloc<float>  dst_f;
+
+    ggml_cuda_set_device(g_main_device);
+    cudaStream_t main_stream = g_cudaStreams[g_main_device][0];
+
+    if (dst_on_device) {
+        dst_ddf = (float *) dst_extra->data_device[g_main_device];
+    } else {
+        dst_ddf = dst_f.alloc(ggml_nelements(dst));
+    }
```
slaren, line 10575: This is not necessary, with ggml-backend all the tensors are always allocated on the device.

src/ggml.c
```diff
@@ -5898,6 +5902,55 @@ struct ggml_tensor * ggml_upscale(
...
+    struct ggml_tensor * result = ggml_new_tensor_1d(ctx, GGML_TYPE_F32, steps);
+
+    result->op = GGML_OP_ARANGE;
+    ((float *)(result->op_params))[0] = start;
+    ((float *)(result->op_params))[1] = stop;
+    ((float *)(result->op_params))[2] = step;
+
+    return result;
+}
+
```
slaren, line 5924:
```suggestion
```

src/ggml.c
```diff
@@ -5898,6 +5902,55 @@ struct ggml_tensor * ggml_upscale(
...
+            int                   dim,
+            int                   max_period) {
+    bool is_node = false;
+
+    if (timesteps->grad) {
+        GGML_ASSERT(false); // TODO: implement backward
+        is_node = true;
+    }
+
+    int acutual_dim = dim;
```
slaren, line 5938:
```suggestion
    int actual_dim = dim;
```

src/ggml.c
```diff
@@ -5898,6 +5902,55 @@ struct ggml_tensor * ggml_upscale(
...
+    if (dim % 2 != 0) {
+        acutual_dim = dim + 1;
+    }
+
+    struct ggml_tensor * result = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, acutual_dim, timesteps->ne[0]);
+
+    result->op = GGML_OP_TIMESTEP_EMBEDDING;
+    result->op_params[0] = dim;
+    result->op_params[1] = max_period;
```
slaren, line 5947: Use `ggml_set_op_params` or `ggml_set_op_params_i32`.

src/ggml.c
```diff
@@ -5898,6 +5902,55 @@ struct ggml_tensor * ggml_upscale(
...
+    GGML_ASSERT(stop > start);
+
+    int64_t steps = (int64_t)ceil((stop - start) / step);
+
+    struct ggml_tensor * result = ggml_new_tensor_1d(ctx, GGML_TYPE_F32, steps);
+
+    result->op = GGML_OP_ARANGE;
+    ((float *)(result->op_params))[0] = start;
+    ((float *)(result->op_params))[1] = stop;
+    ((float *)(result->op_params))[2] = step;
```
slaren, line 5920: Use `ggml_set_op_params`.

src/ggml.c
```diff
@@ -13547,6 +13600,109 @@ static void ggml_compute_forward_pad(
...
+    GGML_ASSERT(dst->nb[0] == sizeof(float));
+
+    const int ith = params->ith;
+    const int nth = params->nth;
+
+    GGML_TENSOR_UNARY_OP_LOCALS
+
+    const float start = ((float*)dst->op_params)[0];
+    const float stop = ((float*)dst->op_params)[1];
+    const float step = ((float*)dst->op_params)[2];
```
slaren, line 13624: Use `memcpy` instead.

src/ggml.c
```diff
@@ -13547,6 +13600,109 @@ static void ggml_compute_forward_pad(
...
+
+    GGML_ASSERT(src0->nb[0] == sizeof(float));
+
+    const int ith = params->ith;
+    const int nth = params->nth;
+
+    GGML_TENSOR_UNARY_OP_LOCALS
+
+    const int dim = dst->op_params[0];
+    const int max_period = dst->op_params[1];
```
slaren, line 13668: Use `ggml_get_op_params_i32`.

src/ggml.c
```diff
@@ -13547,6 +13600,109 @@ static void ggml_compute_forward_pad(
...
+    int half = dim / 2;
+
+    for (int64_t i = 0; i < ne00; i++) {
+        float * embed_data = (float *)((char *)  dst->data +  i*nb1);
+        for (int64_t j = ith; j < half; j += nth) {
+            float timestep = ((float *)src0->data)[i];
+            float freq = (float)exp(-log(max_period) * j / half);
+            float arg = timestep * freq;
+            embed_data[j] = cos(arg);
+            embed_data[j + half] = sin(arg);
```
slaren, line 13682: Also might want to use single precision math functions here for performance.

tests/test-timestep_embedding.cpp
slaren, line 1: It seems that this test can never fail.

slaren (commented):
It would be good to have tests for these operators in `test-backend-ops`.

include/ggml/ggml.h
```diff
@@ -1672,6 +1683,12 @@ extern "C" {
             struct ggml_tensor  * a,
             enum ggml_sort_order  order);
 
+    GGML_API struct ggml_tensor * ggml_arange(
+            struct ggml_context * ctx,
+            float                  start,
+            float                  stop,
+            float                  step);
+
```
ggerganov, line 1691:
```suggestion
            struct ggml_context * ctx,
            float                 start,
            float                 stop,
            float                 step);

```

src/ggml-cuda.cu
```diff
@@ -10458,6 +10553,45 @@ static void ggml_cuda_pad(const ggml_tensor * src0, const ggml_tensor * src1, gg
     ggml_cuda_op_flatten(src0, src1, dst, ggml_cuda_op_pad);
 }
 
+static void ggml_cuda_arange(const ggml_tensor * src0, const ggml_tensor * src1, ggml_tensor * dst) {
+    ggml_tensor_extra_gpu * dst_extra  =            (ggml_tensor_extra_gpu *)  dst->extra;
+
+    const bool  dst_on_device =              dst->backend == GGML_BACKEND_TYPE_GPU;
```
ggerganov, line 10559:
```suggestion
static void ggml_cuda_arange(const ggml_tensor * src0, const ggml_tensor * src1, ggml_tensor * dst) {
    ggml_tensor_extra_gpu * dst_extra = (ggml_tensor_extra_gpu *)  dst->extra;

    const bool dst_on_device = dst->backend == GGML_BACKEND_TYPE_GPU;
```

src/ggml-metal.m
```diff
@@ -2300,6 +2306,46 @@ static bool ggml_metal_graph_compute(
...
+                        const float start = ((float*)dst->op_params)[0];
+                        const float step = ((float*)dst->op_params)[2];
+
+                        id<MTLComputePipelineState> pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_ARANGE_F32].pipeline;
+
+                        [encoder setComputePipelineState:pipeline];
+                        [encoder setBuffer:id_dst  offset:offs_dst  atIndex:0];
+                        [encoder setBytes:&ne0 length:sizeof(ne0) atIndex:1];
+                        [encoder setBytes:&start length:sizeof(start) atIndex:2];
+                        [encoder setBytes:&step length:sizeof(step) atIndex:3];
```
ggerganov, line 2322:
```suggestion
                        [encoder setComputePipelineState:pipeline];
                        [encoder setBuffer:id_dst  offset:offs_dst    atIndex:0];
                        [encoder setBytes:&ne0   length:sizeof(ne0)   atIndex:1];
                        [encoder setBytes:&start length:sizeof(start) atIndex:2];
                        [encoder setBytes:&step  length:sizeof(step)  atIndex:3];
```

src/ggml-metal.m
```diff
@@ -2300,6 +2306,46 @@ static bool ggml_metal_graph_compute(
...
+                        int half = dim / 2;
+
+                        id<MTLComputePipelineState> pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_TIMESTEP_EMBEDDING_F32].pipeline;
+
+                        [encoder setComputePipelineState:pipeline];
+                        [encoder setBuffer:id_src0 offset:offs_src0 atIndex:0];
+                        [encoder setBuffer:id_dst  offset:offs_dst  atIndex:1];
+                        [encoder setBytes:&nb1 length:sizeof(nb1) atIndex:2];
+                        [encoder setBytes:&dim length:sizeof(dim) atIndex:3];
+                        [encoder setBytes:&max_period length:sizeof(max_period) atIndex:4];
```
ggerganov, line 2343:
```suggestion
                        [encoder setComputePipelineState:pipeline];
                        [encoder setBuffer:id_src0 offset:offs_src0 atIndex:0];
                        [encoder setBuffer:id_dst  offset:offs_dst  atIndex:1];
                        [encoder setBytes:&nb1   length:sizeof(nb1) atIndex:2];
                        [encoder setBytes:&dim   length:sizeof(dim) atIndex:3];
                        [encoder setBytes:&max_period length:sizeof(max_period) atIndex:4];
```

src/ggml-metal.metal
```diff
@@ -1959,6 +1959,49 @@ kernel void kernel_pad_f32(
...
+kernel void kernel_arange_f32(
+    device        char * dst,
+    constant   int64_t & ne0,
+    constant   float   & start,
+    constant   float   & step,
+    uint3 tgpig[[threadgroup_position_in_grid]],
+    uint3 tpitg[[thread_position_in_threadgroup]],
+    uint3   ntg[[threads_per_threadgroup]]) {
+
+    device       float * dst_ptr  = (device       float *) dst;
```
ggerganov, line 1971:
```suggestion
    device float * dst_ptr = (device float *) dst;
```

src/ggml.c
```diff
@@ -10262,11 +10315,11 @@ static void ggml_compute_forward_group_norm_f32(
                     for (int64_t i00 = 0; i00 < ne00; i00++) {
                         float v = x[i00] - mean;
                         y[i00] = v;
-                        sum2 += (ggml_float)(v * v);
+                        sum2 += (ggml_float)(v * v) / (ne00 * ne01 * step);
                     }
                 }
             }
-            float variance = sum2 / (ne00 * ne01 * step);
+            float variance = sum2;
```
ggerganov, line 10322: What was the reason to change these normalization to be inside the loop?

ggerganov (approved):
After adding tests to `test-backend-ops` we can merge

src/ggml.c
leejet: Because previously, it calculated the sum of (v * v) and then divided by (ne00 * ne01 * step), I was concerned that in certain situations, it might lead to overflow or precision issues. Therefore, I changed it to calculate the sum of (v * v)/(ne00 * ne01 * step) to obtain the variance.

src/ggml-cuda.cu
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

src/ggml.c
leejet: Changed in the latest commit.

tests/test-timestep_embedding.cpp
leejet: Changed in the latest commit.

src/ggml-cuda.cu
leejet: This part was copied from ggml_cuda_op_flatten, and I'm not sure whether to make changes in this PR or create a separate PR to modify both places?

leejet:
I have modified the code based on the suggestions from the review and added tests to test-backend-ops. There are two uncertain points still awaiting feedback from the maintainer.

src/ggml.c
ggerganov: It shouldn't cause problems since we accumulate into `double` (i.e. `ggml_float`). I've introduced a separate accumulator per row and restored the normalization to be at the end in order to avoid extra division in the loop

ggerganov (approved):

slaren (approved):

PR #748, opened 2024-02-25
add google magika inference example

slaren:
The model can be converted with `convert.py` from [model.h5](https://github.com/google/magika/blob/4460acb5d3f86807c3b53223229dee2afa50c025/assets_generation/models/standard_v1/model.h5)

Example:
```
$ build/bin/magika model.h5.gguf examples/sam/example.jpg README.md src/ggml.c
examples/sam/example.jpg      : jpeg (100.00%) pptx (0.00%) smali (0.00%) shell (0.00%) sevenzip (0.00%)
README.md                     : markdown (100.00%) txt (0.00%) yaml (0.00%) ppt (0.00%) shell (0.00%)
src/ggml.c                    : c (99.97%) asm (0.01%) txt (0.01%) javascript (0.00%) html (0.00%)
```


ggerganov:
> Requires F32 GELU, as the default F16 GELU will result in nan. Not sure how to address this.

Would this patch make it work with the F16 GELU define?

```diff
diff --git a/src/ggml.c b/src/ggml.c
index d710fe7..42a7d45 100644
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -1560,9 +1560,15 @@ inline static void ggml_vec_gelu_f16(const int n, ggml_fp16_t * y, const ggml_fp
 inline static void ggml_vec_gelu_f32(const int n, float * y, const float * x) {
     uint16_t t;
     for (int i = 0; i < n; ++i) {
-        ggml_fp16_t fp16 = GGML_FP32_TO_FP16(x[i]);
-        memcpy(&t, &fp16, sizeof(uint16_t));
-        y[i] = GGML_FP16_TO_FP32(ggml_table_gelu_f16[t]);
+        if (x[i] < -10.0f) {
+            y[i] = 0.0f;
+        } else if (x[i] > 10.0f) {
+            y[i] = x[i];
+        } else {
+            ggml_fp16_t fp16 = GGML_FP32_TO_FP16(x[i]);
+            memcpy(&t, &fp16, sizeof(uint16_t));
+            y[i] = GGML_FP16_TO_FP32(ggml_table_gelu_f16[t]);
+        }
     }
 }
 #else
```

slaren:
Yep, it does work. The problem was that activatios generated by this model can exceed the FP16 maximum finite value of 65519, which results in `inf` when converted to FP16.

examples/magika/main.cpp
```diff
@@ -0,0 +1,398 @@
...
+                        float val = *(const float *)(src_data + i0*t2->nb[0] + i1*t2->nb[1] + i2*t2->nb[2] + i3*t2->nb[3]);
+                        if (val > max) {
+                            max = val;
+                        }
+                    }
+                    *(float *)(dst_data + i1*dst->nb[0] + i2*dst->nb[1] + i3*dst->nb[2]) = max;
+                }
+            }
+        }
+    }, 1, NULL);
```
ggerganov, line 256: Does `ggml_pool_1d` not work here?

slaren: It required a minor change to support 3d tensors, but it seems to work.

ggerganov (approved):
Very cool!

Does it work with CUDA?
I just tried Metal, but it currently lacks the POOL_1D op.

slaren:
CUDA does not support POOL_1D either, so I guess it is CPU only for now.

PR #750, opened 2024-02-27
ggml_status introduction

Xarbirus:
I would like to add the ability to explicitly specify the execution result for the `ggml_backend_graph_plan_compute` and `ggml_backend_graph_compute` functions.

This will allow us to determine whether the execution was aborted by abort_callback or by some other reasons (and allow us to use different reasons). I used the additional `ggml_compute_result_t` type because when executing `ggml_*_compute` multi-threaded, each thread may return its own error/warning, and in order not to lose them, I decided to concatenate exit_code's using OR.

Also I didn't remove `GGML_EXIT_SUCCESS` and `GGML_EXIT_ABORTED` because they can be used by other party libraries that use ggml.


slaren:
I think this is a good change. Some comments:
- `ggml_compute_exit_code` does not need to be a bit field, in practice only one error can happen, and even if in the future there is some bizarre condition that may cause different errors in different threads, I think it is still better to report one error only to keep the API simpler
- I would prefer a more generic name that can be reused in other ggml APIs in the future, such as `ggml_error` or `ggml_status`

ggerganov:
+1 for `ggml_status`

Xarbirus:
Thank you for your review, made changes:

- renamed new type from `ggml_compute_exit_code` to `ggml_status`
- added `ggml_status_to_string` (to cast a status to a string)
- updated statuses, now negative values ​​mean errors and positive values ​​mean warnings

And I get an important question. The `llama_decode` function has its own error codes, for example -1 is used [here](https://github.com/ggerganov/llama.cpp/blob/8c0e8f4e73e275756ad69f9c99b26ead085ca9f0/llama.cpp#L7887) and 1 is used [here](https://github.com/ggerganov/llama.cpp/blob/8c0e8f4e73e275756ad69f9c99b26ead085ca9f0/llama.cpp#L7957). Do I need to add all possible errors from llama.cpp to `ggml_status` or do I need to create a separate enum in llama.cpp that will expand and complement `ggml_status` later (after syncing ggml with llama.cpp)?

slaren:
My opinion is that llama.cpp needs its own `llama_status` with its own error codes, since llama.cpp has its own error conditions that are completely independent and separate of ggml.

Xarbirus:
@slaren , I completely agree with you!

src/ggml.c
```diff
@@ -320,6 +320,16 @@ static ggml_fp16_t ggml_table_exp_f16[1 << 16];
 // precomputed f32 table for f16 (256 KB) (ggml-impl.h)
 float ggml_table_f32_f16[1 << 16];
 
+GGML_CALL const char * ggml_status_to_string(enum ggml_status status) {
```
slaren, line 323: I have a feeling that these `GGML_CALL` are creeping everywhere because nobody knows exactly where they are needed or not.
```suggestion
const char * ggml_status_to_string(enum ggml_status status) {
```

ggerganov (approved):

PR #764, opened 2024-03-14
sync : llama.cpp

ggerganov:
Need to fix the build after the recent changes


slaren:
There is still some issue with the gpt-2 sched example, I am looking into it.

slaren (approved):

PR #765, opened 2024-03-15
spec : add GGUF diagram

ggerganov:
Adding the following GGUF format visualization by @mishig25 to the spec

image


mishig25 (approved):
LGTM !

PR #770, opened 2024-03-21
Fix examples/simple/simple-ctx

wilderfield:
There was an error in how the result tensor was being accessed.

This error led to the example creator defining incorrect expected result.

Please double check my assumption that the output of ggml_mul_mat is the transposed result of A*B.


ggerganov (approved):
Apologies for the confusion!

wilderfield:
> Apologies for the confusion!

No worries, it was good linear algebra review for me.
Thank you for cleaning up the README!

PR #777, opened 2024-03-31
ggml: add type_size to data view offset calc

danbev:
This commit suggests adding the type_size to the calculation of the offset when creating a view of a tensor.

The motivation for this change is that when I tried calling `ggml_view_1d` using an offset like this:
```c
    struct ggml_tensor * view = ggml_view_1d(ctx, x, 5, 0);
```
Where the `x` tensor has 10 elments with values ranging from 0 to 9, and requesting a view with 5 elements, starting at offset zero this works as expected.

But if I try using an offset of 5, with the expectation (which may be incorrect) that this would produce a view with the last 5 elements of the tensor `x`:
```c
    struct ggml_tensor * view = ggml_view_1d(ctx, x, 5, 5);
```
This does not work as I had expected. This commit contains a test with the above use case, which hopefully clarifies that I'm trying to do and perhaps someone can point out what I'm doing wrong, or confirm that this may be an issue with the current implementation.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -2942,7 +2942,8 @@ static struct ggml_tensor * ggml_new_tensor_impl(
 
     void * data = view_src != NULL ? view_src->data : NULL;
     if (data != NULL) {
-        data = (char *) data + view_offs;
+        size_t type_size = ggml_type_size(type);
+        data = (char *) data + (view_offs * type_size);
     }
 
     size_t obj_alloc_size = 0;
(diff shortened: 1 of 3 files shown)


slaren:
The offsets of the views are specified in bytes, and you should do the multiplication yourself in the call to `ggml_view_1d`. This change is not correct. Check other usages of `ggml_view_xd` in the examples.

danbev:
> The offsets of the views are specified in bytes, and you should do the multiplication yourself in the call to `ggml_view_1d`

Thanks for clarifying this, it was not clear to me.

PR #785, opened 2024-04-05
backend: remove duplicate backend typedefs

danbev:
This commit removes the duplicate typedefs for backend structs in `ggml-backend.h`. These typedefs are already defined in `ggml-alloc.h` which `ggml-backend.h` includes.

There might be a case for having the typedefs in `ggml-backend.h` but if this perhaps is an oversight I think it might better to remove them from `ggml-backend.h` to avoid confusion.

### DIFF
--- a/include/ggml/ggml-backend.h
+++ b/include/ggml/ggml-backend.h
@@ -7,10 +7,7 @@
 extern "C" {
 #endif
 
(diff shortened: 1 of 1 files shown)


danbev:
Closing as this is out of date.

PR #806, opened 2024-04-26
feat: implemented sigmoid function

justcho5:
- [x] implement sigmoid function in ggml
- [x] implement metal kernel
- [x] implement cuda kernel

This is required for [encodec.cpp](https://github.com/PABannier/encodec.cpp) and [bark.cpp](https://github.com/PABannier/bark.cpp).

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -511,6 +511,7 @@ extern "C" {
         GGML_UNARY_OP_TANH,
         GGML_UNARY_OP_ELU,
         GGML_UNARY_OP_RELU,
+        GGML_UNARY_OP_SIGMOID,
         GGML_UNARY_OP_GELU,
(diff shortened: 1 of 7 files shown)


justcho5:
@slaren Hey my PR is complete. Should I implement tests?

slaren:
In this case since it is implemented as a `GGML_OP_UNARY` there is already a test, all you have to do to enable it is modify the `supports_op` function of the backends to return `true` for `GGML_UNARY_OP_SIGMOID`. Then you can run `test-backend-ops` to test the CUDA and Metal implementations.

justcho5:
Thank you! Looks like the operation passed the tests on my machine.

PABannier:
That's super helpful! Thanks @justcho5 :)
Encodec.cpp will now be able to support the Metal backend :)

slaren (approved):

PR #808, opened 2024-04-29
feat : add `col2im` kernels

PABannier:
This PR implements the `col2im` kernel to support `ggml_conv_transpose_1d` on Metal. It proceeds as follows:
- [x] Write a C kernel for `col2im`
- [ ] Write a Metal kernel for `col2im`
- [ ] Remove the old implementation of `ggml_conv_transpose_1d`
- [ ] Implement the one using `ggml_col2im`


slaren:
When adding a new op to multiple backends, a test should be added in `test-backend-ops`. It should also be useful to test the implementation while working on it.

Issue #803, opened 2024-04-21
Behavior mismatch between PyTorch `GroupNorm` and `ggml_group_norm`


balisujohn:
Here are my two tester files which include what should be the same operation on the same tensor with PyTorch and ggml. The same tensor is used, and it's saved as a literal in the tester file in each case. It has dimension 43 X 1024.

Pytorch example:
````
print("after norm")
test_norm = torch.nn.GroupNorm(32,1024, eps=1e-06, affine=False).to('cuda')
print(test_norm(tensor))
````
https://github.com/balisujohn/ggml_pytorch_groupnorm_mismatch/blob/master/tester2.py

GGML example: 
````
struct ggml_tensor * result = ggml_group_norm(ctx0, model.a, 32);
````

https://github.com/balisujohn/ggml_pytorch_groupnorm_mismatch/blob/master/examples/simple/simple-backend.cpp

Pytorch Output:
````
after norm
tensor([[[ 1.3370,  1.5613, -0.2031,  ...,  1.9385,  1.7235,  0.7601],
         [ 0.6938, -1.0381,  1.1787,  ...,  1.5918,  1.9686,  1.5100],
         [-0.9839, -0.0334,  0.4642,  ...,  1.8995,  2.0359,  0.2274],
         ...,
         [ 0.2085, -0.2089,  0.8808,  ...,  2.0186,  2.2248,  1.3418],
         [-0.0949,  0.0766,  0.6567,  ..., -1.3484, -1.6598, -1.5336],
         [-0.4434,  0.0802,  1.0636,  ..., -1.7527, -1.9895,  0.2185]]])
````

GGML Output (DATA shows first 3 and last 3 entries of tensor in row major order):
````
NAME:
node_0
TYPE
0
SHAPE:
43
1024
1
1
DATA:
1.38015
1.6132
-0.220296
-1.69287
-1.90835
0.101279
````

torch version: 2.1.0
ggml version: https://github.com/ggerganov/ggml/commit/e1daebbf9d38d510ba456c4d50b4500a73ac2b14 (most recent as of time of posting)

**Comment by slaren** [comment]
Try this:
```C++
    ggml_tensor * a = ggml_reshape_3d(ctx0, model.a, model.a->ne[0], 1, model.a->ne[1]);
    struct ggml_tensor * result = ggml_group_norm(ctx0, a, 32);
```

`ggml_group_norm` normalizes over the first two dimensions (no idea why), and you can workaround that by moving the second dimension to the third.

[...]

PR #813, opened 2024-05-04
added comment explaining group norm behavior in ggml.h

balisujohn:
Adds a comment explaining behavior described in this issue: https://github.com/ggerganov/ggml/issues/803

### DIFF
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -1136,6 +1136,10 @@ extern "C" {
 
     // group normalize along ne0*ne1*n_groups
     // used in stable-diffusion
+    // calculated over first 2 dimensions
+    // To get normalization on first dimension only (pytorch default behavior),
+    // you can move the second dimension of tensor a to the third.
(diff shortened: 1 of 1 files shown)


slaren:
@leejet since you wrote `ggml_group_norm`, can you review if this documentation is accurate?

leejet:
The ggml_group_norm is not calculated over the first 2 dimensions, but rather splits ne2 (which is equivalent to num_channels in PyTorch) into n_groups (equivalent to num_groups in PyTorch) groups and then calculates along each group and dimensions ne0 and ne1. Since most scenarios using group norm involve 3-dimensional or 4-dimensional tensors, I simply assumed the group partitioning occurs along ne2. I've previously commented in the API, `// group normalize along ne0*ne1*n_groups` but in reality, it should be `// group normalize along ne0*ne1*(ne2/n_groups)`.

leejet:
Here is a simple schematic diagram: ne0 => W, ne1 => H, ne2 => C, ne3 => N.

norm

PR #821, opened 2024-05-13
ggml : expose SSE3 and SSSE3 for MSVC when AVX is available

przemoc:
Previously only SSE3 was exposed and only when AVX2 was available, but AVX is sufficient to expect presence of SSE3 and SSSE3.

---

Upstreaming change done in whisper.cpp:
https://github.com/ggerganov/whisper.cpp/pull/2128

### DIFF
--- a/src/ggml-impl.h
+++ b/src/ggml-impl.h
@@ -120,8 +120,15 @@ extern "C" {
 #ifndef __F16C__
 #define __F16C__
 #endif
+#endif
+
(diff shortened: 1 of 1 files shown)


ggerganov:
I believe these changes are now synced after the recent project sync

ggerganov:
Not described, but simply think about the `ggml` code as a git submodule - changes pushed in either `ggml`, `llama.cpp` and `whisper.cpp` are eventually synced to the rest. The syncs are done occasionally, mainly when there are enough significant changes that need to be propagated

PR #831, opened 2024-05-19
Embed yolo files

katsu560:
Some app like yolov3-tiny needs additional files to execute such as label(coco.names) and alphabet labels(100_0.png, ...) files.
If these files are embedded to a model(gguf) file and the app read them from the model file, the app is more portable.

I added below
- added new GGUF_TYPE_NAMEDOBJECT with name(file path) and value(file body) for adding files to gguf
- expanded gguf-py to support NAMEDOBJECT, constants.py, gguf_reader.py, gguf_writer.py
  - please see pull request to llama.cpp
- added gguf-addfile.py script to add files to gguf file
  - add files as NAMEDOBJECT (general.namedobject.N) or add files as NAMEDOBJECT array (general.namedobject[N] with --array option)
- expanded ggml to support NAMEDOBJECT, ggml.h ggml.c
- expanded yolov3-tiny to read coco.names and alphabet labels from gguf file,
  - at first read from gguf, then read from file if failed from gguf

NAMEDOBJECT constructed from name(file path) and value(file body)
```
    struct gguf_nobj {
        uint64_t nname;  // length of name
        char   * name;   // name in utf8
        uint64_t n;      // length of data in bytes
        char   * data;   // data body (file body)
    };
```

function usage:

```
struct gguf_nobj gguf_find_name_nobj(const struct gguf_context * ctx, const char * name)
```
call gguf_find_name_nobj() with const struct gguf_context *ctx and const char *name.
ctx is gguf_context pointer. name is string encoded UTF8 like filename.
search 'name' NAMEDOBJECT and return struct nobj.
if not found, return struct nobj(0, NULL, 0, NULL). so if nobj.n == 0 means 'not found'.
if found, return nobj with nobj.name has name, nobj.n has length of nobj.data, nobj.data has byte stream of data.

```
    struct gguf_nobj nobj = gguf_find_name_nobj(ctx, filename);
    if (nobj.n == 0) {
        return false;
    }
    membuf buf(nobj.data, nobj.data + nobj.n);
    std::istream file_in(&buf);
    if (!file_in) {
        return false;
    }
    std::string line;
    while (std::getline(file_in, line)) {
        labels.push_back(line);
    }
```

script usage:
```
python3 gguf-addfile.py [--array] input-gguf-file output-gguf-file files ...
```
 - add files as NAMEDOBJECT (general.namedobject.N)
 - add files as NAMEDOBJECT array (general.namedobject[N]) with --array option

### DIFF
--- a/examples/yolo/yolov3-tiny.cpp
+++ b/examples/yolo/yolov3-tiny.cpp
@@ -30,6 +30,7 @@ struct yolo_model {
     int height = 416;
     std::vector<conv2d_layer> conv2d_layers;
     struct ggml_context * ctx;
+    struct gguf_context * ggufctx;
 };
 
 struct yolo_layer {
@@ -71,6 +72,7 @@ static bool load_model(const std::string & fname, yolo_model & model) {
         fprintf(stderr, "%s: gguf_init_from_file() failed\n", __func__);
         return false;
     }
+    model.ggufctx = ctx;
     model.width  = 416;
     model.height = 416;
     model.conv2d_layers.resize(13);
@@ -100,9 +102,69 @@ static bool load_model(const std::string & fname, yolo_model & model) {
     return true;
 }
 
+// istream from memory
+#include <streambuf>
+#include <istream>
+
+struct membuf : std::streambuf {
+    membuf(const char * begin, const char * end) {
+        char * b(const_cast<char *>(begin));
+        char * e(const_cast<char *>(end));
+        this->begin = b;
+        this->end = e;
+        this->setg(b, b, e);
+    }
+
+    membuf(const char * base, size_t size) {
+        char * b(const_cast<char *>(begin));
+        this->begin = b;
+        this->end = b + size;
+        this->setg(b, b, end);
+    }
+    
+    virtual pos_type seekoff(off_type off, std::ios_base::seekdir dir, std::ios_base::openmode which = std::ios_base::in) override {
+        if(dir == std::ios_base::cur) {
+            gbump(off);
+        } else if(dir == std::ios_base::end) {
+            setg(begin, end + off, end);
+        } else if(dir == std::ios_base::beg) {
+            setg(begin, begin + off, end);
+        }
+
+        return gptr() - eback();
+    }
+    
+    virtual pos_type seekpos(std::streampos pos, std::ios_base::openmode mode) override {
+        return seekoff(pos - pos_type(off_type(0)), std::ios_base::beg, mode);
+    }
+    
+    char * begin;
+    char * end;
+};
+
+
 static bool load_labels(const char * filename, std::vector<std::string> & labels)
 {
     std::ifstream file_in(filename);
+    if (!file_in) {
+        return false;
+    }
+    std::string line;
+    while (std::getline(file_in, line)) {
+        labels.push_back(line);
+    }
+    GGML_ASSERT(labels.size() == 80);
+    return true;
+}
+
+static bool load_labels_kv(const struct gguf_context * ctx, const char * filename, std::vector<std::string> & labels)
+{
+    struct gguf_nobj nobj = gguf_find_name_nobj(ctx, filename);
+    if (nobj.n == 0) {
+        return false;
+    }
+    membuf buf(nobj.data, nobj.data + nobj.n);
+    std::istream file_in(&buf);
     if (!file_in) {
         return false;
     }
@@ -122,6 +184,27 @@ static bool load_alphabet(std::vector<yolo_image> & alphabet)
             char fname[256];
             sprintf(fname, "data/labels/%d_%d.png", i, j);
             if (!load_image(fname, alphabet[j*128 + i])) {
+                fprintf(stderr, "Cannot load '%s'\n", fname);
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+static bool load_alphabet_kv(const struct gguf_context * ctx, std::vector<yolo_image> & alphabet)
+{
+    alphabet.resize(8 * 128);
+    for (int j = 0; j < 8; j++) {
+        for (int i = 32; i < 127; i++) {
+            char fname[256];
+            sprintf(fname, "data/labels/%d_%d.png", i, j);
+            struct gguf_nobj nobj = gguf_find_name_nobj(ctx, fname);
+            if (nobj.n == 0) {
+                fprintf(stderr, "Cannot find '%s'\n", fname);
+                return false;
+            }
+            if (!load_image_from_memory(nobj.data, nobj.n, alphabet[j*128 + i])) {
                 fprintf(stderr, "Cannot load '%s'\n", fname);
                 return false;
             }
@@ -503,14 +586,20 @@ int main(int argc, char *argv[])
         return 1;
     }
     std::vector<std::string> labels;
-    if (!load_labels("data/coco.names", labels)) {
-        fprintf(stderr, "%s: failed to load labels from 'data/coco.names'\n", __func__);
-        return 1;
+    if (!load_labels_kv(model.ggufctx, "data/coco.names", labels)) {
+        fprintf(stderr, "%s: failed to load labels from 'data/coco.names' in model\n", __func__);
+        if (!load_labels("data/coco.names", labels)) {
+            fprintf(stderr, "%s: failed to load labels from 'data/coco.names'\n", __func__);
+            return 1;
+        }
     }
     std::vector<yolo_image> alphabet;
-    if (!load_alphabet(alphabet)) {
-        fprintf(stderr, "%s: failed to load alphabet\n", __func__);
-        return 1;
+    if (!load_alphabet_kv(model.ggufctx, alphabet)) {
+        fprintf(stderr, "%s: failed to load alphabet from model\n", __func__);
+        if (!load_alphabet(alphabet)) {
+            fprintf(stderr, "%s: failed to load alphabet\n", __func__);
+            return 1;
+        }
     }
     const int64_t t_start_ms = ggml_time_ms();
     detect(img, model, params.thresh, labels, alphabet);
--- a/include/ggml/ggml.h
+++ b/include/ggml/ggml.h
@@ -2258,20 +2258,28 @@ extern "C" {
     //
 
     enum gguf_type {
-        GGUF_TYPE_UINT8   = 0,
-        GGUF_TYPE_INT8    = 1,
-        GGUF_TYPE_UINT16  = 2,
-        GGUF_TYPE_INT16   = 3,
-        GGUF_TYPE_UINT32  = 4,
-        GGUF_TYPE_INT32   = 5,
-        GGUF_TYPE_FLOAT32 = 6,
-        GGUF_TYPE_BOOL    = 7,
-        GGUF_TYPE_STRING  = 8,
-        GGUF_TYPE_ARRAY   = 9,
-        GGUF_TYPE_UINT64  = 10,
-        GGUF_TYPE_INT64   = 11,
-        GGUF_TYPE_FLOAT64 = 12,
+        GGUF_TYPE_UINT8       = 0,
+        GGUF_TYPE_INT8        = 1,
+        GGUF_TYPE_UINT16      = 2,
+        GGUF_TYPE_INT16       = 3,
+        GGUF_TYPE_UINT32      = 4,
+        GGUF_TYPE_INT32       = 5,
+        GGUF_TYPE_FLOAT32     = 6,
+        GGUF_TYPE_BOOL        = 7,
+        GGUF_TYPE_STRING      = 8,
+        GGUF_TYPE_ARRAY       = 9,
+        GGUF_TYPE_UINT64      = 10,
+        GGUF_TYPE_INT64       = 11,
+        GGUF_TYPE_FLOAT64     = 12,
+        GGUF_TYPE_NAMEDOBJECT = 13,
         GGUF_TYPE_COUNT,       // marks the end of the enum
+    };
+
+    struct gguf_nobj {
+        uint64_t nname;
+        char   * name;
+        uint64_t n;
+        char   * data;
     };
 
     struct gguf_context;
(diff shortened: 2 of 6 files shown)


slaren:
Is it really necessary to a new type of object to the GGUF format to do this? The file data could be stored either as an array metadata or as a tensor.

ggerganov:
I agree with @slaren - don't think it's necessary to introduce named object. But the rest of the idea to embed the data in the GGUF file is nice

katsu560:
Thanks for prompt checking, @slaren and @ggerganov .
If current data structure meet embedding files, I agreed no adding NAMEDOBJECT.

But, I think embedding files need 3 elements, such as path name string(GGUF string 2 part as length and string byte stream), length of data, data stream.
I think key string and GGUF_TYPE_STRING has key string, length and bytes stream.
If we can use key string as path name, we can't embed same name file as existing key names, such as general.name, general.version, tokenizer.chat_template, etc.
And someone expects string body has no NULL byte in the way, but as you can see, file body has NULL byte(\0).
So, I added new type as NAMEDOBJECT.

CISC:
You can store the data in an UINT8 array, if you need to store path too you can store it as an array of arrays, ie: [[path, data], [path, data]], though I'm unsure if you would then have to store path as UINT8 too, or if it's allowed to have mixed data?

ggerganov:
You can for example store a KV string array with filenames and for each filename have a U8 tensor for each file containing the binary data:

- "embedded_files": ["my-file.dat", "another-file.bin"]
- tensors:
  - "my-file.dat"
  - "another-file.bin"
  - ...

katsu560:
Thanks comments, and sorry for my late response because of my hard working days.
I seek another way in this weekend, such as array of array, using tensor data.

katsu560:
finally, I added file data as follows;
- store the file path as key with starting '/' to avoid from conflicts to other key names.
    ex. storing file 'data/coco.names' as '/data/coco.names'
    if storing absolute file path '/a/b/c' as '//a/b/c'
- store the file contents as GGUF_TYPE_STRING's value.

So, I deleted all NAMEDOBJECT part.

katsu560:
I also remove dump code from gguf-addfile.py script.

examples/yolo/yolov3-tiny.cpp
```diff
@@ -30,6 +30,7 @@ struct yolo_model {
     int height = 416;
     std::vector<conv2d_layer> conv2d_layers;
     struct ggml_context * ctx;
+    struct gguf_context * ggufctx;
```
ggerganov, line 33:
```suggestion
    struct gguf_context * ctx_gguf;
```

include/ggml/ggml.h
```diff
@@ -2304,22 +2304,23 @@ extern "C" {
...
+    GGML_API uint64_t         gguf_get_val_u64    (const struct gguf_context * ctx, int key_id);
+    GGML_API int64_t          gguf_get_val_i64    (const struct gguf_context * ctx, int key_id);
+    GGML_API double           gguf_get_val_f64    (const struct gguf_context * ctx, int key_id);
+    GGML_API bool             gguf_get_val_bool   (const struct gguf_context * ctx, int key_id);
+    GGML_API const char *     gguf_get_val_str    (const struct gguf_context * ctx, int key_id);
+    GGML_API uint64_t         gguf_get_val_str_len(const struct gguf_context * ctx, int key_id);
+    GGML_API const void *     gguf_get_val_data   (const struct gguf_context * ctx, int key_id);
+    GGML_API int              gguf_get_arr_n      (const struct gguf_context * ctx, int key_id);
+    GGML_API const void *     gguf_get_arr_data   (const struct gguf_context * ctx, int key_id);
+    GGML_API const char *     gguf_get_arr_str    (const struct gguf_context * ctx, int key_id, int i);
```
ggerganov, line 2323:
```suggestion
    GGML_API uint8_t      gguf_get_val_u8     (const struct gguf_context * ctx, int key_id);
    GGML_API int8_t       gguf_get_val_i8     (const struct gguf_context * ctx, int key_id);
    GGML_API uint16_t     gguf_get_val_u16    (const struct gguf_context * ctx, int key_id);
    GGML_API int16_t      gguf_get_val_i16    (const struct gguf_context * ctx, int key_id);
    GGML_API uint32_t     gguf_get_val_u32    (const struct gguf_context * ctx, int key_id);
    GGML_API int32_t      gguf_get_val_i32    (const struct gguf_context * ctx, int key_id);
    GGML_API float        gguf_get_val_f32    (const struct gguf_context * ctx, int key_id);
    GGML_API uint64_t     gguf_get_val_u64    (const struct gguf_context * ctx, int key_id);
    GGML_API int64_t      gguf_get_val_i64    (const struct gguf_context * ctx, int key_id);
    GGML_API double       gguf_get_val_f64    (const struct gguf_context * ctx, int key_id);
    GGML_API bool         gguf_get_val_bool   (const struct gguf_context * ctx, int key_id);
    GGML_API const char * gguf_get_val_str    (const struct gguf_context * ctx, int key_id);
    GGML_API uint64_t     gguf_get_val_str_len(const struct gguf_context * ctx, int key_id);
    GGML_API const void * gguf_get_val_data   (const struct gguf_context * ctx, int key_id);
    GGML_API int          gguf_get_arr_n      (const struct gguf_context * ctx, int key_id);
    GGML_API const void * gguf_get_arr_data   (const struct gguf_context * ctx, int key_id);
    GGML_API const char * gguf_get_arr_str    (const struct gguf_context * ctx, int key_id, int i);
```

src/ggml.c
```diff
@@ -21946,18 +21946,18 @@ struct gguf_context * gguf_init_from_file(const char * fname, struct gguf_init_p
...
+                case GGUF_TYPE_UINT32:      ok = ok && gguf_fread_el  (file, &kv->value.uint32,  sizeof(kv->value.uint32),  &offset); break;
+                case GGUF_TYPE_INT32:       ok = ok && gguf_fread_el  (file, &kv->value.int32,   sizeof(kv->value.int32),   &offset); break;
+                case GGUF_TYPE_FLOAT32:     ok = ok && gguf_fread_el  (file, &kv->value.float32, sizeof(kv->value.float32), &offset); break;
+                case GGUF_TYPE_UINT64:      ok = ok && gguf_fread_el  (file, &kv->value.uint64,  sizeof(kv->value.uint64),  &offset); break;
+                case GGUF_TYPE_INT64:       ok = ok && gguf_fread_el  (file, &kv->value.int64,   sizeof(kv->value.int64),   &offset); break;
+                case GGUF_TYPE_FLOAT64:     ok = ok && gguf_fread_el  (file, &kv->value.float64, sizeof(kv->value.float64), &offset); break;
+                case GGUF_TYPE_BOOL:        ok = ok && gguf_fread_el  (file, &kv->value.bool_,   sizeof(kv->value.bool_),   &offset); break;
+                case GGUF_TYPE_STRING:      ok = ok && gguf_fread_str (file, &kv->value.str,                                &offset); break;
                 case GGUF_TYPE_ARRAY:
```
ggerganov, line 21961:
```suggestion
                case GGUF_TYPE_UINT8:   ok = ok && gguf_fread_el (file, &kv->value.uint8,   sizeof(kv->value.uint8),   &offset); break;
                case GGUF_TYPE_INT8:    ok = ok && gguf_fread_el (file, &kv->value.int8,    sizeof(kv->value.int8),    &offset); break;
                case GGUF_TYPE_UINT16:  ok = ok && gguf_fread_el (file, &kv->value.uint16,  sizeof(kv->value.uint16),  &offset); break;
                case GGUF_TYPE_INT16:   ok = ok && gguf_fread_el (file, &kv->value.int16,   sizeof(kv->value.int16),   &offset); break;
                case GGUF_TYPE_UINT32:  ok = ok && gguf_fread_el (file, &kv->value.uint32,  sizeof(kv->value.uint32),  &offset); break;
                case GGUF_TYPE_INT32:   ok = ok && gguf_fread_el (file, &kv->value.int32,   sizeof(kv->value.int32),   &offset); break;
                case GGUF_TYPE_FLOAT32: ok = ok && gguf_fread_el file, &kv->value.float32, sizeof(kv->value.float32), &offset); break;
                case GGUF_TYPE_UINT64:  ok = ok && gguf_fread_el (file, &kv->value.uint64,  sizeof(kv->value.uint64),  &offset); break;
                case GGUF_TYPE_INT64:   ok = ok && gguf_fread_el (file, &kv->value.int64,   sizeof(kv->value.int64),   &offset); break;
                case GGUF_TYPE_FLOAT64: ok = ok && gguf_fread_el (file, &kv->value.float64, sizeof(kv->value.float64), &offset); break;
                case GGUF_TYPE_BOOL:    ok = ok && gguf_fread_el (file, &kv->value.bool_,   sizeof(kv->value.bool_),   &offset); break;
                case GGUF_TYPE_STRING:  ok = ok && gguf_fread_str(file, &kv->value.str,                                &offset); break;
                case GGUF_TYPE_ARRAY:
```

src/ggml.c
```diff
@@ -22729,18 +22744,18 @@ static void gguf_write_to_buf(const struct gguf_context * ctx, struct gguf_buf *
...
+            case GGUF_TYPE_UINT16:  gguf_bwrite_el  (buf, &kv->value.uint16,  sizeof(kv->value.uint16) ); break;
+            case GGUF_TYPE_INT16:   gguf_bwrite_el  (buf, &kv->value.int16,   sizeof(kv->value.int16)  ); break;
+            case GGUF_TYPE_UINT32:  gguf_bwrite_el  (buf, &kv->value.uint32,  sizeof(kv->value.uint32) ); break;
+            case GGUF_TYPE_INT32:   gguf_bwrite_el  (buf, &kv->value.int32,   sizeof(kv->value.int32)  ); break;
+            case GGUF_TYPE_FLOAT32: gguf_bwrite_el  (buf, &kv->value.float32, sizeof(kv->value.float32)); break;
+            case GGUF_TYPE_UINT64:  gguf_bwrite_el  (buf, &kv->value.uint64,  sizeof(kv->value.uint64) ); break;
+            case GGUF_TYPE_INT64:   gguf_bwrite_el  (buf, &kv->value.int64,   sizeof(kv->value.int64)  ); break;
+            case GGUF_TYPE_FLOAT64: gguf_bwrite_el  (buf, &kv->value.float64, sizeof(kv->value.float64)); break;
+            case GGUF_TYPE_BOOL:    gguf_bwrite_el  (buf, &kv->value.bool_,   sizeof(kv->value.bool_)  ); break;
+            case GGUF_TYPE_STRING:  gguf_bwrite_str (buf, &kv->value.str                               ); break;
```
ggerganov, line 22758:
```suggestion
            case GGUF_TYPE_UINT8:   gguf_bwrite_el (buf, &kv->value.uint8,   sizeof(kv->value.uint8)  ); break;
            case GGUF_TYPE_INT8:    gguf_bwrite_el (buf, &kv->value.int8,    sizeof(kv->value.int8)   ); break;
            case GGUF_TYPE_UINT16:  gguf_bwrite_el (buf, &kv->value.uint16,  sizeof(kv->value.uint16) ); break;
            case GGUF_TYPE_INT16:   gguf_bwrite_el (buf, &kv->value.int16,   sizeof(kv->value.int16)  ); break;
            case GGUF_TYPE_UINT32:  gguf_bwrite_el (buf, &kv->value.uint32,  sizeof(kv->value.uint32) ); break;
            case GGUF_TYPE_INT32:   gguf_bwrite_el (buf, &kv->value.int32,   sizeof(kv->value.int32)  ); break;
            case GGUF_TYPE_FLOAT32: gguf_bwrite_el (buf, &kv->value.float32, sizeof(kv->value.float32)); break;
            case GGUF_TYPE_UINT64:  gguf_bwrite_el (buf, &kv->value.uint64,  sizeof(kv->value.uint64) ); break;
            case GGUF_TYPE_INT64:   gguf_bwrite_el (buf, &kv->value.int64,   sizeof(kv->value.int64)  ); break;
            case GGUF_TYPE_FLOAT64: gguf_bwrite_el (buf, &kv->value.float64, sizeof(kv->value.float64)); break;
            case GGUF_TYPE_BOOL:    gguf_bwrite_el (buf, &kv->value.bool_,   sizeof(kv->value.bool_)  ); break;
            case GGUF_TYPE_STRING:  gguf_bwrite_str(buf, &kv->value.str                               ); break;
```

ggerganov (commented):
Can you try to update `ci/run.sh` to use the new script in the tests?

Adding the files as tensors instead of KV can have some advantages, because the GGUF header remains small. For example, I'm not sure how the [GGUF viewer on HuggingFace](https://huggingface.co/docs/hub/en/gguf#viewer-for-metadata--tensors-info) would handle big data in the header. So it's an option that might be worth exploring

examples/yolo/yolov3-tiny.cpp
katsu560: applied

katsu560:
I revised code as to add files to tensor data.
I also applied your suggestions.

I try to update ci/run.sh later.

include/ggml/ggml.h
```diff
@@ -2342,11 +2343,13 @@ extern "C" {
...
-    GGML_API size_t         gguf_get_tensor_offset(const struct gguf_context * ctx, int i);
-    GGML_API char *         gguf_get_tensor_name  (const struct gguf_context * ctx, int i);
-    GGML_API enum ggml_type gguf_get_tensor_type  (const struct gguf_context * ctx, int i);
+    GGML_API int            gguf_get_n_tensors      (const struct gguf_context * ctx);
+    GGML_API int            gguf_find_tensor        (const struct gguf_context * ctx, const char * name);
+    GGML_API size_t         gguf_get_tensor_offset  (const struct gguf_context * ctx, int i);
+    GGML_API char *         gguf_get_tensor_name    (const struct gguf_context * ctx, int i);
+    GGML_API enum ggml_type gguf_get_tensor_type    (const struct gguf_context * ctx, int i);
+    GGML_API size_t         gguf_get_tensor_size    (const struct gguf_context * ctx, int i);
+    GGML_API int            gguf_find_and_get_tensor(const struct gguf_context * ctx, const char * name, char ** data, size_t * size);
```
ggerganov, line 2352: This is looking better, but still needs some work. Neither of the changes to `gguf` are needed, so try to avoid them and do the same using the existing API. The final PR should not contain any modifications to `ggml.h`. The only one that can remain is the `gguf_get_tensor_size()` helper function

katsu560:
I added two functions to ggml.c, gguf_get_tensor_size and gguf_find_key_array.
I think it is minimum adding.

katsu560:
I also revised ci/run.sh.
I added test code to create gguf file and test by yolov3-tiny for reading files from gguf file.

katsu560:
I fixed script gguf-addfile.py
- fix copying key value other than embedded_files
- refactor code
- remove unused code
- check overwriting output file
- add --force option

include/ggml/ggml.h
```diff
@@ -2305,6 +2305,7 @@ extern "C" {
 
     GGML_API int          gguf_get_n_kv(const struct gguf_context * ctx);
     GGML_API int          gguf_find_key(const struct gguf_context * ctx, const char * key);
+    GGML_API int          gguf_find_key_array(const struct gguf_context * ctx, const char * key, const char * val);
```
ggerganov, line 2308: No need to add `gguf_find_key_array()` - it's result is never used for anything, so we can simply remove it

katsu560:
deleted gguf_find_key_array() and related code from examples/yolo/yolov3-tiny.cpp.
please confirm.

examples/yolo/yolov3-tiny.cpp
```diff
@@ -114,6 +157,28 @@ static bool load_labels(const char * filename, std::vector<std::string> & labels
...
 }
 
+static bool load_labels_gguf(const struct gguf_context * ctx, const char * filename, std::vector<std::string> & labels)
+{
+    int tensor = gguf_find_tensor(ctx, filename);
+    if (tensor == -1) {
+        return false;
+    }
+    const size_t offset = gguf_get_tensor_offset(ctx, tensor);
+    const size_t len = gguf_get_tensor_size(ctx, tensor);
```
ggerganov, line 167: Somehow I didn't notice this before: `gguf_get_tensor_size()` is not needed too. You can instead use:

```suggestion
    const size_t len = ggml_nelemens(tensor);
```

So remove `gguf_get_tensor_size` all together

katsu560: okay, i removed gguf_get_tensor_size from ggml.h, ggml.c, yolov3-tiny.cpp.

katsu560:
I applied minor changes. please check.

PR #839, opened 2024-05-27
fix YaRN for NeoX type

### LINKED ISSUE #835: issues about YaRN
I am using YaRN when implementing DeepSeek V2 models. And current YaRN does not look good me.

@cebtenzzre Could you take a look on this? Correct me if I am wrong.

1. `theta_base *= freq_scale` is done again later in `rope_yarn`:

    https://github.com/ggerganov/ggml/blob/0cbb7c0e053f5419cfbebb46fbf4d4ed60182cf5/src/ggml.c#L14077

    In a _basic_ case (`ext_factor` is 0), the `theta` uses for `cos/sin` is scaled by `freq_scale * freq_scale`. I think this is wrong and this line should be deleted. 

1. value passed to `int64_t i0` is wrong: (data type does not matches, either.)

    https://github.com/ggerganov/ggml/blob/0cbb7c0e053f5419cfbebb46fbf4d4ed60182cf5/src/ggml.c#L14082-L14088

    I think it should be `ic` here.

    (Confirmed when implementing DeepSeek V2 models)

foldl:
This PR will fix #835.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -13831,7 +13831,6 @@ static void ggml_compute_forward_rope_f32(
     int ir = 0;
 
     const float theta_scale = powf(freq_base, -2.0f/n_dims);
-    const float inv_ndims = -1.f/n_dims;
     float corr_dims[2];
     ggml_rope_yarn_corr_dims(n_dims, n_orig_ctx, freq_base, beta_fast, beta_slow, corr_dims);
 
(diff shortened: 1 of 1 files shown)


ggerganov:
Thanks! Do the GPU kernels need to be updated?

foldl:
GPU backends:

1. CUDA: looks Good.
2. Kompute: not sure. Where is `shaderop_rope_xx.h`?
3. Metal: Fixed.
4. SYCL: looks Good.
5. Vulkan: Fixed.

But, unfortunately, I don't have environments to test these GPU fixes.

ggerganov:
Had top merge a few changes from `llama.cpp` and we have some conflicts - sorry about that.
After resolving these, could you try to extend `test-backend-ops` to exercise this code via `ggml_rope_ext()`?

foldl:
Sorry, but I don't think I have time to update `test-backend-ops`. Maybe later, or someone else could help on this.

ggerganov:
superseded by https://github.com/ggerganov/llama.cpp/pull/7617

PR #842, opened 2024-05-28
Fix compiler warning: this use of defined may not be portable

li-plus:
I compile GGML library with `-Wpedantic` flag and get this warning (https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wexpansion-to-defined):
```
warning: this use of "defined" may not be portable [-Wexpansion-to-defined]
  373 | #if FP16_AVAILABLE
      |     ^~~~~~~~~~~~~~
```

Fixed in this PR.


li-plus:
That's true. Is there another way to silence this warning?

slaren:
This should be fixed now.

PR #844, opened 2024-05-29
ggml : add rope tests + fix neox

ggerganov:


ggerganov:
moved to https://github.com/ggerganov/llama.cpp/pull/7617

PR #847, opened 2024-06-02
Update HIPBLAS CMake

daniandtheweb:
The old code for building HIPBLAS with CMake didn't work at all with my self-built ROCm sdk so I've updated it based on the llama.cpp implementation.

This new CMake code works with the self-built packages and it should keep the compatibility with the typical ROCm install made with the package manager (testing is needed for that).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -258,7 +258,20 @@ endif()
 
 # TODO: do not build separate ggml-rocm target (see CUDA build above, or llama.cpp for reference)
 if (GGML_HIPBLAS)
-    list(APPEND CMAKE_PREFIX_PATH /opt/rocm)
+    if ($ENV{ROCM_PATH})
+        set(ROCM_PATH $ENV{ROCM_PATH})
+    else()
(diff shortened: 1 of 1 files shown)


daniandtheweb:
Right now I've tested on Debian with the package manager HIPBlas and everything works fine.

daniandtheweb:
Setting the CC and CXX to rocm's clang and clang++ on Ubuntu 22.04, the same as llama.cpp's CI, works fine.

daniandtheweb:
When I try to replicate the build process from llama.cpp with this: 
`cmake -B build -S . -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang" -DGGML_HIPBLAS=ON`; the build fails while compiling the tests with the error: `unrecognized command-line option '--hip-link'`.

However building specifying -DCMAKE_C_COMPILER and -DCMAKE_CXX_COMPILER instead of -DCMAKE_HIP_COMPILER results in a working build.

ggerganov (approved):

Issue #819, opened 2024-05-12
`ggml_flip` or `ggml_pad_reflect`?


PABannier:
Hello,

I have implemented a custom operation [`ggml_pad_reflec_1d`](https://github.com/ggerganov/ggml/blob/aa00e1676417d8007dbaa47d2ee1a6e06c60d546/src/ggml.c#L7635) on my ggml fork. This is [required for Encodec.cpp](https://github.com/PABannier/encodec.cpp/blob/main/encodec.cpp#L274).
Should I upstream this operation? I would write a `ggml_pad_reflect` supporting 1D and 2D input, as in [PyTorch's nn.ReflectionPad2d](https://pytorch.org/docs/stable/generated/torch.nn.ReflectionPad2d.html)

Alternatively, I was considering implementing a `ggml_flip` operation ([PyTorch's flip](https://pytorch.org/docs/stable/generated/torch.flip.html)). This would allow us to implement a `pad_reflect` operation directly in 'user space' using flipping instead of adding a quite niche operation (`pad_reflect`) to `ggml`.

What are your thoughts? @slaren @ggerganov

ggerganov:
> Alternatively, I was considering implementing a ggml_flip operation ([PyTorch's flip](https://pytorch.org/docs/stable/generated/torch.flip.html)). This would allow us to implement a pad_reflect operation directly in 'user space' using flipping instead of adding a quite niche operation (pad_reflect) to ggml.

I don't have a good idea which option would be more useful, so whatever you think makes more sense - PR welcome

balisujohn:
I like `ggml_pad_reflect` as I need this functionality for tortoise.cpp. @PABannier can I have your permission to turn that code you have there into a PR for ggml so I can get this quickly upstreamed(I don't want to take your credit for the PR without asking).

balisujohn:
I ask because I need a working version of this to proceed with tortoise so I'm going to have to integrate it into at least my local copy of ggml anyway.

PABannier:
@balisujohn Please go on! :)

PR #850, opened 2024-06-06
feat: adds `ggml_pad_reflect_1d`

balisujohn:
This adds an op that mirrors the behavior or PyTorch's [ReflectionPad1d](https://pytorch.org/docs/stable/generated/torch.nn.ReflectionPad1d.html) operation. 

Implementations for cuda and CPU are provided.

Tests still need to be added, so marking as a draft PR from now. 

The cpu version of the op was derived with permission from @PABannier's implementation discussed here: https://github.com/ggerganov/ggml/issues/819


balisujohn:
I added a test in `test-backend-ops` and also added `test-pad-reflect-1d.cpp` which I verified works with and without the `#define GGML_USE_CUBLAS`, though I'm not sure if the cuda versions of the ops are currently tested in the CI/CD or how to accomplish that. 

I think this is ready for review.

balisujohn:
A test I wasn't sure how to add was "check if it correctly fails when called with a pad length not shorter than ne0 of the input tensor."

balisujohn:
Or and wrt the `is_2D` argument I support it but I'd ideally like to leave it to a future PR.

ggerganov:
Some similar comments as in the `ggml_conv_transpose_1d` PR

Regarding the `is_2D`: currently the `im2col` is an exception - the only op that uses this pattern. Probably it is better to change `im2col` to not use it, since the alternative pattern with dimensions suffix (i.e. `_1d`, `_2d`, etc.) is well established

balisujohn:
OK ready for review again : ^)

PABannier:
Hello @slaren @ggerganov ! Do you plan to merge this operation? This would be super helpful for [Encodec.cpp](https://github.com/PABannier/encodec.cpp) to support the Metal backend.

ggerganov:
Hey @PABannier, this PR needs to be updated to the latest `ggml` changes before merging.

PABannier:
Hello @balisujohn ! Do you plan to finish this PR soon? Otherwise, I'm happy to finish yours if you give me the permission to push on your fork.

PR #858, opened 2024-06-16
[SYCL] remove global variables (#7710)

Nuullll:
* separate DPCT helpers outside

* replace global variables with context

* remove useless extra

* update mul_mat condition

* remove duplicate buft initialization

* remove duplicate extra and global work group size

* remove useless backend check

* remove duplicated extras

* use macro for group_size and remove cuda-related

### DIFF
--- a/src/ggml-sycl/backend.hpp
+++ b/src/ggml-sycl/backend.hpp
@@ -1,0 +1,18 @@
+//
+// MIT license
+// Copyright (C) 2024 Intel Corporation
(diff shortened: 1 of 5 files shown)


Nuullll:
The sync script didn't catch ggml-sycl folder. This is a manual sync of https://github.com/ggerganov/llama.cpp/pull/7710

ggerganov:
Oh thanks for noticing - let me see if I can fix the sync scripts

PR #864, opened 2024-06-18
feat: extends `ggml_pad` to allow prefix padding

balisujohn:
modifies `ggml_pad` so you can specify beginning and end padding for the first 2 dimensions of a tensor. Previously, padding could only be added at the end of the first 2 dimensions. The padding is still constant padding with a value of zero, as before. 

Adds new test file `test-pad.cpp`


balisujohn:
I forgot that `ggml_pad` actually supports 4 dimensions as-is, I will add the support for pre-fix padding in all 4 dimensions, so marking as  a draft for now.

balisujohn:
Alright so it supports up to 4d prefix and postfix padding with the CPU version. for cuda it supports up to 3d prefix and postfix padding (the cuda pad kernel was previously limited to 3d tensors and I didn't remove this limitation). Two of the tests in `test-pad.cpp` use 4d tensors so they would need to be commented out to run the tests in cuda mode, though I verified all the other tests work with the cuda kernel. 

leaves the arguments of `ggml_pad` as-is, and adds `ggml_pad_ext` which requires the desired prefix and postfix length to be specified for all 4 dimensions. 

Ready for review :^)

balisujohn:
(I noticed I need to change some variable names in `test-pad.cpp` and there are similar minor fixed needed in the conv-transpose-1d PR)

balisujohn:
this and pad_reflect_1d are ready for review again :^)

PR #874, opened 2024-06-29
Fix bug: missiong pthread link.

delldu:
Test steps
1) mkdir build
2) cd build
3) cmake -DGGML_CUDA=ON -DBUILD_SHARED_LIBS_DEFAULT=OFF -DCMAKE_CUDA_COMPILER=/usr/local/cuda-11.7/bin/nvcc ..
4) make -j 12
/usr/bin/ld: CMakeFiles/test-backend-ops.dir/test-backend-ops.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -426,6 +426,6 @@ set_property(TEST ${TEST_TARGET} PROPERTY ENVIRONMENT "LLVM_PROFILE_FILE=${TEST_
 
 set(TEST_TARGET test-backend-ops)
 add_executable(${TEST_TARGET} ${TEST_TARGET}.cpp)
-target_link_libraries(${TEST_TARGET} PRIVATE ggml)
+target_link_libraries(${TEST_TARGET} PRIVATE ggml pthread)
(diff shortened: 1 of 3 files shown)


tests/CMakeLists.txt
```diff
@@ -426,6 +426,6 @@ set_property(TEST ${TEST_TARGET} PROPERTY ENVIRONMENT "LLVM_PROFILE_FILE=${TEST_
 
 set(TEST_TARGET test-backend-ops)
 add_executable(${TEST_TARGET} ${TEST_TARGET}.cpp)
-target_link_libraries(${TEST_TARGET} PRIVATE ggml)
+target_link_libraries(${TEST_TARGET} PRIVATE ggml pthread)
```
slaren, line 429: This will not work on every OS, eg. it fails on windows. It's not clear why this is needed in your system - `test-backend-ops` does not use pthread, it uses the C++ standard library. If it is really necessary, this may need to be changed to use the `Threads` package instead.

PR #885, opened 2024-07-09
cmake : only enable GGML_NATIVE and x86 flags if not crosscompiling

iboB:
The x86 arch flags are not applicable for Android x86 as per: https://developer.android.com/ndk/guides/abis#86-64

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,9 +50,15 @@ else()
     set(GGML_BLAS_VENDOR_DEFAULT "Generic")
 endif()
 
+if (CMAKE_CROSSCOMPILING)
+    set(GGML_NATIVE_DEFAULT OFF)
+else()
+    set(GGML_NATIVE_DEFAULT ON)
+endif()
+
 # general
 option(GGML_STATIC "ggml: static link libraries"         OFF)
-option(GGML_NATIVE "ggml: enable -march=native flag"     ON)
+option(GGML_NATIVE "ggml: enable -march=native flag"     ${GGML_NATIVE_DEFAULT})
 option(GGML_LTO    "ggml: enable link time optimization" OFF)
 option(GGML_CCACHE "ggml: use ccache if available"       ON)
 
@@ -70,7 +76,7 @@ option(GGML_SANITIZE_ADDRESS   "ggml: enable address sanitizer"   OFF)
 option(GGML_SANITIZE_UNDEFINED "ggml: enable undefined sanitizer" OFF)
 
 # instruction set specific
-if (GGML_NATIVE)
+if (GGML_NATIVE OR NOT GGML_NATIVE_DEFAULT)
     set(INS_ENB OFF)
 else()
     set(INS_ENB ON)


slaren:
Is this just for compliance with eg. the play store, or does the compiler actually reject these flags?

iboB:
> Is this just for compliance with eg. the play store, or does the compiler actually reject these flags? Generally the way people can deal with different architectures is by building multiple versions of ggml and choosing which one to load at run time. This would prevent that.

The binary crashes with `SIGILL ILL_ILLOPN` when it encounters the instructions which the compiler generated because of these flags

slaren:
That is the same that happens in every other platform. I don't really see any reason to treat Android differently, either build natively with `GGML_NATIVE` or set the correct arch flags.

iboB:
> That is the same that happens in every other platform. I don't really see any reason to treat Android differently, either build natively with `GGML_NATIVE` or set the correct arch flags.

Well, you might say that for Windows, but there is Windows-specific code. 

Also I would agree if the defaults were to have these flags off. Thus the Android x86 build would naturally work and one would have to do something to make it crash. But in this case the defaults are on which just breaks Android x86 with a cryptic crash. Since Android x86 is guaranteed not to have these instruction sets, with this small change you can save a lot of headaches to people who target the platform. 

There is no compilation error. Just a cryptic crash. Many people would have trouble debugging this. 

The workaround would be to explicitly set `LLAMA_NATIVE, LLAMA_AVX, LLAMA_AVX2, LLAMA_FMA, LLAMA_F16C` to `OFF`.

slaren:
I am not sure that Android is not guaranteed to have these instruction sets, it is just not guaranteed that every device will support them, but that's the same situation as in windows or linux. Changing the default for Android sounds good, it should be a simple change:
```
if (GGML_NATIVE OR ANDROID)
    set(INS_ENB OFF)
else()
    set(INS_ENB ON)
endif()
```

iboB:
> I am not sure that Android is not guaranteed to have these instruction sets, it is just not guaranteed that every device will support them, but that's the same situation as in windows or linux. Changing the default for Android sounds good, it should be a simple change:
> 
> ```cmake
> if (GGML_NATIVE OR ANDROID)
>     set(INS_ENB OFF)
> else()
>     set(INS_ENB ON)
> endif()
> ```

By guaranteed not to have them I mean that the ABI does not have them. Using AVX changes the calling convention. This means that even if you know you have the AVX instruction set on android, you can't compile with `-mavx` as this will produce uncallable functions (ones that take float args). You can check for availability in your code and decide to use the intrinsics (or inline assembly if you wish), but you can't compile with `-mavx`. Such checks are way beyond the scope of ggml, I think. 

I'm not familiar with existing Android x86 devices, but this is mostly relevant for a simulator on a x86 host (that's how I test it). My host absolute has AVX, but I still get `ILLOPN` the moment I call a function from llama.cpp

With the change you suggested one would just still need to explicitly set `GGML_NATIVE` to `OFF` for Android builds.

Your suggestion is in global scope while mine only affects the x86 branch of the encompassing if. I'm not sure what `INS_ENB` means, but could it also guard Arm-specific options in the future? If it does, then your suggestion would potentially deteriorate Android builds.

What's wrong with just checking for Android in the x86 branch? That's the only affected platform by the PR and in only affects relevant code.

slaren:
`INS_ENB` is used in the top level `CMakeLists.txt` to set the default values of the x86 instruction set flags, so arm would not be affected. It's just a convenient place for changing this since there is already a variable that does this.

You should always disable `GGML_NATIVE` when cross compiling for a different platform, so I don't think it's unreasonable to require users to disable this flag when cross compiling for x86 android on an x86 machine. The reason I don't like the way this PR makes does this is because it changes the behavior of the build script in a way that is very hard to predict, and it doesn't leave any way to enable these architecture flags. I was not aware of AVX or AVX2 changing the calling convention and I could not find any information about it. I could see that happening for static functions, but I would be very surprised if the calling convention exported functions changed depending on the architecture flags. I would appreciate it if you could link me to any information about this.

iboB:
> if you can figure a way to check for that in cmake (eg. check if the native platform and the target platform are different) that could potentially solve the issue.

I can, actually. That's a good idea. I'll force-push a change.

> I was not aware of AVX or AVX2 changing the calling convention and I could not find any information about it. I could see that happening for static functions, but I would be very surprised if the calling convention exported functions changed depending on the architecture flags. I would appreciate it if you could link me to any information about this.

Oops. Yes, you're right. Only static functions are affected. And this makes much more sense, indeed. Otherwise avx libraries wouldn't be usable from non-avx executables 😄

CMakeLists.txt
```diff
@@ -50,9 +50,15 @@ else()
     set(GGML_BLAS_VENDOR_DEFAULT "Generic")
 endif()
 
+if (CMAKE_CROSSCOMPILING)
+    set(GGML_NATIVE_BUILD OFF)
+else()
+    set(GGML_NATIVE_BUILD ON)
+endif()
```
slaren, line 57:
```suggestion
if (CMAKE_CROSSCOMPILING)
    set(GGML_NATIVE_DEFAULT OFF)
else()
    set(GGML_NATIVE_DEFAULT ON)
endif()
```

CMakeLists.txt
```diff
@@ -50,9 +50,15 @@ else()
...
+
 # general
 option(GGML_STATIC "ggml: static link libraries"         OFF)
-option(GGML_NATIVE "ggml: enable -march=native flag"     ON)
+option(GGML_NATIVE "ggml: enable -march=native flag"     ${GGML_NATIVE_BUILD})
```
slaren, line 61:
```suggestion
option(GGML_NATIVE "ggml: enable -march=native flag"     ${GGML_NATIVE_DEFAULT})
```

CMakeLists.txt
```diff
@@ -70,7 +76,7 @@ option(GGML_SANITIZE_ADDRESS   "ggml: enable address sanitizer"   OFF)
 option(GGML_SANITIZE_UNDEFINED "ggml: enable undefined sanitizer" OFF)
 
 # instruction set specific
-if (GGML_NATIVE)
+if (GGML_NATIVE OR NOT GGML_NATIVE_BUILD)
```
slaren, line 79:
```suggestion
if (GGML_NATIVE OR CMAKE_CROSSCOMPILING)
```

slaren (approved):
Minor style changes.

CMakeLists.txt
iboB: I'd rather have the GGML calculated value here, as `CMAKE_CROSSCOMPILING` may lead to false negatives on some rare occasions (and even false positives on even rarer occasions). In case we ever handle them, `GGML_NATIVE_DEFAULT` will carry the correct information.

slaren: Sounds good.
```suggestion
if (GGML_NATIVE OR NOT GGML_NATIVE_DEFAULT)
```

iboB: resolved in force-pushed commit

PR #893, opened 2024-07-18
Vulkan: Initialize vk_buffer_struct handle members to VK_NULL_HANDLE

neobrain:
This prevents invalid frees when destroying a partially initialized vk_buffer_struct. For example, this could happen in `ggml_vk_create_buffer` when running out of device memory.


ggerganov:
cc @0cc4m for review

0cc4m (approved):
That looks good, thank you.

PR #898, opened 2024-07-23
Loop tiling optimizations for scalar path

heshpdx:
Apply a loop tiling technique to the generic path, which provides performance upside for ISAs with enough registers to take advantage of it. Also helps the compiler optimize this path.

For SPEC CPU work, when measured on an Ampere Altra with gcc-13 -O3, llama-cpp showed +20% and whisper-cpp showed +4%.

### DIFF
--- a/src/ggml-quants.c
+++ b/src/ggml-quants.c
@@ -4301,15 +4301,18 @@ void ggml_vec_dot_q4_0_q8_0(int n, float * restrict s, size_t bs, const void * r
     float sumf = 0.0;
 
     for (int i = 0; i < nb; i++) {
-        int sumi = 0;
+        int sumi0 = 0;
+        int sumi1 = 0;
 
         for (int j = 0; j < qk/2; ++j) {
             const int v0 = (x[i].qs[j] & 0x0F) - 8;
             const int v1 = (x[i].qs[j] >>   4) - 8;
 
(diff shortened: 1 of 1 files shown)


src/ggml-quants.c
```diff
@@ -4949,18 +4955,21 @@ void ggml_vec_dot_q5_0_q8_0(int n, float * restrict s, size_t bs, const void * r
...
+        int sumi1 = 0;
 
         for (int j = 0; j < qk/2; ++j) {
             const uint8_t xh_0 = ((qh & (1u << (j + 0 ))) >> (j + 0 )) << 4;
             const uint8_t xh_1 = ((qh & (1u << (j + 16))) >> (j + 12));
 
-            const int32_t x0 = ((x[i].qs[j] & 0x0F) | xh_0) - 16;
-            const int32_t x1 = ((x[i].qs[j] >>   4) | xh_1) - 16;
+            const int32_t x0 = (int8_t)(((x[i].qs[j] & 0x0F) | xh_0) - 16);
+            const int32_t x1 = (int8_t)(((x[i].qs[j] >>   4) | xh_1) - 16);
```
slaren, line 4966: Are these casts necessary?

heshpdx: Functionally they are not needed, but if they exist then gcc can generate the dot-product instruction.

slaren (approved):

PR #899, opened 2024-07-27
sync : llama.cpp

ggerganov:


0cc4m:
I'm not sure what the normal process to sync llama.cpp and ggml is, but it seems Vulkan is broken on both the master and the sync branch (Missing shaders, inconsistent code state). Is there something I can do to help with that?

ggerganov:
Ah, thanks for spotting. Adding a CI workflow that builds with Vulkan would help.

I'll see if I can fix the build before merging this PR

ggerganov:
Should be good now - does it build on your side too @0cc4m?

We can add the CI on `master` later

0cc4m:
> Should be good now - does it build on your side too @0cc4m?
> 
> We can add the CI on `master` later

It does build now, thank you.

zcbenz:
@ggerganov There are some files like `ggml/include/ggml-amx.h` created in a new `/ggml` root directory, is there a reason that they are put in `/ggml` instead of `/src` like other files?

ggerganov:
Oh, the sync script messed up somehow and I didn't even notice. Thanks for letting me know.

PR #900, opened 2024-07-28
add t5 layer norm support

leejet:
The t5 layer norm does not subtract the mean during calculation, so I made a simple modification to ggml_norm to adapt it. This is a straightforward change and has already been validated in the sd.cpp project (since sd3 uses t5). However, due to limitations in my development environment, I have not been able to validate all the modifications. So far, I have only tested the changes for CUDA and CPU. Additionally, I am not very familiar with some environments, such as Kompute, so I have not developed corresponding support. I hope developers familiar with these environments can help improve it.

Ref: https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L230

### DIFF
--- a/include/ggml.h
+++ b/include/ggml.h
@@ -1117,6 +1117,18 @@ extern "C" {
             struct ggml_tensor  * a);
 
     // normalize along rows
+    GGML_API struct ggml_tensor * ggml_norm_ext(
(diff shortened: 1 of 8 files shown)


leejet:
ggml_rms_norm is what i need.

PR #905, opened 2024-07-30
metal : add abort callback

conradev:
This allows aborting metal execution midway through a compute graph.

If an abort callback is set, only two command buffers are enqueued at a time and ggml calls the abort callback after every command buffer executions. I am not able to measure a performance difference with an abort callback configured


conradev:
@ggerganov did I put this PR up in the right repo?

I also have a follow-up change to `llama.cpp` so that `llama_decode` can be aborted (and return an error code when that happens)

ggerganov (approved):

PR #906, opened 2024-07-31
ggml: ignore more msvc warnings

iboB:


src/ggml.c
```diff
@@ -53,6 +53,9 @@
 // disable POSIX deprecation warnings
 // these functions are never going away, anyway
 #pragma warning(disable: 4996)
+
+// unreachable code because of multiple instances of code after GGML_ABORT
+#pragma warning(disable: 4702)
```
slaren, line 58: I tried to remove all the code after `GGML_ABORT`, if I missed something it would be better to fix that instead. I do not get any unreachable code warnings with clang or MSVC 19.40.33813.0.

iboB: The warning is disabled by default. To repro, try building with `/w34702` or with `/w4`. Or just give me your branch and I'll test it.

Alternatively, if removing the unreachable code is desired, I can do that instead in this PR

slaren (approved):

PR #909, opened 2024-08-01
Perf optimization for generic ggml_vec_dot_q5

heshpdx:
More optimizations related to #898. That one showed {20%, 4%} for whisper and llama, and this one is an additional {10%, 2%} on top of that, as measured on an Ampere Altra. We discovered that we can reduce the amount of computations when `qh` is zero inside the loop. So the perf improvement depends on the model and the query, etc.

For those watching, the scalar path is important for benchmarking in SPEC CPUv8, as it allows a complete apples-to-apples comparison of any system, past or future. For example, SPEC CPU 2000 is still used for testing compilers and systems today, since the same code paths are used when running x86-64, aarch64, riscv, and power10, and NONE of those ISAs existed back in 2000. Similarly, when CPUv8 comes out it will live for a long time; and we want to establish the best baseline we can now. Regardless of whether ggml will be chosen from the list of benchmark candidates for CPUv8, this type of work can help the wider community. We hope that the optimizations we offer in the scalar path can provide hints to the folks crafting the ISA optimized paths, to make those even faster.

### DIFF
--- a/src/ggml-quants.c
+++ b/src/ggml-quants.c
@@ -4832,15 +4832,25 @@ void ggml_vec_dot_q5_0_q8_0(int n, float * restrict s, size_t bs, const void * r
         int sumi0 = 0;
         int sumi1 = 0;
 
-        for (int j = 0; j < qk/2; ++j) {
-            const uint8_t xh_0 = ((qh & (1u << (j + 0 ))) >> (j + 0 )) << 4;
-            const uint8_t xh_1 = ((qh & (1u << (j + 16))) >> (j + 12));
+        if (qh) {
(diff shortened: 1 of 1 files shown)


slaren:
I have to say, this optimization seems very unlikely. What is the chance of all the bits being zero? I would guess that if the buckets are distributed evenly, the chance should be very, very low. My intuition would be that the cost of an occasional mispredicted branch would be far higher than the few instructions that this would save. Are you using data representative of real conditions?

heshpdx:
Sorry about this, I got my wires crossed! Closing this for now.

PR #914, opened 2024-08-05
ggml: support forward pass broadcasting in ggml_sub

smeso:
Add support for subtraction between _broadcastable_ tensors.
Only for the forwards pass.
The code is similar to the code used for other operations.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -4597,11 +4597,13 @@ static struct ggml_tensor * ggml_sub_impl(
         struct ggml_tensor * a,
         struct ggml_tensor * b,
         bool inplace) {
-    GGML_ASSERT(ggml_are_same_shape(a, b));
+    GGML_ASSERT(ggml_can_repeat(b, a));
 
     bool is_node = false;
 
     if (!inplace && (a->grad || b->grad)) {
+        // TODO: support backward pass for broadcasting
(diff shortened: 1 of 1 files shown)


src/ggml.c
```diff
@@ -10036,11 +10038,10 @@ static void ggml_compute_forward_sub_f32(
     const struct ggml_tensor * src0 = dst->src[0];
     const struct ggml_tensor * src1 = dst->src[1];
 
-    if (params->ith != 0) {
-        return;
-    }
+    GGML_ASSERT(ggml_can_repeat(src1, src0) && ggml_are_same_shape(src0, dst));
```
ggerganov, line 10041: This should remain `assert` because we've already guaranteed these conditions in `ggml_sub_impl` and here we are mostly hinting the developer about what inputs to expect and verify we don't have a bug in the internal `ggml` logic.

In contrast, in `ggml_sub_impl` we have to use `GGML_ASSERT` instead of `assert` because we don't know what the user of the function will pass as input, so we have to check even in non-debug builds

ggerganov (approved):

src/ggml.c
smeso: Thanks for the fast review!
This should be fixed.

smeso:
I amended the commit message because I mentioned the wrong function name in it.

PR #921, opened 2024-08-12
feat: implement backward computation for more operators

Ronsor:
This PR will add backward computations for most operators once completed.

- [x] Tanh
- [x] Sigmoid
- [ ] GELU + GELU (quick)
- [ ] ELU
- [ ] clamp
- [ ] im2col
- [ ] pad
- [ ] LeakyReLU
- [ ] mean
- [ ] norm
- [ ] concat

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -18212,8 +18212,15 @@ static void ggml_compute_backward(struct ggml_context * ctx, struct ggml_tensor 
                         } break;
                     case GGML_UNARY_OP_TANH:
                         {
-                            GGML_ABORT("fatal error"); // TODO: not implemented
-                        }
+                            if (src0->grad) {
+                                src0->grad = ggml_add_or_set(ctx,
+                                        src0->grad,
+                                        ggml_sub(ctx,
+                                            tensor->grad,
+                                            ggml_mul(ctx, ggml_sqr(ctx, tensor), tensor->grad)),
+                                        zero_table);
+                            }
+                        } break;
                     case GGML_UNARY_OP_ELU:
                         {
                             GGML_ABORT("fatal error"); // TODO: not implemented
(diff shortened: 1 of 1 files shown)


src/ggml.c
```diff
@@ -14160,6 +14449,97 @@ static void ggml_compute_forward_clamp(
...
+        case GGML_TYPE_Q4_0_8_8:
+        case GGML_TYPE_I8:
+        case GGML_TYPE_I16:
+        case GGML_TYPE_I32:
+        case GGML_TYPE_I64:
+        case GGML_TYPE_F64:
+        case GGML_TYPE_COUNT:
+            {
+                GGML_ABORT("fatal error");
+            }
```
ggerganov, line 14539: Change to `default:`

ggerganov (commented):
Should add tests to `tests/test-grad0.cpp`

JohannesGaessler:
I'm currently working on adding training support for the MNIST example in https://github.com/ggerganov/ggml/pull/908 . I have a working backward pass for `im2col` and `pool2d` (the ops needed for the convolutional neural network). I'm currently working on cleaning up the code and putting it into a state that can be reviewed. When I added tests to `test-grad0` I also added a fix to deal with noncontinuous gradients when numerically calculating the gradients to compare against backpropagation; this fix or an equivalent one will also be needed for `clamp`.

ggerganov:
It might be better to wait for @JohannesGaessler to merge #908 and then continue this PR?

Ronsor:
That's probably best, considering the changes needed for the tests.

JohannesGaessler:
I extended the code in `test-backend-ops` to enable checking gradients from backpropagation against numerically calculated gradients. New tests for gradients should be implemented there if possible (the only thing that currently doesn't work is support for FP16). In principle all that should be necessary is to add `ggml_set_param` to the existing tests (though tuning the parameters in such a way that you get good numerical precision for the reference values can be tricky).

Ronsor:
Perfect. I plan to finish this PR this weekend.

PR #924, opened 2024-08-16
yolo : add backend support

rgerganov:


ggerganov (approved):
Metal still needs the `POOL_2D` op, but we can do that later

PR #928, opened 2024-08-24
ggml: remove bad assert

JohannesGaessler:
In https://github.com/ggerganov/ggml/pull/908 I added an assert that checks whether all dimensions of a new tensor are > 0. However, since the ggml API has functions like `ggml_is_empty` that check for zero-sized tensor dimensions this to me implies that it is actually intended for users to be able to create them (though they are an edge case that seems to not be properly handled by large parts of the code). This PR removes the assert again.


slaren (approved):
Zero-size tensors are used in llama.cpp to skip the output layer when the logits are not needed.

PR #929, opened 2024-08-25
feat: refactor cross entropy, add CUDA, fix grad test

JohannesGaessler:
I'm currently working on enabling MNIST training for backends other than CPU. As part of that I'm adding CUDA support for cross entropy loss which I'm spinning out into a separate PR since I think it will make reviewing easier. This PR makes the following changes:

* Refactor the cross entropy CPU code to optimize out the logarithm in $\log(\exp(\mathrm{logit}_i - \mathrm{max}))$. This also removes the need for the workaround where the softmax is scaled from $[0,1]$ to $[\epsilon, 1]$, simplifying the code.
* Increase `eps` for the test in `tests/test-grad0` to reduce the impact of machine precision on the numerical gradient calculation. Increase the range of the logits to ensure that cross entropy is sufficiently linear on the scale of `eps`. On master cross entropy is failing for 123/1000 iterations, with the tuned parameters it's failing for 0/1000 iterations (both for the code on master and the code in this PR).
* Add a CUDA implementation for cross entropy loss. Expose `sum_rows_f32_cuda` so that the code can be re-used to combine the partial cross entropy results into a scalar value.

### DIFF
--- a/src/ggml-cuda/cross-entropy-loss.cu
+++ b/src/ggml-cuda/cross-entropy-loss.cu
@@ -1,0 +1,106 @@
+#include "common.cuh"
+#include "cross-entropy-loss.cuh"
+#include "sumrows.cuh"
+
+#include <cmath>
+#include <cstdint>
+
+static __global__ void cross_entropy_loss_f32(const float * logits, const float * labels, float * dst, const int nclasses, const int k) {
+    const int warp_id = threadIdx.x / WARP_SIZE;
+    const int lane_id = threadIdx.x % WARP_SIZE;
+    const int i0 = blockDim.x*blockIdx.x + warp_id*WARP_SIZE;
+
+    const int ne_tmp = WARP_SIZE*nclasses;
+
+    extern __shared__ float tmp_all[];
+    float * tmp_logits = tmp_all + (2*warp_id + 0)*ne_tmp;
+    float * tmp_labels = tmp_all + (2*warp_id + 1)*ne_tmp;
+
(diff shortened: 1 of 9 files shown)


tests/test-backend-ops.cpp
```diff
@@ -1730,6 +1730,27 @@ struct test_flash_attn_ext : public test_case {
     }
 };
 
+// GGML_OP_CROSS_ENTROPY_LOSS
+struct test_cross_entropy_loss : public test_case {
+    const ggml_type type;
+    const std::array<int64_t, 4> ne;
+
+    std::string vars() override {
+        return VARS_TO_STR1(ne);
```
slaren, line 1739: This should also include the parameter `type`.

ggerganov (approved):

src/ggml-cuda/cross-entropy-loss.cu
```diff
@@ -0,0 +1,108 @@
...
+    GGML_ASSERT( dst->type == GGML_TYPE_F32);
+
+    GGML_ASSERT(ggml_is_contiguous(src0));
+    GGML_ASSERT(ggml_is_contiguous(src1));
+    GGML_ASSERT(ggml_is_contiguous(dst));
+
+    const int64_t ne00  = src0->ne[0];
+    const int64_t nrows = ggml_nrows(src0);
+
+    const int64_t ne01_padded = GGML_PAD(nrows, WARP_SIZE);
```
slaren, line 89: This variable is unused, and generates a warning.

slaren (approved):

PR #932, opened 2024-08-26
tests: add gradient tests for all backends

JohannesGaessler:
My goal with this PR is to add tests that check gradients for the backwards pass in a generic way for all backends. My current state is that I can successfully calculate and check gradients for CUDA as long as all ops are FP32 and supported. This PR is nowhere near usable but I'm opening it anyways to discuss the best way to implement the tests. Right now I'm doing the tests via modification of `tests/test-grad0` (and some hacks) but I think long-term it would make more sense to re-use the code in `tests/test-backend-ops`. From my perspective the easiest way to do this would be extend the existing code constructs in `tests/test-backend-ops` with a new mode that checks gradients numerically. @slaren your feedback would be appreciated.

### DIFF
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -1,3 +1,20 @@
+// This file defines tests for various GGML ops and backends.
+// For the forward pass it asserts that the results of multiple backends computing the same GGML ops are consistent.
+// For the backwards pass it asserts that the gradients from backpropagation are consistent
+// with the gradients obtained via the method of finite differences ("grad" mode, this is optional).
+// It is also possible to check the performance ("perf" mode).
+//
+// this file has three sections: Section 1 does general setup, section 2 defines the GGML ops to be tested,
+// and section 3 defines which tests to run.
+// Quick start for adding a new GGML op: Go to section 2 and create a struct that inherits from test_case,
+// then go to section 3 and add an instantiation of your struct.
+
+
+// ##############################
+// ## Section 1: General Setup ##
+// ##############################
+
+
 #include <ggml.h>
 #include <ggml-alloc.h>
 #include <ggml-backend.h>
@@ -216,7 +233,10 @@ static double nmse(const float * a, const float * b, size_t n) {
 // maximum absolute asymmetry between a and b
 // asymmetry: (a - b) / (a + b)
 // This is more stable than relative error if one of the values fluctuates towards zero.
-static double mean_abs_asymm(const float * a, const float * b, const size_t n, const std::vector<float> expected_vals) {
+// n: number of values to compare.
+// expected_vals: optional vector of expected values for a. If expected_vals is not empty, filter out all comparisons where
+//     a does not match any of the expected values. Needed for noncontinuous gradients where the numerical calculation can fail.
+static double mean_abs_asymm(const float * a, const float * b, const size_t n, const std::vector<float> & expected_vals) {
     double sum = 0.0f;
 
     size_t nvalid = 0;
@@ -710,7 +730,7 @@ struct test_case {
         const std::vector<float> expect = grad_expect();
 
         ggml_init_params params = {
-            /* .mem_size = */ ggml_tensor_overhead()*128 + 2*ggml_graph_overhead(),
+            /* .mem_size = */ ggml_tensor_overhead()*128 + 2*ggml_graph_overhead_custom(GGML_DEFAULT_GRAPH_SIZE, true),
             /* .mem_base = */ NULL,
             /* .no_alloc = */ true,
         };
@@ -734,6 +754,11 @@ struct test_case {
         if (out->grad == nullptr) {
             printf("backwards pass not supported \n");
             ggml_free(ctx);
+            return true;
+        }
+        if (out->type != GGML_TYPE_F32) {
+            ggml_free(ctx);
+            printf("not supported [%s->type != FP32]\n", out->name);
             return true;
         }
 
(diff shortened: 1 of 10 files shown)


slaren:
I am not very familiar with the `test-grad0` (or any of the training code), but I think this could be very well integrated into `test-backend-ops`. The `eval` mode could also be extended to optionally generate and compare backwards graphs, to compare the result with the CPU backend, though I don't know if that would add anything of value over just checking the gradients by computing them numerically.

JohannesGaessler:
I pushed an extension to `test-backend-ops` that adds a new mode for numerically checking the gradients calculated by the backwards pass. Some comments:

* I am using the mean absolute asymmetry to judge whether the results from the backwards pass are correct. The asymmetry of $a$ and $b$ is defined as $(a - b) / (a + b)$. It is similar to the mean absolute relative error but more stable if the reference value goes towards zero. The threshold for most tests is $10^{-4}$ or $10^{-3}$ if the gradient calculation is more difficult.
* For some of the more difficult operations 5 instead of 3 points are used to estimate the gradient.
* I reduced the input sizes for most operations where I got the impression that it was selected arbitrarily. The primary reason is simply that the numerical calculation of gradients is very slow. I am also skipping test cases where more than 10000 would be checked. My secondary reason for reducing the input sizes is that for non-scalar outputs the results are combined via summing them up and the fewer input values there are to this sum the more precisely the numerical gradients can be calculated.
* I added missing `GGML_ASSERT` to unsupported ops that would fail once the backwards pass is invoked. I did this for consistency but I think it's kind of a bad solution. I think long-term it would be good to at least have a CPU implementation for the backwards pass of all ops so that the compute graph can be safely constructed and simply checked with `ggml_backend_supports_op` (right now there is additional logic in the test code to prevent crashes from the failed asserts).
* I added CUDA support for `SUM`, `ADD1`, and `NEG`.

tests/test-backend-ops.cpp
```diff
@@ -643,6 +704,223 @@ struct test_case {
...
+
+
+        if (!ggml_is_scalar(out)) {
+            out = ggml_sum(ctx, out);
+            ggml_set_name(out, "sum_of_out");
+        }
+
+        ggml_build_forward_expand(gf, out);
+        ggml_graph_cpy(gf, gb);
+        ggml_build_backward_expand(ctx, gf, gb, true); // TODO why can the results sometimes be wrong with keep == false?
```
ggerganov, line 780: Unless we are computing second-order gradient (i.e. backward of backward) the `keep` should be set to `false` (though it should work with `keep == true` as well). In this case, if the results are sometimes wrong when `keep == false`, this would mean there is likely a bug somewhere.

The computation of backward graphs assumes that the initial `grad` tensors are initialized with zeros. This was the purpose of `ggml_graph_reset()`, though today this function only makes sense for the CPU backend.

How do I reproduce this issue?

JohannesGaessler: On the last commit I also no longer have issues with `keep == false`. The problem I was seeing was likely caused by https://github.com/ggerganov/ggml/pull/943 and I misinterpreted the cause.

ggerganov (approved):
Some of the gradient tests are failing occasionally (SIN, COS, ROPE) - I guess the error threshold might have to be adjusted, but we can do that as we go.

Might want to wait for @slaren review as well before merging

JohannesGaessler:
The failure rate of `SIN` and `COS` was ~12%. With the adjusted parameters the failure rate for a threshold of $10^{-4}$ was ~4%. For the increased threshold of $10^{-3}$ the failure rate is ~0.1%.

tests/test-backend-ops.cpp
```diff
@@ -701,11 +1014,22 @@ struct test_get_rows : public test_case {
...
         ggml_tensor * rows = ggml_new_tensor_2d(ctx, GGML_TYPE_I32, r, b);
+        ggml_set_name(rows, "rows");
         if (v) {
             rows = ggml_view_2d(ctx, rows, r/2, b, rows->nb[1], 0);
+            ggml_set_name(rows, "view_of_rows");
         }
+
+        if (ggml_is_matrix(in) && ggml_is_vector(rows)) {
+            ggml_set_param(ctx, in);
+        }
```
slaren, line 1028: Why?

tests/test-backend-ops.cpp
```diff
@@ -866,21 +1266,77 @@ struct test_bin_bcast : public test_case {
...
         ggml_tensor * a = ggml_new_tensor_4d(ctx, type, ne[0]*nr[0], ne[1]*nr[1], ne[2]*nr[2], ne[3]*nr[3]);
+        ggml_set_name(a, "a");
+
         ggml_tensor * b = ggml_new_tensor(ctx, type, 4, ne.data());
+        ggml_set_name(b, "b");
+
+        if (op == ggml_add || ggml_are_same_shape(a, b)) {
+            ggml_set_param(ctx, a);
+            ggml_set_param(ctx, b);
+        }
```
slaren, line 1277: Why?

slaren (commented):
There are a lot more cases like the last two. If every created tensor needs to be a param, do it automatically in the `ggml_new_tensor` functions in parent class. If the reason some are disabled is because ggml does not support backwards pass, then extract that to a function that makes that clear, and disable it by default.
It is very important that `test-backend-ops` remains very simple to add new test cases for, and nobody is going to understand if they need to add calls to `ggml_set_param` for their test or not.

JohannesGaessler:
>There are a lot more cases like the last two. If every created tensor needs to be a param, do it automatically in the ggml_new_tensor functions in parent class. If the reason some are disabled is because ggml does not support backwards pass, then extract that to a function that makes that clear, and disable it by default.

There are some tricky edge cases like `CROSS_ENTROPY_LOSS` where the backwards pass calculates gradients only for the logits but not the labels even though a variation of the labels leads to a change in the output. Would something like this be okay:

* Add a function like `should_be_param(const struct ggml_tensor * t)` to determine which tensors should be made parameters (and that defaults to false).
* Extend `ggml_new_tensor` with an optional argument for the tensor name that can be used to resolve edge cases in `should_be_param` when it's not possible to do so by e.g. type.

slaren:
I think it would be enough to add some comments explaining the cases where `ggml_set_param` is omitted, and remove the commented ones from the operators that do not support a backwards pass.

tests/test-backend-ops.cpp
```diff
@@ -2649,6 +3476,8 @@ int main(int argc, char ** argv) {
             mode = MODE_TEST;
         } else if (strcmp(argv[i], "perf") == 0) {
             mode = MODE_PERF;
+        } else if (strcmp(argv[i], "grad") == 0) {
+            mode = MODE_GRAD;
```
slaren, line 3480: The `usage` function should be updated to mention this mode.

tests/test-backend-ops.cpp
```diff
@@ -2606,6 +3421,18 @@ static bool test_backend(ggml_backend_t backend, test_mode mode, const char * op
...
+            if (test->eval_grad(backend, op_name)) {
+                n_ok++;
+            }
+        }
+        printf("  %zu/%zu tests passed\n", n_ok, test_cases.size());
+
+        return n_ok == test_cases.size();
+    }
+
     // run tests
```
slaren, line 3436: This comment should be moved before the grad test.

JohannesGaessler:
I added a small bit of documentation to the top of the file to explain the intended usage as well as a heavily commented example struct for a new op that covers the bare minimum needed for a new test. Going through the code I noticed that it's possible to avoid explicit logic for `ggml_set_param` for at least part of the ops so I did that where possible.

PR #937, opened 2024-08-26
SYCL: do not skip CPU if explicitly selected

smeso:
It can be useful to be able to use SYCL on CPU
e.g. to run tests on a machine without a GPU,
like GH's free runners.

With this commit a user can enable the use
of the CPU with the following env variable:

export ONEAPI_DEVICE_SELECTOR="opencl:cpu"


slaren:
This may get more attention from the SYCL developers if the PR is opened in the llama.cpp repository instead.

smeso:
Oh, thanks for the heads up!
Then I'll close this PR and open a similar one in llama.cpp.

PR #946, opened 2024-09-05
vulkan: correctly report support for OP_CONT

smeso:
test-backend-ops fails because ggml_cont aborts
when invoked passing an unsupported type.

This commit makes ggml_cont tests pass


slaren:
The changes look good to me. @0cc4m Can you take a look at this and #947 #948?

0cc4m (approved):

0cc4m:
@slaren Thank you, I checked all of them and they're good.

PR #949, opened 2024-09-05
ggml/examples: add backend support for numerical optimization

JohannesGaessler:
The ultimate goal of this PR is to add backend support for numerical optimization, namely Adam and L-BFGS. As of right now the corresponding computations are done by a single thread outside any of the GGML graphs. As a consequence only a single thread is used and only the CPU backend is compatible. I think the correct way to remedy this is to make the optimizers part of the GGML compute graphs. This also fixes some allocation issues where the optimization code allocates extra tensors to hold persistent extra data for the optimizers. 

As of right now this PR contains my WIP version that only supports stochastic gradient descent and the CPU backend. The training is ~3x faster than on master (but the overall rate of convergence is worse than fully featured Adam).

The overall design that I envision is that the optimizer is specified when creating the backwards graph. If no optimizer is specified, calculate the gradients without touching the weights. If an optimizer is specified, apply it to all parameters after the gradients have been calculated by adding an extra GGML op on top (could probably be optimized to overwrite gradients that are no longer needed). During backwards graph creation also specify any extra tensors needed for the optimizer so they can be correctly allocated for all backends. Functions like `ggml_opt` would then mainly be calling the backwards graph in a loop and check convergence. One potential issue is that the convergence logic would require calls to `ggml_backend_tensor_get` which would make `ggml.c` depend on `ggml_backend.c` (which it currently does not). If that is a problem the optimization code could maybe be moved to a new file like `ggml-algo.c`.

If there are issues with my design please let me know early.

### DIFF
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -264,8 +264,8 @@ mnist_model mnist_model_init_from_file(const std::string & fname, const std::str
 
         model.conv1_bias = ggml_get_tensor(model.ctx_weight, "conv1.bias");
         GGML_ASSERT(model.conv1_bias->type == GGML_TYPE_F32);
-        GGML_ASSERT(model.conv1_bias->ne[0] == MNIST_HW);
-        GGML_ASSERT(model.conv1_bias->ne[1] == MNIST_HW);
+        GGML_ASSERT(model.conv1_bias->ne[0] == 1);
+        GGML_ASSERT(model.conv1_bias->ne[1] == 1);
         GGML_ASSERT(model.conv1_bias->ne[2] == MNIST_CNN_NCB);
         GGML_ASSERT(model.conv1_bias->ne[3] == 1);
 
@@ -278,8 +278,8 @@ mnist_model mnist_model_init_from_file(const std::string & fname, const std::str
 
         model.conv2_bias = ggml_get_tensor(model.ctx_weight, "conv2.bias");
         GGML_ASSERT(model.conv2_bias->type == GGML_TYPE_F32);
-        GGML_ASSERT(model.conv2_bias->ne[0] == MNIST_HW/2);
-        GGML_ASSERT(model.conv2_bias->ne[1] == MNIST_HW/2);
+        GGML_ASSERT(model.conv2_bias->ne[0] == 1);
+        GGML_ASSERT(model.conv2_bias->ne[1] == 1);
         GGML_ASSERT(model.conv2_bias->ne[2] == MNIST_CNN_NCB*2);
         GGML_ASSERT(model.conv2_bias->ne[3] == 1);
 
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -556,15 +556,6 @@ void mnist_model_train(mnist_model & model, const float * images, const float * 
                 ggml_backend_graph_compute(model.backend, gb_opt);
                 ggml_graph_reset(gb_grad); // Set gradients to zero, do not reset optimizer.
             }
-            for (int j = 0; j < gb_grad->n_nodes; ++j) {
-                struct ggml_tensor * node = gb_grad->nodes[j];
-
-                if (node->op != GGML_OP_OPT_STEP_ADAM) {
-                    continue;
-                }
-
-                node->op_params[0]++;
-            }
 
             ggml_backend_tensor_get(model.loss,   &loss,         0, ggml_nbytes(model.loss));
             ggml_backend_tensor_get(model.logits, logits.data(), 0, ggml_nbytes(model.logits));
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -1,3 +1,5 @@
+#include "ggml-alloc.h"
+#include "ggml-backend.h"
 #include "ggml.h"
 
 #include "mnist-common.h"
@@ -158,16 +160,20 @@ mnist_model mnist_model_init_from_file(const std::string & fname) {
     mnist_model model;
     fprintf(stderr, "%s: loading model weights from '%s'\n", __func__, fname.c_str());
 
-    struct gguf_init_params params = {
-        /*.no_alloc   =*/ false,
-        /*.ctx        =*/ &model.ctx_weight,
-    };
-    gguf_context * ctx = gguf_init_from_file(fname.c_str(), params);
-    if (!ctx) {
-        fprintf(stderr, "%s: gguf_init_from_file() failed\n", __func__);
-        exit(1);
+    struct gguf_context * ctx_be; // be == backend
+
+    {
+        struct gguf_init_params params = {
+            /*.no_alloc   =*/ true,
+            /*.ctx        =*/ &model.ctx_weight,
+        };
+        ctx_be = gguf_init_from_file(fname.c_str(), params);
+        if (!ctx_be) {
+            fprintf(stderr, "%s: gguf_init_from_file() failed\n", __func__);
+            exit(1);
+        }
     }
-    model.arch = gguf_get_val_str(ctx, gguf_find_key(ctx, "general.architecture"));
+    model.arch = gguf_get_val_str(ctx_be, gguf_find_key(ctx_be, "general.architecture"));
     fprintf(stderr, "%s: model arch is %s\n", __func__, model.arch.c_str());
 
     if (model.arch == "mnist-fc") {
@@ -239,6 +245,43 @@ mnist_model mnist_model_init_from_file(const std::string & fname) {
     } else {
         fprintf(stderr, "%s: unknown model arch: %s\n", __func__, model.arch.c_str());
     }
+    model.buf_weightt = ggml_backend_alloc_ctx_tensors(model.ctx_weight, model.backend);
+
+    void * buf_tmp = malloc(model.size_weight);
+    struct ggml_context * ctx_ggml_tmp;
+    {
+        struct ggml_init_params params = {
+            /*.mem_size   =*/ model.size_weight,
+            /*.mem_buffer =*/ buf_tmp,
+            /*.no_alloc   =*/ false,
+        };
+        ctx_ggml_tmp = ggml_init(params);
+    }
+    struct gguf_context * ctx_gguf_tmp;
+    {
+        struct gguf_init_params params = {
+            /*.no_alloc   =*/ false,
+            /*.ctx        =*/ &ctx_ggml_tmp,
+        };
+        ctx_gguf_tmp = gguf_init_from_file(fname.c_str(), params);
+        if (!ctx_gguf_tmp) {
+            fprintf(stderr, "%s: gguf_init_from_file() failed\n", __func__);
+            exit(1);
+        }
+    }
+    for (const std::string & s : {"fc1.weight", "fc1.bias", "fc2.weight", "fc2.bias"}) {
+        const struct ggml_tensor * src = ggml_get_tensor(ctx_ggml_tmp,     s.c_str());
+        struct       ggml_tensor * dst = ggml_get_tensor(model.ctx_weight, s.c_str());
+        GGML_ASSERT(ggml_nbytes(src) == ggml_nbytes(dst));
+        ggml_backend_tensor_set(dst, src->data, 0, ggml_nbytes(dst));
+    }
+
+    gguf_free(ctx_gguf_tmp);
+    ggml_free(ctx_ggml_tmp);
+    free(buf_tmp);
+
+    gguf_free(ctx_be);
+
     fprintf(stderr, "%s: successfully loaded weights from %s\n", __func__, fname.c_str());
     return model;
 }
@@ -408,6 +451,8 @@ void mnist_model_build(mnist_model & model, const int nbatch) {
     GGML_ASSERT(model.loss->ne[1] == 1);
     GGML_ASSERT(model.loss->ne[2] == 1);
     GGML_ASSERT(model.loss->ne[3] == 1);
+
+    model.buf_compute = ggml_backend_alloc_ctx_tensors(model.ctx_compute, model.backend);
 }
 
 mnist_eval_result mnist_model_eval(const mnist_model & model, const float * images, const float * labels, const int nex, const int nthreads) {
@@ -419,17 +464,27 @@ mnist_eval_result mnist_model_eval(const mnist_model & model, const float * imag
     {
         const int64_t t_start_us = ggml_time_us();
 
+        float loss;
+        std::vector<float> logits(model.nbatch*MNIST_NCLASSES);
+
+        GGML_ASSERT(sizeof(loss)  == ggml_nbytes(model.loss));
+        GGML_ASSERT(logits.size() == ggml_nelements(model.logits));
+
         GGML_ASSERT(nex % model.nbatch == 0);
         for (int iex0 = 0; iex0 < nex; iex0 += model.nbatch) {
-            memcpy(model.images->data, images + iex0*MNIST_NINPUT,   ggml_nbytes(model.images));
-            memcpy(model.labels->data, labels + iex0*MNIST_NCLASSES, ggml_nbytes(model.labels));
-            ggml_graph_compute_with_ctx(model.ctx_compute, gf, nthreads);
+            ggml_backend_tensor_set(model.images, images + iex0*MNIST_NINPUT,   0, ggml_nbytes(model.images));
+            ggml_backend_tensor_set(model.labels, labels + iex0*MNIST_NCLASSES, 0, ggml_nbytes(model.labels));
 
-            result.loss.push_back(*ggml_get_data_f32(model.loss));
+            ggml_backend_graph_compute(model.backend, gf);
+
+            ggml_backend_tensor_get(model.loss,   &loss,         0, ggml_nbytes(model.loss));
+            ggml_backend_tensor_get(model.logits, logits.data(), 0, ggml_nbytes(model.logits));
+
+            result.loss.push_back(loss);
 
             for (int iexb = 0; iexb < model.nbatch; ++iexb) {
-                const float * logits_data = ggml_get_data_f32(model.logits) + iexb*MNIST_NCLASSES;
-                result.pred.push_back(std::max_element(logits_data, logits_data + MNIST_NCLASSES) - logits_data);
+                const float * logits_iexb = logits.data() + iexb*MNIST_NCLASSES;
+                result.pred.push_back(std::max_element(logits_iexb, logits_iexb + MNIST_NCLASSES) - logits_iexb);
             }
         }
 
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -530,13 +530,16 @@ mnist_eval_result mnist_model_eval(mnist_model & model, const float * images, co
 void mnist_model_train(mnist_model & model, const float * images, const float * labels, const int nex, const int nepoch, const float val_split) {
     const int64_t t_start_us = ggml_time_us();
 
+    // gf == graph forward, forward pass only.
     struct ggml_cgraph * gf = ggml_new_graph_custom(model.ctx_compute, GGML_DEFAULT_GRAPH_SIZE, /*grads =*/ true); // Forward pass.
     ggml_build_forward_expand(gf, model.loss);
 
-    struct ggml_cgraph * gb_grad = ggml_graph_dup(model.ctx_compute, gf); // Backward pass, gradients.
+    // gb_grad == graph backward gradients, forward pass, then backward pass to calculate gradients.
+    struct ggml_cgraph * gb_grad = ggml_graph_dup(model.ctx_compute, gf);
     ggml_build_backward_expand(model.ctx_compute, gf, gb_grad, /*accumulate =*/ true, false);
 
-    struct ggml_cgraph * gb_opt = ggml_graph_dup(model.ctx_compute, gf); // Backward pass, gradients + optimizer.
+    // gb_opt == graph backward optimize, forward pass, then backward pass to calculate gradients, then optimizer step.
+    struct ggml_cgraph * gb_opt = ggml_graph_dup(model.ctx_compute, gb_grad);
     ggml_build_opt_adamw(model.ctx_compute, gf, gb_opt, 1e-3f, 0.9f, 0.999f, 1e-8f, 0.0f);
 
     model.buf_compute = ggml_backend_alloc_ctx_tensors(model.ctx_compute, model.backend);
@@ -556,8 +559,6 @@ void mnist_model_train(mnist_model & model, const float * images, const float * 
         for (; iex0 < iex_split; iex0 += model.nbatch_physical) {
             ggml_backend_tensor_set(model.images, images + iex0*MNIST_NINPUT,   0, ggml_nbytes(model.images));
             ggml_backend_tensor_set(model.labels, labels + iex0*MNIST_NCLASSES, 0, ggml_nbytes(model.labels));
-
-            ggml_backend_graph_compute(model.backend, gf); // Always compute forward pass.
 
             // With a period of nbatch_logical/nbatch_physical iterations:
             if ((iex0 + model.nbatch_physical) % model.nbatch_logical != 0) {
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -527,7 +527,7 @@ mnist_eval_result mnist_model_eval(mnist_model & model, const float * images, co
     return result;
 }
 
-void mnist_model_train(mnist_model & model, const float * images, const float * labels, const int nex) {
+void mnist_model_train(mnist_model & model, const float * images, const float * labels, const int nex, const int nepoch, const float val_split) {
     const int64_t t_start_us = ggml_time_us();
 
     struct ggml_cgraph * gf = ggml_new_graph_custom(model.ctx_compute, 16384, true); // Forward pass.
@@ -542,15 +542,18 @@ void mnist_model_train(mnist_model & model, const float * images, const float * 
     model.buf_compute = ggml_backend_alloc_ctx_tensors(model.ctx_compute, model.backend);
     ggml_graph_reset(gb_opt); // Set gradients to zero, reset optimizer.
 
-    for (int epoch = 0; epoch < 30; ++epoch) {
+    const int iex_split = ((int)((1.0f - val_split)*nex) / model.nbatch_logical) * model.nbatch_logical;
+
+    for (int epoch = 0; epoch < nepoch; ++epoch) {
         fprintf(stderr, "%s: epoch %02d start...", __func__, epoch);
         const int64_t t_start_us = ggml_time_us();
 
         float loss;
         std::vector<float> logits(model.nbatch_physical*MNIST_NCLASSES);
+        int iex0 = 0;
 
-        mnist_eval_result result;
-        for (int iex0 = 0; iex0 < nex; iex0 += model.nbatch_physical) {
+        mnist_eval_result result_train;
+        for (; iex0 < iex_split; iex0 += model.nbatch_physical) {
             ggml_backend_tensor_set(model.images, images + iex0*MNIST_NINPUT,   0, ggml_nbytes(model.images));
             ggml_backend_tensor_set(model.labels, labels + iex0*MNIST_NCLASSES, 0, ggml_nbytes(model.labels));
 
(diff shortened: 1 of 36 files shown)


JohannesGaessler:
>One potential issue is that the convergence logic would require calls to ggml_backend_tensor_get which would make ggml.c depend on ggml_backend.c (which it currently does not).

Actually, a much bigger issue is that for `ggml_backend_graph_compute` a pointer to a backend is needed.

slaren:
I don't think that's a problem. ggml-backend was designed to be not require many changes to the core ggml code, but since then I think it has become the standard way to use ggml, and it doesn't make much sense to maintain the subset of the API that only works with the CPU backend. We should move all the CPU backend code to a separate file, and make all the core ggml functions explicitly compatible with ggml-backend.

The design looks good to me.

JohannesGaessler:
I forgot: the current code also has an extension to the GGML backend interface with `memset_tensor` in order to clear specific tensors (since right now I think the only way to do it would be to allocate zeroed memory and invoke `set_tensor`).

JohannesGaessler:
The Adam optimizer needs to know the current iteration since it does a warmup. I'm currently passing this information via `ggml_tensor.op_params` but the downside with this approach is that the information regarding the iteration is duplicated across all tensors. But at the same time I don't think it would be a good idea to add a global state to the forward pass when right now all relevant information is encapsulated in `ggml_tensor`.

JohannesGaessler:
I pushed a working prototype for CUDA MNIST training/evaluation (fully connected only). Compared to PyTorch the training on my RTX 3090 is ~45x faster (1.25s vs. 56.58s) but with such a small model you're basically just measuring overhead. The CUDA evaluation is actually slower than the CPU evaluation, presumably because the model is too small to make GPU acceleration worthwhile given the additional overhead.

One issue that I still have is how to handle the combination of GGUF+backends other than CPU. Right now I'm allocating a temporary context that just stores the data in RAM but it feels kind of clunky. Is there a better way to do this?

slaren:
> One issue that I still have is how to handle the combination of GGUF+backends other than CPU. Right now I'm allocating a temporary context that just stores the data in RAM but it feels kind of clunky. Is there a better way to do this?

Check the way the [magika](https://github.com/ggerganov/ggml/blob/master/examples/magika/main.cpp) example does this: make a `no_alloc` gguf context, call `ggml_backend_alloc_ctx_tensors`, then load the data from file using `gguf_get_tensor_offset`.

src/ggml.c
```diff
@@ -18651,6 +18771,16 @@ void ggml_build_backward_expand(struct ggml_context * ctx, struct ggml_cgraph *
...
+    for (int i = 0; i < gf->n_nodes; i++) {
+        struct ggml_tensor * node = gf->nodes[i];
+
+        if (node->flags & GGML_TENSOR_FLAG_PARAM) {
+            GGML_PRINT_DEBUG("%s: found root node %p\n", __func__, (void *) node);
+            struct ggml_tensor * opt_step = ggml_opt_step_adam(ctx, node, 1.0f, 0.001f, 0.9f, 0.999f, 1e-8f);
+            ggml_build_forward_expand(gb, opt_step);
+        }
+    }
+
```
ggerganov, line 18783: > The overall design that I envision is that the optimizer is specified when creating the backwards graph. If no optimizer is specified, calculate the gradients without touching the weights. If an optimizer is specified, apply it to all parameters after the gradients have been calculated by adding an extra GGML op on top (could probably be optimized to overwrite gradients that are no longer needed).

Purely from API PoV, it might be better to have separate calls that expand the graph with an optimizer computation (e.g. `ggml_build_opt` or something similar) that would optionally be called after `ggml_build_backward`.

JohannesGaessler: Do you have an opinion on what to do with the current `ggml_opt` API? If we keep it the addition of tensors for optimization could be done in `ggml_opt_init`.

ggerganov: I'm not really sure how good is the design of the existing `ggml_opt` API. I think we can afford to change it significantly, since it is not really adopted by other projects. We can even implement a new API in parallel and when we know which one is better - remove the other.

`ggml_opt_init` adding the optimization graph sounds OK to me. The question is if there would be use cases where you would want to create an optimizer, but not immediately "apply" it to a graph. Maybe you might want to apply the same optimizer to multiple graphs? If there is some case like this, then a separate `ggml_build_opt`-like step might make sense.

include/ggml-backend.h
```diff
@@ -234,6 +234,7 @@ extern "C" {
     GGML_API void ggml_backend_tensor_alloc(ggml_backend_buffer_t buffer, struct ggml_tensor * tensor, void * addr);
     GGML_API void ggml_backend_view_init(struct ggml_tensor * tensor);
 
+    GGML_API bool ggml_backend_load_from_gguf(const char * fname, struct ggml_context * ctx_ggml, struct gguf_context * ctx_gguf);
```
ggerganov, line 237: Is this temporary? Seems like it does not belong to `ggml-backend`. More like a utility function in user code.

JohannesGaessler: What API should we use long-term for loading data from GGUF? I was thinking that since the pattern for tensors created in program code is initialization -> backend allocation -> data setting this would be the equivalent way to do it for GGUF.

slaren: There was some talk in llama.cpp about moving some of the loading code to ggml, including mmap support, so that other ggml applications can benefit from it. I am not sure how that API should look, though. It may be good to add this as a first step, but most likely it will need a different API to be able to achieve all the goals.

src/ggml-backend.c
```diff
@@ -2244,3 +2267,48 @@ bool ggml_backend_compare_graph_backend(ggml_backend_t backend1, ggml_backend_t
...
+    void * buf = malloc(buf_size);
+
+    const int n_tensors = gguf_get_n_tensors(ctx_gguf);
+    for (int i = 0; i < n_tensors; i++) {
+        const char * name = gguf_get_tensor_name(ctx_gguf, i);
+
+        struct ggml_tensor * tensor = ggml_get_tensor(ctx_ggml, name);
+        if (!tensor) {
+            return false;
+        }
```
slaren, line 2287: It is probably better to just skip the tensors that are not present in the `ggml_context`, because that's a normal case when using multiple backends and each tensor may be stored in a different buffer type and different `ggml_context`.
Also note that as it is, this will leak the buffer and file handle.

JohannesGaessler:
I think for `ggml_opt_step_adam` the parameter `sched` is not needed. It's essentially a way to adjust the learning rate via a callback but I think something like this should be done one level further up via rather than in the tensors.

src/ggml.c
JohannesGaessler: I think it's definitely better to have a separate call for adding the optimizer. That way gradient accumulation can be implemented relatively easily by defining one graph that calculates just the gradients and one that also invokes the optimizer.

JohannesGaessler:
Alright, I think this PR now has the minimum (in terms of functionality) where I think merging it would make sense. What is still missing is an update to the documentation and more general CUDA support for `GGML_OP_REPEAT_BACK`. Long-term I would also want a way to define datasets in GGML and a more high-level API (comparable to the current `ggml_opt` one).

src/ggml.c
```diff
@@ -19147,10 +19318,29 @@ void ggml_graph_reset(struct ggml_cgraph * cgraph) {
     GGML_ASSERT(cgraph->grads != NULL);
 
     for (int i = 0; i < cgraph->n_nodes; i++) {
+        struct ggml_tensor * node = cgraph->nodes[i];
         struct ggml_tensor * grad = cgraph->grads[i];
 
+        // initial gradients of loss should be 1, 0 otherwise
         if (grad) {
-            ggml_set_zero(grad);
+            if (node->flags & GGML_TENSOR_FLAG_LOSS) {
```
ggerganov, line 19326: Would it make sense to add a sanity check that only one node is of type `LOSS` to prevent misuse of the API? (unless there are situations where multiple `LOSS` nodes are valid)

examples/mnist/mnist-common.cpp
```diff
@@ -443,81 +527,141 @@ mnist_eval_result mnist_model_eval(const mnist_model & model, const float * imag
...
+
+            // With a period of nbatch_logical/nbatch_physical iterations:
+            if ((iex0 + model.nbatch_physical) % model.nbatch_logical != 0) {
+                // For the first nbatch_logical/nbatch_physical - 1 iterations, only calculate gradients and accumulate them:
+                ggml_backend_graph_compute(model.backend, gb_grad);
+            } else {
+                // For the last iteration, calculate gradients and also apply the optimizer:
+                ggml_backend_graph_compute(model.backend, gb_opt);
+                ggml_graph_reset(gb_grad); // Set gradients to zero, do not reset optimizer.
+            }
```
ggerganov, line 570: This would never call `ggml_backend_graph_compute(model.backend, gb_grad);` if `nbatch_logical == nbatch_physical`

src/ggml.c
JohannesGaessler: If you specify multiple loss nodes the result would be that the effective loss would be the sum of individual loss tensors. I wouldn't consider this a misuse of the API since it would be a relatively convenient way to add extra terms to the loss (e.g. l1/l2 regularization).

examples/mnist/mnist-common.cpp
JohannesGaessler: It's not necessary. `gb_opt` contains all nodes in `gb_grad` so the gradients will still be calculated prior to the optimization step; I added a comment to clarify.

examples/mnist/mnist-common.h
```diff
@@ -44,28 +52,41 @@ struct mnist_model {
...
         {
             struct ggml_init_params params = {
                 /*.mem_size   =*/ size_compute,
-                /*.mem_buffer =*/ buf_compute,
-                /*.no_alloc   =*/ false,
+                /*.mem_buffer =*/ nullptr,
+                /*.no_alloc   =*/ true,
             };
             ctx_compute = ggml_init(params);
         }
```
slaren, line 92: Shouldn't `size_compute` and `size_weight` should be reduced now? this is allocating very big buffers for what looks like just for the metadata.

JohannesGaessler: Thanks, I forgot.

ggerganov:
> As of right now this PR contains my WIP version that only supports stochastic gradient descent and the CPU backend. The training is ~3x faster than on master (but the overall rate of convergence is worse than fully featured Adam).

Is this still the case? Seems like the PR now implements Adam, but the convergence is still slower compared the `mnist-train-fc.py`. Do you know why is that?

```bash
# ggml (mnist-fc)
mnist_model_train: epoch 29 start...done, took 0.59s, train_loss=0.058282, train_acc=98.46%, val_loss=0.083196+-0.025347, train_acc=97.63+-0.28%

# py
Epoch [30/30], Step [60000/60000], Loss: 0.0086, Accuracy: 99.91%
```

JohannesGaessler:
>Is this still the case? Seems like the PR now implements Adam, but the convergence is still slower compared the mnist-train-fc.py. Do you know why is that?

To be clear, I was specifically referring to the speed of epochs, not to the speed of convergence. The speed of epochs has not significantly changed from adding momentum. However, I added an optimization to the PyTorch script that loads the data asynchronously so the PyTorch baseline to compare against has become comparatively faster. Currently there are the following relevant differences:

* PyTorch uses a physical batch size of 1000, GGML a physical batch size of 500 (to demonstrate gradient accumulation).
* GGML only uses 95% of the training data to demonstrate setting a validation split, PyTorch uses 100%.
* GGML tensor initialization is very crude, PyTorch has likely more sophisticated methods.

I didn't look at this until now but the training loss with PyTorch seems to decrease much faster per epoch than with GGML and I don't think this can be explained with the above factors; I'll investigate.

examples/mnist/mnist-common.cpp
```diff
@@ -443,81 +527,142 @@ mnist_eval_result mnist_model_eval(const mnist_model & model, const float * imag
...
+            const int64_t t_epoch_us = ggml_time_us() - t_start_us;
+            const double t_epoch_s = 1e-6*t_epoch_us;
+            fprintf(stderr, "done, took %.2lfs, train_loss=%.6lf, train_acc=%.2f%%", t_epoch_s, loss_mean, percent_correct);
+        }
+
+        if (iex_split < nex) {
+            const std::pair<double, double> loss = mnist_loss(result_val);
+            const std::pair<double, double> acc  = mnist_accuracy(result_val, labels + iex_split*MNIST_NCLASSES);
+
+            fprintf(stderr, ", val_loss=%.6lf+-%.6lf, train_acc=%.2f+-%.2f%%", loss.first, loss.second, 100.0*acc.first, 100.0*acc.second);
```
ggerganov, line 614:
```suggestion
            fprintf(stderr, ", val_loss=%.6lf+-%.6lf, val_acc=%.2f+-%.2f%%", loss.first, loss.second, 100.0*acc.first, 100.0*acc.second);
```

JohannesGaessler:
I get comparable results between PyTorch and GGML in terms of training loss when I add the following two modifications: disable dataset shuffling for PyTorch and set the GGML physical batch size to 1000. The latter is a bug since by definition the physical batch size should have no effect beyond differences in rounding error. For the dataset shuffling I would have intuitively expected that this is only relevant for generalization but it seems that it also improves the rate at which the model gets better on the training set.

JohannesGaessler:
I figured out the problem: I incorrectly assumed that the `ggml_tensor.grad` pointers would be constant for my implementation. So while the original gradients are being added to the `GGML_ADD` tensors that eventually used as gradients they are never incremented and thus remain zero. The accumulation steps prior to the last one are effectively just being discarded. I think the fix will be to do in-place additions in `ggml_compute_backward`.

ggerganov:
> I figured out the problem: I incorrectly assumed that the `ggml_tensor.grad` pointers would be constant for my implementation. So while the original gradients are being used as input for the `GGML_ADD` tensors that are eventually being used as gradients they are never incremented and thus remain zero. The accumulation steps prior to the last one are effectively just being discarded. I think the fix will be to do in-place additions in `ggml_compute_backward`.

So if I understand correctly, the following call is basically a noop atm:

https://github.com/ggerganov/ggml/blob/c1d13df2c8e75eadc7cb19e46f3352a58e6fc42f/examples/mnist/mnist-common.cpp#L563-L566

The reason is because `ggml_backend_graph_compute(model.backend, gb_opt);` ends up using the gradients from the `gb_opt` graph which so far haven't been updated. Instead we have been updating the gradients of the `gb_grad` graph.

I tried your idea, which I think is simply to set the `inplace = true` in `ggml_add_or_set`:

```diff
diff --git a/src/ggml.c b/src/ggml.c
index de61438..483a3b2 100644
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -18129,7 +18129,7 @@ static struct ggml_tensor * ggml_add_or_set(struct ggml_context * ctx, struct gg
     if (ggml_hash_contains(zero_table, a)) {
         return b;
     } else {
-        return ggml_add_impl(ctx, a, b, false);
+        return ggml_add_impl(ctx, a, b, true);
     }
 }
```

But it seems we are still missing something, as the training accuracy dropped:

```
mnist_model_train: epoch 29 start...done, took 0.58s, train_loss=0.182481, train_acc=95.08%, val_loss=0.166181+-0.027298, train_acc=96.07+-0.35%
```

JohannesGaessler:
I've pushed a WIP fix that works but has bad performance to clarify the problem. The original gradients are initialized with zero and need to be incremented with the sum tensors after each accumulation step to get correct results. Unrelated to the problem with accumulation there are also two other issues: the wrong graph was being copied for `gb_opt` and the execution of the forward graph is not needed because the backwards graphs include all of its tensors (the latter only matters for performance).

JohannesGaessler:
>I tried your idea, which I think is simply to set the inplace = true in ggml_add_or_set:

The problem is the upper branch where the tensor is in the zero table. In that case there needs to be an in-place addition instead of a replacement. But so far I have not been able to make that work so there is likely still some other issue.

JohannesGaessler:
Sorry, the supposed fix had two bugs that happened to cancel each other out.

JohannesGaessler:
I pushed a proper fix. The correct handling of gradient accumulation needs some extra bookkeeping to track the gradients of parameters and whether they should be accumulated, I added a new tensor flag for this.

JohannesGaessler:
Actually, now that I think about it it would maybe be better to do this via a hash set instead of via tensor modification since whether or not a gradient should be accumulated is a property of the compute graph rather than the gradient tensor. But by that logic the existing code in `ggml_build_backward` that is modifying the tensors is also bad.

Also: there was some inconsistent use of `ggml_cgraph.nodes[i]->grad` vs. `ggml_cgraph.grads[i]` that was causing problems.

include/ggml.h
```diff
@@ -569,10 +570,12 @@ extern "C" {
         GGML_LOG_LEVEL_DEBUG = 5
     };
 
+    // this tensor...
     enum ggml_tensor_flag {
-        GGML_TENSOR_FLAG_INPUT  = 1,
-        GGML_TENSOR_FLAG_OUTPUT = 2,
-        GGML_TENSOR_FLAG_PARAM  = 4,
+        GGML_TENSOR_FLAG_INPUT    = 1, // ...is an input for the GGML comptue graph
+        GGML_TENSOR_FLAG_OUTPUT   = 2, // ...is an output for the GGML comptue graph
```
slaren, line 576:
```suggestion
        GGML_TENSOR_FLAG_INPUT    = 1, // ...is an input for the GGML compute graph
        GGML_TENSOR_FLAG_OUTPUT   = 2, // ...is an output for the GGML compute graph
```

tests/test-backend-ops.cpp
```diff
@@ -3159,14 +3229,15 @@ static bool test_backend(ggml_backend_t backend, test_mode mode, const char * op
...
 
-
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_F32, {10, 5, 4, 3}, {1, 1, 1, 1}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_F32, {10, 5, 4, 3}, {2, 1, 1, 1}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_F32, {10, 5, 4, 3}, {1, 2, 1, 1}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_F32, {10, 5, 4, 3}, {1, 1, 2, 1}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_F32, {10, 5, 4, 3}, {1, 1, 1, 2}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_I32, {10, 5, 4, 3}, {2, 1, 1, 1}));
-    test_cases.emplace_back(new test_repeat(GGML_TYPE_I16, {10, 5, 4, 3}, {1, 1, 1, 2}));
+    for (const int64_t & ne3 : {1, 3}) { // CUDA backwards pass only supports ne3 == 1
```
slaren, line 3232:
```suggestion
    for (int ne3 : {1, 3}) { // CUDA backwards pass only supports ne3 == 1
```

tests/test-backend-ops.cpp
```diff
@@ -3476,6 +3567,9 @@ static bool test_backend(ggml_backend_t backend, test_mode mode, const char * op
     }
 
     test_cases.emplace_back(new test_cross_entropy_loss());
+    for (const float & wd : {0.0f, 1e-2f}) {
```
slaren, line 3570:
```suggestion
    for (float wd : {0.0f, 1e-2f}) {
```

tests/test-backend-ops.cpp
```diff
@@ -2666,6 +2691,51 @@ struct test_cross_entropy_loss : public test_case {
...
+    const ggml_type type;
+    const std::array<int64_t, 4> ne;
+    const float alpha;
+    const float beta1;
+    const float beta2;
+    const float eps;
+    const float wd;
+
+    std::string vars() override {
+        return VARS_TO_STR2(type, ne);
```
slaren, line 2705: All the parameters relevant to the test need to be returned here so that we know what is being tested.

src/ggml-cuda/out-prod.cu
```diff
@@ -0,0 +1,59 @@
...
+
+    const float alpha = 1.0f;
+    const float beta = 0.0f;
+
+    GGML_ASSERT(ne2 == 1);
+    GGML_ASSERT(ne3 == 1);
+    CUBLAS_CHECK(cublasSetStream(handle, stream));
+
+    const cublasOperation_t src1_cublas_op = ggml_is_transposed(src1) ? CUBLAS_OP_N : CUBLAS_OP_T;
+    const int64_t           ldb            = ggml_is_transposed(src1) ? ne11        : ne10;
```
slaren, line 52: I think `ldb = nb11/sizeof(float)` should work more generically. Otherwise it will fail silently with some types of views.

src/ggml-backend.c
```diff
@@ -246,6 +246,20 @@ GGML_CALL void ggml_backend_tensor_get(const struct ggml_tensor * tensor, void *
...
+    GGML_ASSERT(buf != NULL && "tensor buffer not set");
+    GGML_ASSERT(tensor->data != NULL && "tensor not allocated");
+    GGML_ASSERT(offset + size <= ggml_nbytes(tensor) && "tensor write out of bounds");
+
+    if (!size) {
+        return;
+    }
+
+    buf->iface.memset_tensor(buf, tensor, value, offset, size);
+}
```
ggerganov, line 261: This would make the reason for failure on other backends more clear:

```suggestion
    if (!size) {
        return;
    }
    
    GGML_ASSERT(buf->iface.memset_tensor != NULL && "memset not supported by backend buffer");

    buf->iface.memset_tensor(buf, tensor, value, offset, size);
}
```

tests/test-backend-ops.cpp
JohannesGaessler: Thanks, I forgot.

src/ggml-cuda/out-prod.cu
JohannesGaessler: I think that will result in the wrong logic. What I need here is to return a constant value for `ldb` regardless of whether or not `src1` was transposed, unless I'm missing something the value of `nb11/sizeof(float)` would change.

slaren: Yes, you are right. I was thinking about strides, but didn't take into account that cublas already expects a transposed matrix. I suggest this change instead:

```diff
diff --git a/src/ggml-cuda/out-prod.cu b/src/ggml-cuda/out-prod.cu
index ee621d3..7807474 100644
--- a/src/ggml-cuda/out-prod.cu
+++ b/src/ggml-cuda/out-prod.cu
@@ -7,25 +7,15 @@ void ggml_cuda_out_prod(ggml_backend_cuda_context & ctx, ggml_tensor * dst) {
     const ggml_tensor * src0 = dst->src[0];
     const ggml_tensor * src1 = dst->src[1];

+    GGML_TENSOR_BINARY_OP_LOCALS
+
     GGML_ASSERT(src0->type == GGML_TYPE_F32);
     GGML_ASSERT(src1->type == GGML_TYPE_F32);
     GGML_ASSERT(dst->type  == GGML_TYPE_F32);
     GGML_ASSERT(ggml_is_contiguous(src0));
     GGML_ASSERT(ggml_is_contiguous(dst));

-    const int64_t ne00 = src0->ne[0];
-    const int64_t ne01 = src0->ne[1];
-
-    const int64_t ne10 = src1->ne[0];
-    const int64_t ne11 = src1->ne[1];
-
     GGML_ASSERT(ne01 == ne11);
-
-    const int64_t ne0 = dst->ne[0];
-    const int64_t ne1 = dst->ne[1];
-    const int64_t ne2 = dst->ne[2];
-    const int64_t ne3 = dst->ne[3];
-
     GGML_ASSERT(ne0 == ne00);
     GGML_ASSERT(ne1 == ne10);

@@ -48,8 +38,8 @@ void ggml_cuda_out_prod(ggml_backend_cuda_context & ctx, ggml_tensor * dst) {
     GGML_ASSERT(ne3 == 1);
     CUBLAS_CHECK(cublasSetStream(handle, stream));

-    const cublasOperation_t src1_cublas_op = ggml_is_transposed(src1) ? CUBLAS_OP_N : CUBLAS_OP_T;
-    const int64_t           ldb            = ggml_is_transposed(src1) ? ne11        : ne10;
+    const cublasOperation_t src1_cublas_op =  ggml_is_transposed(src1) ? CUBLAS_OP_N : CUBLAS_OP_T;
+    const int64_t           ldb            = (ggml_is_transposed(src1) ?        nb10 :        nb11)/sizeof(float);
     CUBLAS_CHECK(
         cublasSgemm(handle, CUBLAS_OP_N, src1_cublas_op,
                 ne0, ne1, ne01,
diff --git a/tests/test-backend-ops.cpp b/tests/test-backend-ops.cpp
index 729d4b2..6c4de06 100644
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -1679,9 +1679,10 @@ struct test_out_prod : public test_case {
     const int64_t n;
     const int64_t k;
     const std::array<int64_t, 2> bs; // dims 3 and 4
+    const bool trans_b;

     std::string vars() override {
-        return VARS_TO_STR6(type_a, type_b, m, n, k, bs);
+        return VARS_TO_STR7(type_a, type_b, m, n, k, bs, trans_b);
     }

     double max_nmse_err() override {
@@ -1690,13 +1691,21 @@ struct test_out_prod : public test_case {

     test_out_prod(ggml_type type_a = GGML_TYPE_F32, ggml_type type_b = GGML_TYPE_F32,
             int64_t m = 32, int64_t n = 32, int64_t k = 32,
-            std::array<int64_t, 2> bs = {10, 10})
-        : type_a(type_a), type_b(type_b), m(m), n(n), k(k), bs(bs) {}
+            std::array<int64_t, 2> bs = {10, 10},
+            bool trans_b = false)
+        : type_a(type_a), type_b(type_b), m(m), n(n), k(k), bs(bs), trans_b(trans_b) {}

     ggml_tensor * build_graph(ggml_context * ctx) override {
         ggml_tensor * a = ggml_new_tensor_4d(ctx, type_a, m, k, bs[0], bs[1]);
-        ggml_tensor * b = ggml_new_tensor_4d(ctx, type_b, n, k, bs[0], bs[1]);
         ggml_set_name(a, "a");
+
+        ggml_tensor * b;
+        if (trans_b) {
+            b = ggml_new_tensor_4d(ctx, type_b, k, n, bs[0], bs[1]);
+            b = ggml_transpose(ctx, b);
+        } else {
+            b = ggml_new_tensor_4d(ctx, type_b, n, k, bs[0], bs[1]);
+        }
         ggml_set_name(b, "b");

         ggml_tensor * out = ggml_out_prod(ctx, a, b);
@@ -3432,6 +3441,7 @@ static bool test_backend(ggml_backend_t backend, test_mode mode, const char * op
             test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 1, 16, {10, 10}));

             test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 16, 16, { 1,  1}));
+            test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 16, 16, { 1,  1}, true));
             test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 16, 16, {10,  1}));
             test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 16, 16, {10,  1}));
             test_cases.emplace_back(new test_out_prod(type_a, type_b, 256, 16, 16, {10, 10}));
```

JohannesGaessler: Thanks, I also added an assert for the other stride being `sizeof(float)` since this is required by cuBLAS.

tests/test-backend-ops.cpp
```diff
@@ -837,22 +838,11 @@ struct test_case {
...
-        for (struct ggml_tensor * t = ggml_get_first_tensor(ctx); t != nullptr; t = ggml_get_next_tensor(ctx, t)) {
-            if (!t->grad) {
-                continue;
-            }
 
-            std::vector<float> tmp(ggml_nelements(t->grad));
-            ggml_backend_tensor_set(t->grad, tmp.data(), 0, ggml_nbytes(t->grad));
-        }
+        initialize_tensors(ctx); // Randomizes all tensors (including gradients).
+        ggml_graph_reset(gb);    // Sets gradients to 1 if loss, 0 otherwise.
```
ggerganov, line 843: The comment about `ggml_graph_reset` that it sets 1 to loss and 0 otherwise should be added in `ggml.h`

ggerganov (commented):
Going back to an earlier comment by @slaren:

> Something to consider is that to support multiple GPUs and fallback to CPU for unimplemented ops in the backends, it is necessary to use ggml_backend_sched.

Should we attempt to do this within this PR or after we merge the existing changes? Do we see any obstacles to achieve this?

Overall, I think the changes are quite good. I'm not familiar with other training codebases, so not sure if we are missing something obvious from functionality perspective.

JohannesGaessler:
My priorities: while `ggml_backend_sched` would be nice to have I think it's more important to properly define datasets with functionality such as data shuffling and asynchronous data pre-loading (long-term probably also GGUF support for very large datasets that don't fit in RAM). Using such datasets I would then write a more high-level API that trains a feed-forward neural network given a dataset and compute graph (and optinally labels) as input. In that high-level API I would then start using `ggml_backend_sched`.

Right now I have a prototype for a dataset in user space.

slaren (approved):

PR #952, opened 2024-09-07
vulkan: mul_mat: fix UB with small warps

smeso:
When the device's warp size is less than 16,
it is possible for [loadstride_a](https://github.com/ggerganov/ggml/blob/master/src/vulkan-shaders/mul_mm.comp#L114) and [loadstride_b](https://github.com/ggerganov/ggml/blob/master/src/vulkan-shaders/mul_mm.comp#L115) to be set to 0.
Because they are calculated as: the workgroup size, multiplied by LOAD_VEC_* (which can be 1) and divided by 16. And the workgroup size is set to be the same as the warp/subgroup size.

The loadstride_* variables are used as increments in the loops that populate the buffers used for the multiplication.

When they are 0, they cause an infinite loop.
But infinite loops without side-effects are UB and the values of loadstride_* are known at compile time.
So, the compiler quietly optimizes all the loops away. As a consequence, the buffers are not populated and the multiplication result is just a matrix with all elements set to 0.

We prevent the UB by making sure that the workgroup size will never be less than 16, even if our device has a smaller warp size (e.g. 8).

### DIFF
--- a/src/ggml-vulkan.cpp
+++ b/src/ggml-vulkan.cpp
@@ -1145,11 +1145,11 @@ static void ggml_vk_load_shaders(vk_device& device) {
     // mulmat
     std::initializer_list<uint32_t> warptile_l = { 128, 128, 128, 16, device->subgroup_size * 2, 64, 2, 4, 4, device->subgroup_size };
     std::initializer_list<uint32_t> warptile_m = { 128,  64,  64, 16, device->subgroup_size, 32, 2, 4, 2, device->subgroup_size };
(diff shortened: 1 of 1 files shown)


ggerganov:
cc @0cc4m

0cc4m:
It's true that the shader was written for devices with warp size of 32 or 64. It breaks for smaller values. Does it even output correct results with warp size 16 or is the result still wrong?

I don't think I have a way to test this.

smeso:
When I tested it, it passed all tests (e.g. test-backend-ops). So it seems to work, but I don't know if there is any corner case in which it would return incorrect results. I can add and run more tests if you can think of anything else that is worth trying.

0cc4m (approved):
@smeso No, your confirmation that it works is enough. Looks good, this won't affect other devices.

PR #961, opened 2024-09-16
vulkan: fix build for GGML_VULKAN_RUN_TESTS, add TFLOPS to logging

jeffbolznv:
GGML_VULKAN_RUN_TESTS currently has build errors like:

```
ggml-vulkan.cpp(5251,5): error C2664: 'void ggml_vk_dispatch_pipeline(ggml_backend_vk_context *,vk_context &,vk_pipeline &,const std::initializer_list<vk::DescriptorBufferInfo> &,size_t,const void *,std::array<uint32_t,3>)': cannot convert argument 4 from 'initializer list' to 'const std::initializer_list<vk::DescriptorBufferInfo> &'
ggml-vulkan.cpp(5497,10): error C2039: 'staging': is not a member of 'ggml_backend_vk_context'
```

and runtime assertions like:
```
ggml-vulkan.cpp:2486: GGML_ASSERT(pipeline->descriptor_set_idx < pipeline->descriptor_sets.size()) failed
```

Fix these, and also add teraflops calculation to the matmul tests.


jeffbolznv:
Ping for review.

I'm curious, is it better to do these sort of fixes in llama.cpp or in ggml?

ggerganov:
Sorry for the delay. Pinging @0cc4m as they are the main maintainer of the vulkan backend.
If we don't get a response in a day or two, will try to review myself - please ping again if necessary.

> I'm curious, is it better to do these sort of fixes in llama.cpp or in ggml?

`llama.cpp` has more eyes in general, so chances for review are better. Although some of the Vulkan PRs are also stale over there (https://github.com/ggerganov/llama.cpp/pull/9407).

src/ggml-vulkan.cpp
```diff
@@ -1,3 +1,6 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: MIT
+
```
0cc4m, line 3: Why this note? I don't think this is appropriate. If we put a copyright note here, it'd have to mention others before Nvidia.

0cc4m (commented):
Apart from that copyright note, I think this is fine.

src/ggml-vulkan.cpp
jeffbolznv: It's company policy to add a copyright notice in files that are modified. I'm happy to add other notes here as well, maybe `Copyright (c) 2023-2024 The ggml authors` like is in some other files?

0cc4m: I'm the main author of this file, but many people have contributed. Should we mention them all? The more significant ones? Or just a group?

@ggerganov Do you have an opinion here?

ggerganov: > @ggerganov Do you have an opinion here?

I don't really understand the necessity for adding these SPDX identifiers in the source files. The copyright and license of the project is already specified in the [LICENSE](https://github.com/ggerganov/ggml/blob/master/LICENSE) file, so these identifiers are IMO pointless.

> Should we mention them all? The more significant ones? Or just a group?

We definitely don't want to maintain redundant copyright notices and license identifiers in the source files. There is commit history already available which provides all that information in greater detail. Any entity that wants to claim copyright should be added to the [AUTHORS](https://github.com/ggerganov/ggml/blob/master/AUTHORS) list - this is the main purpose for the file to exist and it's much easier to maintain.

IMO the simplest solution is to remove the identifiers.

jeffbolznv: OK, I think the AUTHORS file will suffice. I've regenerated that file and removed the copyright comments.

0cc4m (approved):

PR #973, opened 2024-09-30
ggml : fix ggml_cast

iboB:
restoring a line which was (I assume accidentally) deleted in #966


JohannesGaessler:
Thank you.

JohannesGaessler (approved):

PR #980, opened 2024-10-02
ggml/ex: calculate accuracy in graph, adapt MNIST

JohannesGaessler:
This PR extends the MNIST example to allow the calculation of the model accuracy within a GGML graph instead of in user code. This is particularly relevant when shuffling data since it's much easier to guarantee that the correct values are being compared. The accuracy is calculated by first calculating `ARGMAX` (CUDA implementation added) of the logits and the labels and then counting the number of equal elements (new GGML op `COUNT_EQUAL`). Because the output is an integer the result does not change depending on the number of threads/CUDA blocks and the use of atomic adds. I considered an implementation where the accuracy is calculated in a fused way from the logits and labels but I think that that would be less reusable with a negligible difference for overall performance. For training all relevant statistics can now be obtained by just copying them from the output tensors.

I added a `warp_reduce_sum(int)` function for CUDA which makes calls with a half argument ambiguous so it has become necessary to do explicit casts to float in some places.

### DIFF
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -116,6 +116,11 @@ static void init_tensor_uniform(ggml_tensor * tensor, float min = -1.0f, float m
     } else if (tensor->type == GGML_TYPE_I8 || tensor->type == GGML_TYPE_I16 || tensor->type == GGML_TYPE_I32) {
         // This is going to create some weird integers though.
         ggml_backend_tensor_set(tensor, data.data(), 0, ggml_nbytes(tensor));
+    } else if (tensor->type == GGML_TYPE_I64) {
+        // Integers with a size of 8 bytes can be set by mirroring the float data, the specific values are again not really meaningful.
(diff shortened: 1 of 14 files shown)


tests/test-backend-ops.cpp
```diff
@@ -113,7 +113,7 @@ static void init_tensor_uniform(ggml_tensor * tensor, float min = -1.0f, float m
             }
         }
         ggml_backend_tensor_set(tensor, dataq.data(), 0, dataq.size());
-    } else if (tensor->type == GGML_TYPE_I8 || tensor->type == GGML_TYPE_I16 || tensor->type == GGML_TYPE_I32) {
+    } else if (tensor->type == GGML_TYPE_I8 || tensor->type == GGML_TYPE_I16 || tensor->type == GGML_TYPE_I32 || tensor->type == GGML_TYPE_I64) {
```
slaren, line 116: This will cause a read out of bounds with I64, since `data` only has 32-bit floats.

slaren (approved):

PR #981, opened 2024-10-03
sync : llama.cpp

ggerganov:
TODO: update examples


examples/mnist/mnist-common.h
```diff
@@ -57,17 +57,18 @@ struct mnist_model {
...
+            for (size_t i = 0; i < ggml_backend_reg_count(); ++i) {
+                fprintf(stderr, "  - %s\n", ggml_backend_reg_name(ggml_backend_reg_get(i)));
             }
             exit(1);
         }
 
         fprintf(stderr, "%s: using %s backend\n", __func__, backend_name.c_str());
-        backend = ggml_backend_reg_init_backend(backend_index, nullptr);
+
+        ggml_backend_t backend = ggml_backend_init_by_name(backend_name.c_str(), NULL);
```
slaren, line 71: The equivalent in the current API would be to find a device instead. Something like tis:
```C++
    mnist_model(const std::string & backend_name) {
        const ggml_backend_dev_t dev = ggml_backend_dev_by_name(backend_name.c_str());
        if (reg == nullptr) {
            fprintf(stderr, "%s: ERROR: backend %s not found, available:\n", __func__, backend_name.c_str());
            for (size_t i = 0; i < ggml_backend_reg_count(); ++i) {
                fprintf(stderr, "  - %s\n", ggml_backend_reg_name(ggml_backend_reg_get(i)));
            }
            exit(1);
        }

        fprintf(stderr, "%s: using %s backend\n", __func__, backend_name.c_str());

        ggml_backend_t backend = ggml_backend_dev_init(dev, NULL);
```

ggerganov: Thanks, can you also verify the changes in `test-backend-buffer.cpp`

slaren: `test-backend-buffer` should also test all the devices, however the test is very outdated and doesn't really test anything useful at this point. It may be better to remove it entirely. I may try to create a more comprehensive test to ensure that backends are implementing the API correctly in the future.

PR #982, opened 2024-10-03
examples: add dataset, data shuffling to MNIST

JohannesGaessler:
This PR adds an explicit datatype to represent a dataset to the MNIST example. The dataset also has support for shuffling the data. This works by separating the data into shards and creating a vector of integers as a permutation of those shards. The dataset is shuffled by shuffling the permutation which then dictates which shards are retrieved with `get_batch`. With a shard size of 1 this then requires a lot more individual data transfers per batch, making the CUDA training ~10x slower compared to transferring the entire batch in one contiguous block. With a shard size of 10 the CUDA training is still ~2x slower but this is largely compensated by faster convergence per epoch. For actual use the overhead from shuffling is probably negligible, especially once asynchronous data loading is implemented.

My next goals will be to add dataset support to GGML, create a more high-level API, add `ggml_backend_sched` support, and add more tests. I think after that it would make sense to start looking into actual applications such as finetuning in llama.cpp.

### DIFF
--- a/examples/mnist/mnist-common.cpp
+++ b/examples/mnist/mnist-common.cpp
@@ -14,7 +14,7 @@
 #include <string>
 #include <utility>
 
-bool mnist_image_load(const std::string & fname, float * buf, const int nex) {
+bool mnist_image_load(const std::string & fname, mnist_dataset & dataset) {
     auto fin = std::ifstream(fname, std::ios::binary);
     if (!fin) {
         fprintf(stderr, "failed to open images file %s\n", fname.c_str());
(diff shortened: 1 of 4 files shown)


slaren (approved):
It may make sense to allocate the dataset in a host buffer, since that should make the transfers to the GPU significantly faster. You can use the `ggml_backend_dev_host_buffer_type` to obtain a host buffer type for a device, if it is supported by the backend. It may also be faster to queue all the transfers with `ggml_backend_tensor_set_async`.

PR #987, opened 2024-10-09
ggml-alloc : remove buffer_id from leaf_alloc

danbev:
This commit removes the `buffer_id` field from the `leaf_alloc` struct.

The motivation for is that this field is only written to and never read/used as far as I can tell. Each `tensor_alloc` has a `buffer_id` field and this is what caused me to look into this more closely, to understand what the `buffer_id` in `leaf_alloc` was used for.

* [Notes on Graph Allocator](https://github.com/danbev/learning-ai/blob/main/notes/ggml.md#graph-allocator-galloc) (which is where/how I noticed this)

### DIFF
--- a/src/ggml-alloc.c
+++ b/src/ggml-alloc.c
@@ -348,7 +348,6 @@ struct tensor_alloc {
 };
 
 struct leaf_alloc {
-    int buffer_id;
     struct tensor_alloc leaf;
 };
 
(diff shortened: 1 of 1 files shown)


slaren:
Yes this is correct. At some point `buffer_id` was moved to `tensor_alloc` from `node_alloc` and `leaf_alloc`, but `leaf_alloc` was not updated.

slaren (approved):

PR #992, opened 2024-10-16
sync : llama.cpp

ggerganov:


ggerganov:
Oh, I accidentally squashed the sync. Will force push `master` to keep the commits separate

PR #996, opened 2024-10-22
increase cuda_cpy block size

bssrdf:
This PR gives a small performance boost to cuda backend cpy op.  All test cases in test-backend-ops passed.


JohannesGaessler (approved):
You should be able to get even better performance by replacing the scalar copies with copies of `half2`/`float2`.

bssrdf:
Thanks for approving. I read somewhere that vectorized global memory read does not improve throughput much. It helps with loading from shared memory.

PR #1006, opened 2024-11-04
sync : llama.cpp

ggerganov:
TODO:
- [x] fix build and tests


slaren:
The `test-opt` should just be disabled until it is updated in #988, since the opt interface has been removed it cannot be updated.

slaren:
I disabled all tests and examples that depend on `ggml_opt`. They should be re-enabled or removed in #988.

PR #1015, opened 2024-11-15
fix(metal) : offset integer overflows

pminev:
-- While running StableDiffusion.cpp locally with Metal some offsets overflow and results in incorrect calculations

There are tests in `test-backend-ops.cpp` for[ im2col](https://github.com/ggerganov/ggml/blob/9d0708e863f3aa2fc1eb0b75d433303c30bd0dbc/tests/test-backend-ops.cpp#L3397-L3398) which are too big and are taking a long time, so they are commented for now, but can be used for testing the overflow.


ggerganov (approved):

ggerganov:
Need to rebase on latest `master` and we can merge.

PR #1016, opened 2024-11-15
sync : llama.cpp

ggerganov:


slaren:
The moved files weren't removed, they are duplicated now.

slaren:
You probably need to update the oneAPI etc SYCL libraries. See #10267 for more details.

mudler:
> You probably need to update the oneAPI etc SYCL libraries. See [ggerganov/llama.cpp#10267](https://github.com/ggerganov/llama.cpp/pull/10267) for more details.

Gonna check in a bit. Thank you @slaren!

mudler:
that was it indeed! just for reference if anyone is facing this, had to update the Docker base image from `intel/oneapi-basekit:2024.2.0-devel-ubuntu22.04` to `intel/oneapi-basekit:2025.0.0-0-devel-ubuntu22.04`

PR #1019, opened 2024-11-16
feat: add `GGML_UNARY_OP_ARGMAX` Metal kernel

PABannier:
This PR implements the Metal kernel used for the `GGML_UNARY_OP_ARGMAX` operation.

It is necessary for [Encodec.cpp](https://github.com/PABannier/encodec.cpp) to run on the Metal backend.

### DIFF
--- a/src/ggml-metal/ggml-metal.m
+++ b/src/ggml-metal/ggml-metal.m
@@ -3485,7 +3485,8 @@ static void ggml_metal_encode_node(
                 [encoder setComputePipelineState:pipeline];
                 [encoder setBuffer:id_src0 offset:offs_src0        atIndex:0];
                 [encoder setBuffer:id_dst  offset:offs_dst         atIndex:1];
-                [encoder setBytes:&nb01    length:sizeof( int64_t) atIndex:2];
+                [encoder setBytes:&ne00    length:sizeof( int64_t) atIndex:2];
+                [encoder setBytes:&nb01    length:sizeof(uint64_t) atIndex:3];
 
                 [encoder dispatchThreadgroups:MTLSizeMake(nrows, 1, 1) threadsPerThreadgroup:MTLSizeMake(1, 1, 1)];
             } break;
--- a/src/ggml-metal/ggml-metal.m
+++ b/src/ggml-metal/ggml-metal.m
@@ -348,6 +348,7 @@ enum ggml_metal_kernel_type {
     GGML_METAL_KERNEL_TYPE_SUM_ROWS,
     GGML_METAL_KERNEL_TYPE_POOL_2D_AVG_F32,
     GGML_METAL_KERNEL_TYPE_POOL_2D_MAX_F32,
+    GGML_METAL_KERNEL_TYPE_ARGMAX,
 
     GGML_METAL_KERNEL_TYPE_COUNT
 };
@@ -869,6 +870,7 @@ static struct ggml_backend_metal_context * ggml_metal_init(ggml_backend_dev_t de
         GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_SIN,                           sin,                            true);
         GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_COS,                           cos,                            true);
         GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_SUM_ROWS,                      sum_rows,                       true);
+        GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_ARGMAX,                        argmax,                         true);
         GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_POOL_2D_AVG_F32,               pool_2d_avg_f32,                true);
         GGML_METAL_ADD_KERNEL(GGML_METAL_KERNEL_TYPE_POOL_2D_MAX_F32,               pool_2d_max_f32,                true);
     }
@@ -996,6 +998,7 @@ static bool ggml_metal_supports_op(const struct ggml_backend_metal_device_contex
         case GGML_OP_RMS_NORM:
         case GGML_OP_GROUP_NORM:
             return has_simdgroup_reduction;
+        case GGML_OP_ARGMAX:
         case GGML_OP_NORM:
         case GGML_OP_ROPE:
             return true;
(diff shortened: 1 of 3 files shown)


ggerganov:
Thanks, will merge these soon. Just need to sync the https://github.com/ggerganov/llama.cpp/pull/10238 from `llama.cpp` to avoid resolving conflicts manually.

src/ggml-metal/ggml-metal.m
```diff
@@ -3469,6 +3472,21 @@ static void ggml_metal_encode_node(
...
+                const int64_t nrows = ggml_nrows(src0);
+
+                id<MTLComputePipelineState> pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_ARGMAX].pipeline;
+
+                [encoder setComputePipelineState:pipeline];
+                [encoder setBuffer:id_src0 offset:offs_src0        atIndex:0];
+                [encoder setBuffer:id_dst  offset:offs_dst         atIndex:1];
+                [encoder setBytes:&ne00    length:sizeof( int64_t) atIndex:2];
+
+                [encoder dispatchThreadgroups:MTLSizeMake(1, 1, 1) threadsPerThreadgroup:MTLSizeMake(nrows, 1, 1)];
```
ggerganov, line 3488: Threadgroup sizes are typically maximum of 1024, so this implementation will stop working when `nrows` is larger than 1024. Instead, we should launch multiple threadgroups and distribute the rows across them.

Also, the implementation currently assumes `src0` is contiguous, so add a `GGML_ASSERT` for that or extend the implementation by providing and taking account of the strides.

ggerganov (changes requested):

PABannier:
@ggerganov Thanks for the review. I added your changes:
1. Taking into account the strides to avoid asserting that `src0` is contiguous,
2. Parallelizing over threadgroups instead of threads.

I have a follow-up question regarding the tests/CI: is the `argmax` Metal kernel tested? Looking at the logs I only found one place where `argmax` is mentioned: `19:   ARGMAX(type=f32,ne=[10,100,1,1]): not supported [BLAS]`. Am I missing something?

src/ggml-metal/ggml-metal.metal
```diff
@@ -1344,6 +1344,22 @@ kernel void kernel_ssm_scan_f32(
     }
 }
 
+kernel void kernel_argmax(
+        device   const void * x,
+        device      int32_t * dst,
+        constant    int64_t & nb01,
+        uint tgpig[[threadgroup_position_in_grid]]) {
+    device const float * x_row = (device const float *) ((device const char *) x + tgpig * nb01);
```
ggerganov, line 1352: We should still add `GGML_ASSERT(ggml_is_contiguous_1(src0));` because this assumes that the rows are contiguous. Also add `GGML_ASSERT(nb00 == ggml_type_size(src0->type));` since the elements in the row have to be contiguous as well.

ggerganov (approved):

src/ggml-metal/ggml-metal.metal
PABannier: Done :)

src/ggml-metal/ggml-metal.metal
```diff
@@ -1344,6 +1344,23 @@ kernel void kernel_ssm_scan_f32(
...
+
+    dst[tgpig] = 0;
+
+    for (int i = 0; i < ncols; i++) {
+        if (x_row[i] > x_row[dst[tgpig]]) {
+            dst[tgpig] = i;
+        }
+    }
```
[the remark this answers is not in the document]
ggerganov: @PABannier Do you plan to give this a try? We can merge as it is and optimize later. Or you can take a look at the CUDA https://github.com/ggerganov/llama.cpp/pull/10441 implementation and try to optimize this in a similar way.

If you prefer to keep it like this for now, just add a TODO comment to optimize it in the future so we don't forget.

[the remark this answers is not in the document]
PABannier: Oh thanks for the guidance @ggerganov ! I was actually thinking about it, and trying to understand the implementation of the softmax Metal kernel. I definitely want to give it a try. Will try to push something by the end of the week!

PABannier:
@ggerganov I pushed a SIMD implementation of the kernel. All tests are passing :)

I'd like to benchmark both the non-SIMD and SIMD implementations of the kernel. Is there an existing snippet of code that benchmarks the latency and throughput of the kernel as @slaren did in https://github.com/ggerganov/llama.cpp/pull/10441 ?

slaren:
You can obtain the performance measurements with `test-backend-ops -o ARGMAX perf`. The same tests that I used should already be there, but you can add your own to `make_test_cases_perf`.

PABannier:
Thanks @slaren ! I run `test-backend-ops -o ARGMAX perf` after adding the 3 test cases.

Here are the results:

With SIMD vectorization:
```
ARGMAX(type=f32,ne=[100,10,1,1]):                   286720 runs -     3.53 us/run -        3 kB/run -    1.07 GB/s
ARGMAX(type=f32,ne=[1024,12,1,1]):                   90112 runs -    11.38 us/run -       48 kB/run -    4.03 GB/s
ARGMAX(type=f32,ne=[5438,3,1,1]):                    65536 runs -    16.11 us/run -       63 kB/run -    3.77 GB/s
```

Without SIMD vectorization:
```
ARGMAX(type=f32,ne=[100,10,1,1]):                    73728 runs -    14.75 us/run -        3 kB/run -    0.26 GB/s
ARGMAX(type=f32,ne=[1024,12,1,1]):                   16384 runs -   120.66 us/run -       48 kB/run -    0.38 GB/s
ARGMAX(type=f32,ne=[5438,3,1,1]):                     8192 runs -   626.23 us/run -       63 kB/run -    0.10 GB/s
```

This is indeed much faster! Thanks for the suggestion @slaren !

This is ok for me on my side if you want to merge :)

tests/test-backend-ops.cpp
```diff
@@ -3821,6 +3824,10 @@ static std::vector<std::unique_ptr<test_case>> make_test_cases_perf() {
     test_cases.emplace_back(new test_bin_bcast(ggml_add, GGML_TYPE_F32, {4096, 1, 1, 1}, {1,   1, 1, 1}));
     test_cases.emplace_back(new test_bin_bcast(ggml_add, GGML_TYPE_F32, {4096, 1, 1, 1}, {1, 512, 1, 1}));
 
+    test_cases.emplace_back(new test_argmax(GGML_TYPE_F32, { 100, 10, 1, 1}));
+    test_cases.emplace_back(new test_argmax(GGML_TYPE_F32, {1024, 12, 1, 1}));
+    test_cases.emplace_back(new test_argmax(GGML_TYPE_F32, {5438,  3, 1, 1}));
+
```
slaren, line 3830: The reason there aren't any argmax perf test cases currently is because the changes haven't been synced from llama.cpp yet, but it might better to remove these to avoid merge conflicts in the next sync.
```suggestion
```

slaren (approved):

PR #1020, opened 2024-11-17
sync : llama.cpp

ggerganov:


ggerganov:
@JohannesGaessler The `test-opt` seg faults from time to time:

https://github.com/ggml-org/ci/tree/results/ggml/17/8ebfcc5f125085d51e0953b2d8230c21358650/ggml-4-x86-cuda-v100#ctest_release

I can reproduce this also on `master` with my CUDA box by letting the following command run for a while:

```bash
while ./bin/test-opt  ; do date ; done
```

The CPU backend would also occasionally fail in the `test_gradient_accumulation` test:

```
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=2, subtest=weights): OK
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=2, subtest=results): OK
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=3, subtest=grads): FAIL
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=3, subtest=weights): OK
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=3, subtest=results): OK
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=4, subtest=grads): FAIL
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=4, subtest=weights): OK
  test_gradient_accumulation(high_level=no, nbatch_physical=1, loss_type=mean, epoch=4, subtest=results): OK
  test_regression(subtest=weights): OK
  116/118 tests passed
  Backend CPU: FAIL
```

JohannesGaessler:
When running `test-opt` on a loop I also eventually see a test failure but the test failure manifests in a different way than on Georgi's machine. I always get a failure in `test_regression`. However, when I remove all other tests `test_regression` is no longer failing. The only common reference between the tests is an instance of `ggml_backend_sched_t`. When I modified the tests to initialize and free a dedicated instance for each test I got the same failure pattern as Georgi. Also looking at the tests I'm noticing that we don't actually have any test code outside of `test-opt` that utilizes `ggml_backend_sched` so we can't conclusively tell whether `ggml_backend_sched` or `ggml_opt` is causing a failure in `test-opt`. Even without the optimization code, `ggml_backend_sched` is a fairly complex component and I think it would make sense to add tests (but if we do this we should coordinate). I would in principle be willing to write tests for `ggml_backend_sched` but I don't feel very confident in my understanding of the code and will likely require assistance.

cc: @slaren

slaren:
It is indirectly tested in any test that run llama.cpp. I agree it would be good to have tests for it, but it's not an easy component to write unit tests for. At some point I will probably rewrite it in C++ with testing in mind. I don't see how it could cause `ggml_opt_grad_acc` to return `NULL`, however.

JohannesGaessler:
I think I worded my post poorly. I agree that in this particular instance the bug is overwhelmingly likely in `ggml_opt`. I was just thinking that tests would be nice to have in general.

PR #1022, opened 2024-11-18
ggml-opt: fix data corruption

JohannesGaessler:
Small fixup to https://github.com/ggerganov/ggml/pull/988 .

Since there is now a distinction between statically and dynamically allocated gradients some workarounds for dynamically allocated gradients can be removed. This simplifies the code a bit and reduces the possible causes for the issues in https://github.com/ggerganov/ggml/pull/1020 .

### DIFF
--- a/src/ggml-opt.cpp
+++ b/src/ggml-opt.cpp
@@ -14,51 +14,51 @@
 #include <vector>
 
 struct ggml_opt_dataset {
-    struct ggml_context   * ctx;
-    ggml_backend_buffer_t   buf;
-    struct ggml_tensor    * data;
-    struct ggml_tensor    * labels;
+    struct ggml_context   * ctx    = nullptr;
+    ggml_backend_buffer_t   buf    = nullptr;
+    struct ggml_tensor    * data   = nullptr;
+    struct ggml_tensor    * labels = nullptr;
 
-    int64_t ndata;
-    int64_t ndata_shard;
-    size_t  nbs_data;
-    size_t  nbs_labels;
+    int64_t ndata       = -1;
+    int64_t ndata_shard = -1;
+    size_t  nbs_data    = -1;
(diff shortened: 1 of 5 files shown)


JohannesGaessler:
Looking at the code I'm finding small defects which (while not a problem currently) could lead to bugs depending on how the API is used. I think it would make sense to accumulate the resulting small changes into a single PR to reduce the time needed for reviewing and testing. I'll convert this to a draft for now.

src/ggml-opt.cpp
```diff
@@ -237,25 +237,33 @@ static ggml_tensor * map_tensor(std::map<ggml_tensor *, ggml_tensor *> & tensor_
...
-        ggml_build_forward_expand(new_graph, map_tensor(tensor_map, ctx, graph->nodes[i]));
+    GGML_ASSERT(dst->n_leafs == src->n_leafs);
+    for (int i = 0; i < src->n_nodes; i++) {
+        ggml_build_forward_expand(dst, map_tensor(tensor_map, ctx, src->nodes[i]));
     }
-    for (int i = 0; i < graph->n_nodes; ++i) {
-        const size_t igrad_src = ggml_hash_find(&graph->visited_hash_set, graph->nodes[i]);
-        const size_t igrad_dst = ggml_hash_find(&new_graph->visited_hash_set, new_graph->nodes[i]);
-        graph->grads[igrad_dst]     = new_graph->grads[igrad_src];
-        graph->grad_accs[igrad_dst] = new_graph->grad_accs[igrad_src];
```
JohannesGaessler, line 255: This seems to be why the tests were sometimes crashing. The direction of the copy is wrong so garbage could be written that is supposed to only be copied.

ggerganov: Can confirm `test-opt` no longer crashes or fails.

slaren (approved):

PR #1023, opened 2024-11-20
FIX: CMAKE_RUNTIME_OUTPUT_DIDRECTORY is only set when GGML_STANDALONE

petersktang:
configure_file for ggml-common.h, ggml-metal.metal, ggml-metal-impl.h should not be copied when not GGML_STANDALONE.


petersktang:
change incorrect. please ignore.

PR #1033, opened 2024-11-30
GGUF: ggml backend support for writing tensor data

JohannesGaessler:
This PR adds ggml backend support for writing tensor data to a GGUF file. Currently a workaround is needed where the data is first copied to new tensors with data in RAM, which the GGUF code can then access via `memcpy`. This PR makes it so that instead a fake tensor is reconstructed from `gguf_tensor_info` which can then be passed to `ggml_backend_tensor_get`. I'm not sure whether this is the best solution; a lot of the fileds in `gguf_tensor_info` are the same as in `ggml_tensor`, is there a reason why you couldn't just directly store a `ggml_tensor` as one of the fields in `gguf_tensor_info`?

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -6380,6 +6380,7 @@ struct gguf_tensor_info {
     uint64_t offset; // offset from start of `data`, must be a multiple of `ALIGNMENT`
 
     // for writing API
+    ggml_backend_buffer_t buffer;
     const void * data;
     size_t size;
 };
@@ -7317,6 +7318,7 @@ void gguf_add_tensor(
 
     ctx->infos[idx].type   = tensor->type;
     ctx->infos[idx].offset = 0;
+    ctx->infos[idx].buffer = tensor->buffer;
     ctx->infos[idx].data   = tensor->data;
     ctx->infos[idx].size   = ggml_nbytes(tensor);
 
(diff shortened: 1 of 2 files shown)


slaren:
It should be ok to store the tensor in `gguf_tensor_info`, but I think it would require a refactor to avoid duplicating the data since the gguf loader also uses this struct to load the tensor info.

JohannesGaessler:
I did a refactor to store a `ggml_tensor` instead of effectively mirrored fields. It seems to work correctly for MNIST but I think I'll open a PR in the llama.cpp repository to ensure that it works there as well (there are also some slight API changes that I would suggest). While I'm at it I'll also tackle https://github.com/ggerganov/ggml/issues/1038 as well.

src/ggml.c
```diff
@@ -6404,43 +6394,43 @@ static size_t gguf_type_size(enum gguf_type type) {
...
+
+    /* if (INT64_MAX/info->ne[2] <= info->ne[0]*info->ne[1]) { */
+    /*     fprintf(stderr, "%s: invalid number of elements (%" PRIu64 ")\n", __func__, info->ne[2]); */
+    /*     return false; */
+    /* } */
+
+    /* if (INT64_MAX/info->ne[3] <= info->ne[0]*info->ne[1]*info->ne[2]) { */
+    /*     fprintf(stderr, "%s: invalid number of elements (%" PRIu64 ")\n", __func__, info->ne[3]); */
+    /*     return false; */
+    /* } */
```
ggerganov, line 6433: Why are these checks commented?

JohannesGaessler: This was just something I did for a WIP version. I have a version with more changes and the checks re-eenabled on my local machine. I'll make a PR to llama.cpp either today or tomorrow.

PR #1034, opened 2024-11-30
feat: add `GGML_PAD_REFLECT_1D` operation

PABannier:
This PR adds the CPU and Metal kernels of the `PAD_REFLECT_1D` operation. A description of the operation can be found [here](https://pytorch.org/docs/stable/generated/torch.nn.ReflectionPad1d.html).

It is yet another operation that needs to be supported for [Encodec.cpp](https://github.com/PABannier/encodec.cpp) to work.


ggerganov (approved):

tests/test-pad-reflect-1d.cpp
ggerganov, line 1: We should refactor these type of test programs into a single test that has reference results for the operators (such as conv, conv_transpose, pad_reflect, etc.) and verifies all of them in a single run using just the CPU backend.

PR #1039, opened 2024-12-03
sync : llama.cpp

ggerganov:


src/ggml-cpu/amx/amx.d
slaren, line 1: This file and `mmq.d` should be removed.

Issue #1058, opened 2025-01-02
Add support for loading out-of-tree backends


rgerganov:
We are developing a proprietary ggml backend and we won't be able to open-source it any time soon (if at all). We still want to be able to load our backend in runtime and use it for all of the `llama` binaries in `llama.cpp`. Unfortunately, this is not possible at the moment because only a [pre-defined set](https://github.com/ggerganov/ggml/blob/master/src/ggml-backend-reg.cpp#L533-L552) of in-tree backends are being loaded. A quick workaround for this is the following patch which we use as stop-gap solution:

```
@@ -537,6 +537,11 @@ void ggml_backend_load_all_from_path(const char * dir_path) {
     bool silent = false;
 #endif
 
+    // check the environment variable GGML_BACKEND to load an out-of-tree backend
+    const char * env_backend = std::getenv("GGML_BACKEND");
+    if (env_backend) {
+        ggml_backend_load_best(env_backend, silent, dir_path);
+    }
     ggml_backend_load_best("blas", silent, dir_path);
     ggml_backend_load_best("cann", silent, dir_path);
     ggml_backend_load_best("cuda", silent, dir_path);
```

Are there any plans for adding support for dynamic loading of out-of-tree ggml backends?

I see the API is versioned with `GGML_BACKEND_API_VERSION` but it seems there is no clear separation of the header definitions needed for 3rd party backend implementations. Right now I have to include `include/ggml.h`, `include/ggml-alloc.h`, `include/ggml-backend.h` and `src/ggml-backend-impl.h` to build our backend.

We have been also implementing a [PJRT](https://github.com/openxla/xla/blob/main/xla/pjrt/c/README.md) plugin where they provide header definitions and API versioning. I think we can do the same for ggml.

slaren:
I do not intend to add this functionality, but contributions to do this would be welcome.

Backends can use all the ggml API in ggml-base, which excludes the backend registry and other backends. `GGML_BACKEND_API_VERSION` should be bumped when there are incompatible changes to the ggml-base ABI, but I cannot guarantee however that this will be done reliably until we start versioning ggml.

rgerganov:
As a first step I posted a patch for setting the backend name with env var. The next steps would be to add all backend APIs to public headers and enforce versions when loading.

slaren:
> The next steps would be to add all backend APIs to public headers and enforce versions when loading.

[...]

PR #1059, opened 2025-01-03
ggml : allow setting the backend name with env variable

rgerganov:
ref: #1058


slaren:
I am not sure that this is the best way to implement this, because `ggml_backend_load_best` only looks in a list of pre-defined paths that may not be entirely obvious to the user, and this behavior may change in the future. It may make more sense to use `ggml_backend_load` with the full file path in the environment variable.

slaren (approved):

PR #1078, opened 2025-01-21
fix: a bug in examples/simple/simple-backend

Yangxiaoz:
there is an error result in ggml/examples/simple/simple-backend: 
erro_result

and The correct result should be as shown in the Readme file：
CORRECT_RESULT

The above error occurs when traversing the matrix elements index while printing

### DIFF
--- a/examples/simple/simple-backend.cpp
+++ b/examples/simple/simple-backend.cpp
@@ -190,9 +190,9 @@ int main(void) {
     ggml_backend_tensor_get(result, out_data.data(), 0, ggml_nbytes(result));
 
     // expected result:
-    // [ 60.00 110.00 54.00 29.00
-    //  55.00 90.00 126.00 28.00
-    //  50.00 54.00 42.00 64.00 ]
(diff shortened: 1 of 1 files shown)


FSSRepo:
Could you also make the change in `simple-ctx`?

Yangxiaoz:
Thank  you for taking the time to check this PR.
But the problem raised in  this PR seems to exist only in simple-backend, and there is no  corresponding bug in simple-ctx.




Y.
***@***.***



        



         Original
         
       
From:Steward Garcia ***@***.***&gt;
Sent Time:2025-02-8- 00:08
To:ggerganov/ggml ***@***.***&gt;
Cc:Shawn yang ***@***.***&gt;, Author ***@***.***&gt;
Subject:Re: [ggerganov/ggml] fix: a bug in examples/simple/simple-backend (PR#1078)




Could you also make the change in simple-ctx?

FSSRepo:
yeah, I realized

Yangxiaoz:
I think there is no problem, you can merge it directly.   @FSSRepo

PR #1081, opened 2025-01-23
Add option to not print stack on abort

WilliamTambellini:
Add option/envvar to disable stack printing on abort. 
Also link some unittests with Threads to fix a link error on ubuntu/g++11 (undef ref to create_pthread).

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -128,6 +128,9 @@ static void ggml_print_backtrace_symbols(void) {
 #endif
 
 static void ggml_print_backtrace(void) {
+    const char* GGML_NO_BACKTRACE = getenv("GGML_NO_BACKTRACE");
+    if (GGML_NO_BACKTRACE)
+        return;
     char attach[32];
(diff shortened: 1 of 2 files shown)


src/ggml.c
```diff
@@ -128,6 +128,9 @@ static void ggml_print_backtrace_symbols(void) {
 #endif
 
 static void ggml_print_backtrace(void) {
+    const char* GGML_NO_BACKTRACE = getenv("GGML_NO_BACKTRACE");
+    if (GGML_NO_BACKTRACE)
+        return;
```
slaren, line 133:
```suggestion
    const char * GGML_NO_BACKTRACE = getenv("GGML_NO_BACKTRACE");
    if (GGML_NO_BACKTRACE) {
        return;
    }
```

slaren (approved):

matiaslin (approved):
Cool. Tested locally and we see no stack when aborting if `GGML_NO_BACKTRACE` is specified.

WilliamTambellini:
tks Diego

PR #1097, opened 2025-02-02
cmake: Make SSE4.2 optional

ckastner:
This makes it possible to build microarchitectural level x86-64-v1 ([link](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels).

Even though this probably has little practical use, -v1 is still the baseline for amd64 on Debian, so this simplifies package building. Of course higher levels are provided as well, but providing -v1 is a must.

Note that I did not actually test this for the MSVC case (don't have it), I simply assumed that what was said for FMA and F16C also applies to the older SSE4.2.


ckastner:
Looking at other parts of the codebase, I realize that this change is incomplete and would lead to inconsistencies, eg with GGML_CPU_ALL_VARIANTS.

I'll re-file once I have a complete solution.

PR #1103, opened 2025-02-05
rpc: fix known RCE in rpc-server

retr0reg:
This is a issue of the `rpc-server` that I started working on around December last year, the vulnerability derives from the fact that `ggml_backend_cpu_buffer_cpy_tensor` in `./ggml/src/ggml-backend.cpp`'s `cpy_tensor` action, registered usually as `.cpy_tensor`:

```cpp
static bool ggml_backend_cpu_buffer_cpy_tensor(ggml_backend_buffer_t buffer, const struct ggml_tensor * src, struct ggml_tensor * dst) {
    if (ggml_backend_buffer_is_host(src->buffer)) {
        memcpy(dst->data, src->data, ggml_nbytes(src));
        return true;
    }
    return false;

    GGML_UNUSED(buffer);
}
```

The `ggml_backend_cpu_buffer_cpy_tensor` operation seemed secure since the implementation of the strict boundary checks (really strick) in `deserialize_tensor()` after [`GHSA-wcr5-566p-9cwj`](https://github.com/ggerganov/llama.cpp/security/advisories/GHSA-wcr5-566p-9cwj), [`GHSA-5vm9-p64x-gqw9j`](https://github.com/ggerganov/llama.cpp/security/advisories/GHSA-5vm9-p64x-gqw9), [`GHSA-mqp6-7pv6-fqjf`](https://github.com/ggerganov/llama.cpp/security/advisories/GHSA-mqp6-7pv6-fqjf). Which made `buf->data` pointer manipulation and `write-what-where` and `read-what-where` impossible. However, this tiny leftover from the last patch that can cause RCE from sophisticated exploitation that plays with *partial-writing* and *`ggml`'s unique memory management* (I wrote a 10k word writeup for it very interesting). The major issue here is how `ggml_nbytes` is depends on a `Tensor`'s `nb[]` / `ne[]`. This allows out-of-bound writing (or copying) when `src` applied for a larger context where `src->data` can be in range and passes `deserialize_tensor` checks, but copies over `dst->data` when we try to manipulate `ggml_nbytes(src)` something larger than `ggml_backend_buffer_get_size(tensor->buffer);` (which is the actual size of the buffer)

To mitigate this, checks are added in `rpc_server::copy_tensor`, before `dst` and `src` being pass into `dst_buf->iface.cpy_tensor` after `deserialize_tensor` checks;

```cpp
    uint64_t src_size   = (uint64_t) ggml_nbytes(src);
    uint64_t dst_data   = (uint64_t) dst->data;
    uint64_t dst_base   = (uint64_t) ggml_backend_buffer_get_base(dst->buffer);
    uint64_t dst_buf_sz = (uint64_t) ggml_backend_buffer_get_size(dst->buffer);

    if (dst_data + src_size > dst_base + dst_buf_sz) {
        GGML_PRINT_DEBUG("[%s] out-of-bounds write in rpc_server::copy_tensor:\n"
                         "    write range : [0x%" PRIx64 ", 0x%" PRIx64 "]\n"
                         "    buffer base: [0x%" PRIx64 ", 0x%" PRIx64 "]\n",
                         __func__,
                         dst_data,
                         dst_data + src_size,
                         dst_base,
                         dst_base + dst_buf_sz);
        ggml_free(ctx);
        return false;
    }
```

Here `dst_data` *(`dst->data`)* + `src_size` *(`ggml_nbytes(src)`)* is compared to`dst_base` (the `dst` context, `ggml_backend_buffer_get_base(dst->buffer);`) + `dst_buf_sz` *(real `dst` buffer size, `ggml_backend_buffer_get_size(dst->buffer);`)*. Prevents the `memcpy` in `ggml_backend_cpu_buffer_cpy_tensor` (the actual sink) from copying `ggml_nbytes(src)` *(size)* of `src->data` to `dst->data`, with size that's larger than the `dst->data` + `dst->size` (the actual `dst` buffer size). Preventing the overflow (out-of-bound write) from happening, and eliminating the chance of any furthermore exploitations

### DIFF
--- a/src/ggml-rpc/ggml-rpc.cpp
+++ b/src/ggml-rpc/ggml-rpc.cpp
@@ -1045,7 +1045,28 @@ bool rpc_server::copy_tensor(const rpc_msg_copy_tensor_req & request, rpc_msg_co
         ggml_free(ctx);
         return false;
     }
-    GGML_PRINT_DEBUG("[%s] src->buffer: %p, dst->buffer: %p\n", __func__, (void*)src->buffer, (void*)dst->buffer);
+
+    uint64_t src_size   = (uint64_t) ggml_nbytes(src);
+    uint64_t dst_data   = (uint64_t) dst->data;
+    uint64_t dst_base   = (uint64_t) ggml_backend_buffer_get_base(dst->buffer);
+    uint64_t dst_buf_sz = (uint64_t) ggml_backend_buffer_get_size(dst->buffer);
+
+    if (dst_data + src_size > dst_base + dst_buf_sz) {
+        GGML_PRINT_DEBUG("[%s] out-of-bounds write in rpc_server::copy_tensor:\n"
+                         "    write range : [0x%" PRIx64 ", 0x%" PRIx64 "]\n"
+                         "    buffer base: [0x%" PRIx64 ", 0x%" PRIx64 "]\n",
+                         __func__,
+                         dst_data,
+                         dst_data + src_size,
+                         dst_base,
+                         dst_base + dst_buf_sz);
+        ggml_free(ctx);
+        return false;
+    }
+
+    GGML_PRINT_DEBUG("[%s] src->buffer: %p, dst->buffer: %p\n",
+                     __func__, (void*) src->buffer, (void*) dst->buffer);
+
     response.result = ggml_backend_buffer_copy_tensor(src, dst);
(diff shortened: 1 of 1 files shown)


slaren (approved):

ggerganov (approved):

rgerganov:
@retr0reg kudos!

PR #1114, opened 2025-02-13
fix: wget doesn't work due to invalid option

okamototk:
wget option is invalid in download model scripts. Let me send PR.


ggerganov:
You are using an old `wget`, the option exists in newer versions:

https://stackoverflow.com/questions/4686464/how-can-i-show-the-wget-progress-bar-only/32491843#32491843

okamototk:
@ggerganov Sorry for about that.

PR #1116, opened 2025-02-17
cpu: Shift score for SSE4.2

ckastner:
It seems that SSE4.2 first appeared in Nehalem, and FMA/F16C in Ivy Bridge. This precedence would also match that given by the microarchitectural levels agreed to by Intel and AMD:

   https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

### DIFF
--- a/src/ggml-cpu/cpu-feats-x86.cpp
+++ b/src/ggml-cpu/cpu-feats-x86.cpp
@@ -266,6 +266,10 @@ static int ggml_backend_cpu_x86_score() {
     int score = 0;
     cpuid_x86 is;
 
+#ifdef GGML_SSE42
+    if (!is.SSE42()) { return 0; }
+    score += 1<<2;
+#endif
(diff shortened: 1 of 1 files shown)


slaren:
The score given to each feature should roughly represent the performance gain expected from it. SSE4.2 is much more important than F16C or FMA by itself. In practice F16C and FMA are always used together with AVX or later, so this doesn't really matter either way.

ckastner:
It seems I misunderstood the purpose of the score, thanks for the clarification.

Though the check for SSE4.2 should still precede the check for F16C/FMA, right? Nehalem has SSE4.2 but not FMA/F16C, but unless I'm mistaken SSE4.2 would not get picked up there because the check for FMA/F16C before that return early if unavailable.

ckastner:
As discussed in #1120, this PR is ineffective, based on a misunderstanding, therefore retracting.

PR #1118, opened 2025-02-20
ggml_compute_forward_concat() for arbitrary tensor type

vmobilis:
It is needed for embeddings in [stable-diffusion.cpp](https://github.com/leejet/stable-diffusion.cpp) when weight type is not f32.

And thanks for the program! )

### DIFF
--- a/src/ggml-cpu/ggml-cpu.c
+++ b/src/ggml-cpu/ggml-cpu.c
@@ -6094,6 +6094,135 @@ static void ggml_compute_forward_repeat_back(
 
 // ggml_compute_forward_concat
 
+static void ggml_compute_forward_concat_any(
+    const struct ggml_compute_params * params,
+    struct ggml_tensor * dst) {
+
+    const struct ggml_tensor * src0 = dst->src[0];
+    const struct ggml_tensor * src1 = dst->src[1];
+
+    const size_t len = src0->nb[0];
+
+    const int ith = params->ith;
+    const int nth = params->nth;
+
+    GGML_TENSOR_BINARY_OP_LOCALS
+
+    const int32_t dim = ggml_get_op_params_i32(dst, 0);
+
+    GGML_ASSERT(dim >= 0 && dim < 4);
+
+    int64_t o[4] = {0, 0, 0, 0};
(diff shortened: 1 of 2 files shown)


src/ggml-cpu/ggml-cpu.c
```diff
@@ -6094,6 +6094,135 @@ static void ggml_compute_forward_repeat_back(
...
 // ggml_compute_forward_concat
 
+static void ggml_compute_forward_concat_any(
+    const struct ggml_compute_params * params,
+    struct ggml_tensor * dst) {
+
+    const struct ggml_tensor * src0 = dst->src[0];
+    const struct ggml_tensor * src1 = dst->src[1];
+
+    const size_t len = src0->nb[0];
```
ggerganov, line 6104: This should take the `ggml_type_size` instead. Also we should add a `GGML_ASSERT` in `ggml_concat` that the 2 tensors are the same type.

vmobilis: @ggerganov, thanks for review, is everything correct now?

vmobilis: @ggerganov, thanks for review, is everything correct now?

Or should be ggml_type_size() used for all weight types, for consistency?

For example, change:
https://github.com/ggml-org/ggml/blob/9a4acb374565f4146b8d6eb1cffdcd7d437d1ba2/src/ggml-cpu/ggml-cpu.c#L6104
To:
`GGML_ASSERT(ggml_type_size(src0->type) == sizeof(float));`

ggerganov: Yes. The check should be directly in `ggml_concat` so that it applies to all implementations of this OP - not just the CPU.

vmobilis:
@ggerganov, it seems that my patch is wrong, the correct one is https://github.com/leejet/stable-diffusion.cpp/pull/601.

vmobilis:
@ggerganov, is anything wrong with commit?

ggerganov (approved):

ggerganov:
We can merge - I thought that something was wrong based on your last comment so didn't review.

PR #1121, opened 2025-02-24
Support pure float16 add/sub/mul/div operations in the CUDA (and CPU) backend

cmdr2:
This change increases the operator coverage for the float16 data type in the CUDA backend (for add/sub/mul/div).

At present, ggml requires the second tensor to be converted to float32, which doubles the VRAM requirement, and makes ggml a bit unintuitive. Especially since it's completely possible to add or multiply two float16 tensors in ggml.

float16 is fairly common for inference, and since ggml is a tensor library for ML, it's not uncommon to want to add or multiply two large tensors. Requiring the `src1` tensor to be float32 doubles the VRAM requirement for that tensor.

I tested that this works (on my 3060 12 GB), has half the peak VRAM usage (compared to float32+float32 ops), is decently faster than float32+float32 addition. Also `test-backend-ops` continues to pass.

[Example program for float16 addition](https://gist.github.com/cmdr2/7a739f5f626c8e7b52fa4b3f87f66f1d), [Example program for float32 addition](https://gist.github.com/cmdr2/fa03a4184530ee8128146c54042fd6c2).

Related issue: https://github.com/ggml-org/ggml/issues/455
> For example, ggml_mul currently also works with just F32 input, which prevents from having 1D F16 norm tensors. This is not a huge drawback since these tensors are usually small, but would be nice to also support F16.

Thanks!
PS: I'm new to ggml, apologies if I missed something obvious! Happy to fix.

### DIFF
--- a/src/ggml-cpu/ggml-cpu.c
+++ b/src/ggml-cpu/ggml-cpu.c
@@ -1333,15 +1333,35 @@ inline static void ggml_vec_cpy_i32(const int n, int32_t * y, const int32_t * x)
 inline static void ggml_vec_set_f16(const int n, ggml_fp16_t * x, const int32_t v) { for (int i = 0; i < n; ++i) x[i] = v; }
 inline static void ggml_vec_set_bf16(const int n, ggml_bf16_t * x, const ggml_bf16_t v) { for (int i = 0; i < n; ++i) x[i] = v; }
 inline static void ggml_vec_add_f32 (const int n, float * z, const float * x, const float * y) { for (int i = 0; i < n; ++i) z[i]  = x[i] + y[i]; }
+inline static void ggml_vec_add_f16 (const int n, ggml_fp16_t * z, const ggml_fp16_t * x, const ggml_fp16_t * y) {
+    for (int i = 0; i < n; ++i) {
+        z[i] = GGML_FP32_TO_FP16(GGML_FP16_TO_FP32(x[i]) + GGML_FP16_TO_FP32(y[i]));
+    }
+}
 inline static void ggml_vec_add1_f32(const int n, float * z, const float * x, const float   v) { for (int i = 0; i < n; ++i) z[i]  = x[i] + v;    }
 inline static void ggml_vec_acc_f32 (const int n, float * y, const float * x)                  { for (int i = 0; i < n; ++i) y[i] += x[i];        }
 inline static void ggml_vec_acc1_f32(const int n, float * y, const float   v)                  { for (int i = 0; i < n; ++i) y[i] += v;           }
 inline static void ggml_vec_sub_f32 (const int n, float * z, const float * x, const float * y) { for (int i = 0; i < n; ++i) z[i]  = x[i] - y[i]; }
+inline static void ggml_vec_sub_f16 (const int n, ggml_fp16_t * z, const ggml_fp16_t * x, const ggml_fp16_t * y) {
+    for (int i = 0; i < n; ++i) {
+        z[i] = GGML_FP32_TO_FP16(GGML_FP16_TO_FP32(x[i]) - GGML_FP16_TO_FP32(y[i]));
+    }
+}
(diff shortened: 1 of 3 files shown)


JohannesGaessler (commented):
If at all possible, please also add a CPU implementation and a test case for `test-backend-ops` to assert that the implementations are consistent.

cmdr2:
@JohannesGaessler Thanks, no problem. It looks like the [current CPU implementation for fp16 addition](https://github.com/ggml-org/ggml/blob/9a4acb374565f4146b8d6eb1cffdcd7d437d1ba2/src/ggml-cpu/ggml-cpu.c#L4277) isn't working correctly when adding a N-sized tensor with a single element tensor.

Also, I had to change the [initial assert](https://github.com/ggml-org/ggml/blob/9a4acb374565f4146b8d6eb1cffdcd7d437d1ba2/src/ggml-cpu/ggml-cpu.c#L4284) to match what the [f32 CPU implementation does](https://github.com/ggml-org/ggml/blob/9a4acb374565f4146b8d6eb1cffdcd7d437d1ba2/src/ggml-cpu/ggml-cpu.c#L4059C17-L4059C80) (`can_repeat` instead of `same_shape`), otherwise it would fail to run several test cases.

For e.g. this test produces incorrect values on the CPU: `ADD(type=f16,ne=[1,1,1,1],nr=[32,1,1,1])`. I printed the computed values, and the CUDA backend values are correct, while the CPU values are incorrect.

So I'm not sure if fp16 addition is actually used with the CPU backend right now. But I'll dig further into why the CPU implementation for fp16 isn't giving correct responses.

WilliamTambellini:
@cmdr2  tks. We are interested by that PR.

cmdr2:
@JohannesGaessler I've added fp16 op support for add/sub/mul/div on the CPU backend as well. Also added test cases in `test-backend-ops`.

`test-backend-ops` passes.

Can you please take a look when you get a chance? Thanks!

cmdr2:
As a side note, since I saw a plan to refactor `ggml-cpu.c` - it would really help if it was written as a cpp file, since we could use template functions to cut down a huge number of redundant lines of code. Maybe that's already the plan?

Right now, as you know, there are several copies of each operator function, and many of them still don't support broadcasting or non-contiguous cases, simply because that's not been copied from the fp32 implementation.

We could easily de-duplicate a lot of them using function templates (including the permutations of `src0`/`src1`/`dst` data types).

I'm happy to help, if it's simply a lack of manpower. In a different PR, of course.

Thanks

ggerganov:
> As a side note, since I saw a plan to refactor `ggml-cpu.c` - it would really help if it was written as a cpp file, since we could use function templates to cut down a huge number of redundant lines of code. Maybe that's already the plan?
> 
> Right now, as you know, there are several copies of each operator function, and many of them still don't support broadcasting or non-contiguous cases, simply because that's not been copied from the fp32 implementation.
> 
> We could easily de-duplicate a lot of them using function templates (including the permutations of `src0`/`src1`/`dst` data types).
> 
> I'm happy to help, if it's simply a lack of manpower. In a different PR, of course.
> 
> Thanks

Yes, we should switch to a C++ implementation and reduce code duplication. The most important thing is to not go overboard with fancy C++ features and keep them at a very minimum. Basically  just templates + trivial containers such as `std::vector` where it would make sense.

ggerganov (approved):

JohannesGaessler (approved):

JohannesGaessler:
You'll most likely probably need to adjust the `supports_op` function for the Metal backend. It should be fairly straightforward.

cmdr2:
@JohannesGaessler Thanks, this is what I'm planning in `ggml-metal.m`. Does this look okay?

```cpp
case GGML_OP_ADD:
case GGML_OP_SUB:
case GGML_OP_MUL:
case GGML_OP_DIV:
    return op->src[0]->type == GGML_TYPE_F32;
```

cmdr2:
@JohannesGaessler Sent a PR for this: https://github.com/ggml-org/ggml/pull/1123

Thanks for your help!

PR #1138, opened 2025-03-10
cmake: Enable specifying exact PowerPC CPU architecture

ckastner:
Useful for when build host != use host. For example Debian's build daemons build packages for other hosts, so we want to target their CPUs, and not the build daemon's.

https://gcc.gnu.org/onlinedocs/gcc/RS_002f6000-and-PowerPC-Options.html#index-mcpu-10


slaren (commented):
For consistency, automatic CPU detection should be enabled with `GGML_NATIVE`:

```cmake
if (GGML_NATIVE)
    # detect CPU
else()
    if (GGML_CPU_POWERPC_CPUTYPE)
        list(APPEND ARCH_FLAGS -mcpu=${GGML_CPU_POWERPC_CPUTYPE})
    endif()
endif()
```

CMakeLists.txt
```diff
@@ -126,7 +126,8 @@ option(GGML_RVV              "ggml: enable rvv"              ON)
 option(GGML_VXE              "ggml: enable vxe"              ON)
 
 option(GGML_CPU_ALL_VARIANTS "ggml: build all variants of the CPU backend (requires GGML_BACKEND_DL)" OFF)
-set(GGML_CPU_ARM_ARCH "" CACHE STRING "ggml: CPU architecture for ARM")
+set(GGML_CPU_ARM_ARCH        "" CACHE STRING "ggml: CPU architecture for ARM")
+set(GGML_CPU_POWERPC_CPUTYPE "" CACHE STRING "ggml: CPU type for PowerPC")
```
slaren, line 130: Maybe we should consolidate all of these into a single option in the future, but it's not important right now.

ckastner: I'm happy to propose something once I've done more of a deep dive into other architectures. I'm not yet that familiar with the nuances of the other architectures, or how they would map. For example, `-march` vs. PowerPC's  `-mcpu` and `-mtune`.

slaren: I am not sure either, but probably an option to set the value of `-march` or `-mcpu` would work well enough in most cases.

PR #1141, opened 2025-03-12
cpu: use templates for de-duplicating some operators

cmdr2:
Note: This PR only deletes lines from `ggml-cpu.c`, it does not modify any functions in it. I'm not sure why git diff tries to combine them as changes. Standard `diff` shows the correct output. Actual diff of `ggml-cpu.c`: https://gist.github.com/cmdr2/a76df5af311417619788e8330b1908b3

This PR de-duplicates some of the easily-templatized functions in `ggml-cpu.c`. It takes inspiration from `binbcast.cu`.

* Binary: `add`, `sub`, `mul`, `div`
* Unary: `abs`, `sgn`, `neg`, `step`, `tanh`, `elu`, `relu`, `sigmoid`, `hardsigmoid`, `exp`, `hardswish`, `sqr`, `sqrt`, `sin`, `cos`, `log`

This removes the op implementation functions from `ggml-cpu.c` (around 2000 lines). As a side-effect, all the functions now support bf16 as well as non-contiguous `src1`.

The next PRs will attempt to:
1. Use the `traits` table to remove the ugly-looking mass of template parameters and if-else conditions. That will enable cleaner support of quantized operations. For e.g. taking inspiration from the [quantized add function](https://github.com/ggml-org/ggml/blob/2efc170b66649686addc760bee6d74381eb7c15c/src/ggml-cpu/ggml-cpu.c#L4624).
2. Use row-based type conversions, to work with the existing traits table functions.
3. Move even more operator functions to C++ files.

The performance is the same as the current implementation. It also passes all the runners on [ggml-ci](https://github.com/ggml-org/ggml/commits/cmdr2-op-templates/), which tested non-contiguous inputs (in `SAM`) and `vDSP` on Mac.


cmdr2:
I'm happy to modify the approach based on feedback. This was one way to get this de-duplicated, without changing too many things at once.

cmdr2:
Also, please let me know if the templates-based approach is too ugly. I can fast-track the `traits` table-based approach, and use row-wise conversion functions instead.

I was trying to minimize the amount of changes in this PR, which is why I left that out. The existing implementation doesn't use row-wise conversions either, so this PR just refactors that.

cmdr2:
I have a cleaner solution coming up shortly, please hold off.

PR #1142, opened 2025-03-12
examples: change GPT-2 KV cache to fp16 to take advantage of tensor cores

bssrdf:
This PR changed default KV cache type to FP16 to use tensor cores if available. This brings the inference performance much closer to ```llama.cpp```.


ggerganov (commented):
This is a bit irrelevant for the purpose of this example, because it is not meant to deliver optimal performance, but rather to demonstrate `ggml` usage. I think we can simply add a comment here that either `GGML_TYPE_F16` or `GGML_TYPE_F32` can be used.

bssrdf:
@ggerganov, fair enough! I have reverted back to FP32 and added a comment. Thanks for reviewing.

ggerganov (approved):

PR #1144, opened 2025-03-13
cpu: de-duplicate some of the operators and refactor

cmdr2:
Note: This PR only deletes lines from `ggml-cpu.c`, it does not modify any functions in it. I'm not sure why git diff tries to combine them as changes. Standard `diff` shows the correct diff of `ggml-cpu.c`: https://gist.github.com/cmdr2/a76df5af311417619788e8330b1908b3

This PR de-duplicates some of the easily-templatized functions in `ggml-cpu.c`. It takes inspiration from `binbcast.cu` and `common.cuh`.

* Binary: `add`, `sub`, `mul`, `div`
* Unary: `abs`, `sgn`, `neg`, `step`, `tanh`, `elu`, `relu`, `sigmoid`, `hardsigmoid`, `exp`, `hardswish`, `sqr`, `sqrt`, `sin`, `cos`, `log`

This removes the op implementation functions from `ggml-cpu.c` (around 2000 lines). As a side-effect, all the functions now support bf16 as well as non-contiguous `src1`.

The performance is the same as the current implementation. It also passes all the runners on [ggml-ci](https://github.com/ggml-org/ggml/commits/cmdr2-op-templates2/), which tested non-contiguous inputs (in `SAM`) and `vDSP` on Mac.


src/ggml-cpu/binary_ops.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+template <float (*op)(float, float), typename src0_t, typename src1_t, typename dst_t>
+static inline void vec_binary_op_non_contiguous(const int64_t n, const int64_t ne10, const int64_t nb10, dst_t * z, const src0_t * x, const src1_t * y) {
+    constexpr auto src0_to_f32 = type_conversion_table<src0_t>::to_f32;
+    constexpr auto src1_to_f32 = type_conversion_table<src1_t>::to_f32;
+    constexpr auto f32_to_dst  = type_conversion_table<dst_t >::from_f32;
+
+    for (int i = 0; i < n; i++) {
+        int i10 = i % ne10;
+        src1_t * y_ptr = (src1_t *)((char *)y + i10*nb10);
```
ggerganov, line 44: Silence some warnings:

```suggestion
        const src1_t * y_ptr = (const src1_t *)((const char *)y + i10*nb10);
```

src/ggml-cpu/binary_ops.cpp
```diff
@@ -0,0 +1,157 @@
...
+        z[i] = f32_to_dst(op(src0_to_f32(x[i]), src1_to_f32(*y_ptr)));
+    }
+}
+
+template <float (*op)(float, float), typename src0_t, typename src1_t, typename dst_t>
+static void apply_binary_op(const struct ggml_compute_params * params, struct ggml_tensor * dst) {
```
ggerganov, line 50: In the `llama.cpp` coding guidelines (https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md#coding-guidelines) we decided to omit optional keywords `struct` and `enum` in C++ code. We can adopt the same guidelines for `ggml` to make things more consistent:

```suggestion
static void apply_binary_op(const ggml_compute_params * params, ggml_tensor * dst) {
```

src/ggml-cpu/binary_ops.h
ggerganov, line 1: Source files should use dashes instead of underscores.

src/ggml-cpu/unary_ops.cpp
```diff
@@ -0,0 +1,186 @@
...
+    const struct ggml_tensor * src0 = dst->src[0];
+
+    GGML_ASSERT(ggml_is_contiguous_1(src0) && ggml_is_contiguous_1(dst) && ggml_are_same_shape(src0, dst));
+
+    GGML_TENSOR_UNARY_OP_LOCALS
+
+    GGML_ASSERT( nb0 == sizeof(dst_t));
+    GGML_ASSERT(nb00 == sizeof(src0_t));
+
+    const auto [ir0, ir1] = get_thread_range(params, src0);
```
ggerganov, line 88: Hopefully all C++ compilers support structured bindings today. If we encounter issues, we might have to return simple struct of integers.

ggerganov (approved):

ggerganov:
Wait for @slaren's review before merging. Thanks!

cmdr2:
> Wait for slaren's review before merging. Thanks!

@ggerganov Yes, will do :)

Thanks for the comments, fixed them in the latest commit. The CI runner also [passes](https://github.com/ggml-org/ggml/commits/cmdr2-op-templates2).

src/ggml-cpu/binary_ops.cpp
cmdr2: Thanks, good point.

src/ggml-cpu/binary_ops.cpp
cmdr2: Sorry about that! Fixed all the instances of this pattern. No warnings shown now on my Linux build for these files.

slaren:
I will review when I have a chance, but I may not be able to do it for a few days.

cmdr2:
Thanks @slaren No hurry :)

@slaren @ggerganov I made some progress on the next PR (a bigger refactor). It's working after the refactor, and passes the runners on ggml-ci. `ggml-cpu.c` is at about 3500 lines after this refactoring (down from ~15,000 lines).

Just thought of sharing the approach early, to see if you see any obvious red-flags.

I'll split it into two PRs:
1. Move the [SIMD Mapping lines](https://github.com/ggml-org/ggml/blob/ffed2f047dc523957744e72bab654320de2da0a9/src/ggml-cpu/ggml-cpu.c#L425-L1304) into a separate header file, and the [`ggml_vec_` function lines](https://github.com/ggml-org/ggml/blob/ffed2f047dc523957744e72bab654320de2da0a9/src/ggml-cpu/ggml-cpu.c#L1405-L2401) into a separate header file. The vec file imports the SIMD mappings file. No other changes.
2. Move all the operator function lines (except `mul_mat`) into a separate C++ file. This file imports the vec functions header.

The operator functions file needed a very small number of cosmetic changes, like replacing direct references to `type_traits_cpu` with `ggml_get_type_traits_cpu()`, using `static_cast<ggml_op_pool>(opts[0])` etc.

At a broad level, does this approach raise any obvious red-flags for you? It seems to work on all the ggml-ci runners, and doesn't seem to skip any CPU extension. But I'll test and read the diff a lot more before submitting the PRs.

Thanks

cmdr2:
@slaren Gentle ping, thanks :)

src/ggml-cpu/binary-ops.cpp
```diff
@@ -0,0 +1,157 @@
...
+        if (is_src1_contiguous) {
+            // src1 is broadcastable across src0 and dst in i1, i2, i3
+            const int64_t nr0 = ne00 / ne10;
+
+            for (int64_t r = 0; r < nr0; ++r) {
+#ifdef GGML_USE_ACCELERATE
+                if (vDSP_op != nullptr) {
+                    vDSP_op((float *)src1_ptr, 1, (float *)src0_ptr + r*ne10, 1, (float *)dst_ptr + r*ne10, 1, ne10);
+                    continue;
+                }
```
slaren, line 106: These cast are generating warnings, but can be avoided entirely.
```suggestion
                if constexpr (std::is_same_v<src0_t, float> && std::is_same_v<src1_t, float> && std::is_same_v<dst_t, float>) {
                    if (vDSP_op != nullptr) {
                        vDSP_op(src1_ptr, 1, src0_ptr + r*ne10, 1, (float *)dst_ptr + r*ne10, 1, ne10);
                        continue;
                    }
                }
```

src/ggml-cpu/binary-ops.cpp
```diff
@@ -0,0 +1,157 @@
...
+    } else if (src0->type == GGML_TYPE_BF16 && src1->type == GGML_TYPE_F32  && dst->type == GGML_TYPE_F32) {
+        apply_binary_op<op, ggml_bf16_t, float, float>(params, dst);
+    } else if (src0->type == GGML_TYPE_F16  && src1->type == GGML_TYPE_F32  && dst->type == GGML_TYPE_F16) {
+        apply_binary_op<op, ggml_fp16_t, float, ggml_fp16_t>(params, dst);
+    } else if (src0->type == GGML_TYPE_F16  && src1->type == GGML_TYPE_F32  && dst->type == GGML_TYPE_F32) {
+        apply_binary_op<op, ggml_fp16_t, float, float>(params, dst);
+    } else {
+        fprintf(stderr, "%s: unsupported types: dst: %s, src0: %s, src1: %s\n", __func__,
+            ggml_type_name(dst->type), ggml_type_name(src0->type), ggml_type_name(src1->type));
+        GGML_ABORT("fatal error");
```
slaren, line 139:
```suggestion
        GGML_ABORT("%s: unsupported types: dst: %s, src0: %s, src1: %s\n", __func__,
            ggml_type_name(dst->type), ggml_type_name(src0->type), ggml_type_name(src1->type));
```

slaren (approved):
Sorry for taking for so long to get to this, other than the minor suggestions it looks good to me.

src/ggml-cpu/binary-ops.cpp
```diff
@@ -0,0 +1,157 @@
+#include "binary-ops.h"
+
+#if defined(GGML_USE_ACCELERATE)
+#include <Accelerate/Accelerate.h>
+
+#define vDSP_fn_t(f) void (*f)(const float *, vDSP_Stride, const float *, vDSP_Stride, float *, vDSP_Stride, vDSP_Length)
```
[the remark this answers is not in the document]
cmdr2: Thanks, done!

cmdr2:
Thanks @slaren , no problem! I've fixed the comments. Shall I go ahead and squash-merge?

ggerganov:
Yup, go ahead. Thanks!

LostRuins:
Hi all, apologies for the necropost. 

I was trying to port some old code in TTS.cpp that was using an earlier version of the `ggml` library to a newer version of `ggml`, and I'm running into a new assert after this pull. 

Specifically, there is now a constraint for `ggml_are_same_shape(src0, src1)` at https://github.com/ggml-org/ggml/blob/2621d5ad344906d2dc8c5dbfc112f83e75269e49/src/ggml-cpu/binary-ops.cpp#L65 that constraint was not there before, for example in `ggml_compute_forward_mul_f32` https://github.com/ggml-org/ggml/blob/4b344baace2f9013939717906c15021c7b483d6e/src/ggml-cpu/ggml-cpu.c#L5432

May I check what should I change in order to avoid this assert? Is there a reason for this new limitation?

Thanks!

PR #1158, opened 2025-03-26
ci: disable test-opt for now

qnixsynapse:
SYCL backend doesn't implement optimizer and training and it is not on my todo list for now.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -34,6 +34,7 @@ cd $sd/../
 SRC=`pwd`
 
 CMAKE_EXTRA=""
+CTEST_EXTRA=""
 
 if [ ! -z ${GG_BUILD_CUDA} ]; then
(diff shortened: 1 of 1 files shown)


ci/run.sh
```diff
@@ -54,6 +54,8 @@ if [ ! -z ${GG_BUILD_SYCL} ]; then
     fi
     export ONEAPI_DEVICE_SELECTOR="level_zero:0"
     export ZES_ENABLE_SYSMAN=1
+    # No plan to implement backward pass for now / disable test-opt
+    export GG_BUILD_LOW_PERF=1
```
ggerganov, line 58: We should avoid to modify these build environment settings.

An alternative approach is to introduce `CTEST_EXTRA`, similar to `CMAKE_EXTRA` and use that to pass `-E test-opt` to `ctest` when SYCL is enabled.

ggerganov (approved):

PR #1159, opened 2025-03-26
Update example readme/shell output for gpt-2-backend

rememberlenny:
Currently when you run through the example steps, the printed logs are recommending a bin target that doesn't exist. Updating to the correct reference.


ggerganov (approved):
There are a couple of more instances of `./bin/gpt-2` below that should be corrected.

PR #1165, opened 2025-03-29
cpu: refactor SIMD mappings and vectorized op functions into separate files

cmdr2:
This change only moves the lines related to SIMD Mapping and vectorized op functions into separate files. It does not modify them.

The two pre-computed tables are shared using an `extern`.

This change is required in order to move all the operator functions into separate C++ files (since they require vec functions, which in turn requires SIMD mappings).

This change passes all the runners at [ggml-ci](https://github.com/ggml-org/ggml/commits/cmdr2-refactor-simd-vec/).

Thanks!


cmdr2:
I understand you're all a bit busy right now, so I'm trying to keep the PRs very focused and easy to review. :)

The next PR will only focus on moving the operator functions to separate C++ files. Those functions require a few cosmetic changes, which is why I didn't include that bigger change in this PR.

Thanks

PR #1167, opened 2025-03-29
cpu: move all the operators into a separate c++ file (except mul_mat)

cmdr2:
// This shrinks the size of `ggml-cpu.c` to around 3500 lines (down from 15k lines a few weeks ago). A future PR will organize `ops.cpp` into separate files, and use templates to simplify it further.

It does not modify any lines in `ggml-cpu.c`, it only removes lines. [Actual diff](https://gist.github.com/cmdr2/f686a2e189261d68cb9cd1fd544d691b).

This PR looks big, but only does a few simple things:
1. Moves all the lines related to SIMD Mapping into `simd-mappings.h`
2. Moves all the lines related to vectorized fns into `vec.h` (and `vec.cpp` for non-inline functions)
3. Moves all the operator functions into `ops.cpp` (except `mul_mat`, which has some C-specific behavior)

**Detailed changes:**
* **SIMD Mapping** - No modifications were made to the SIMD Mapping lines.
* **Vectorized functions** - Three minor changes:
    * fix 2 warnings where `ggml_float` was being cast implicitly to `float`. [diff](https://github.com/cmdr2/ggml/commit/a624b8738d802b8c9405d436157716c0eb814234)
    * share the pre-computed table using `extern`.
    * use `GGML_UNUSED` directly instead of the `UNUSED` macro.
* **Operator functions** - Minor cosmetic changes to work in a separate C++ file - [actual diff](https://gist.github.com/cmdr2/4829c87a9cc398db96ca3c0185fb05c8)
    * explicit cast to `float *` or `ggml_fp16_t *` in functions that expected that type, e.g. `ggml_fp16_to_fp32_row()`.
    * `static_cast<ggml_op_pool>(op)` for converting an int to an enum.
    * use the `ggml_get_type_traits_cpu()` function instead of accessing `type_traits_cpu` directly.
    * use `std::hardware_destructive_interference_size` instead of `hardware_destructive_interference_size`.
    * use `GGML_UNUSED` directly instead of the `UNUSED` macro.

This change passes all the runners at [ggml-ci](https://github.com/ggml-org/ggml/commits/cmdr2-refactor-simd-vec/).

Please let me know if this approach requires any changes, happy to fix!

Thanks!


slaren (approved):

src/ggml-cpu/ops.h
```diff
@@ -0,0 +1,128 @@
...
+void ggml_compute_forward_map_custom1(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_map_custom2(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_map_custom3(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_cross_entropy_loss(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_cross_entropy_loss_back(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+void ggml_compute_forward_opt_step_adamw(const struct ggml_compute_params * params, struct ggml_tensor * dst);
+
+#ifdef __cplusplus
+}
+#endif
```
ggerganov, line 128:
```suggestion
#endif

```

Fix a compile warning:

```
In file included from /Users/ggerganov/development/github/ggml/src/ggml-cpu/ggml-cpu.c:14:
/Users/ggerganov/development/github/ggml/src/ggml-cpu/ops.h:128:7: warning: no newline at end of file [-Wnewline-eof]
  128 | #endif
      |       ^
1 warning generated.
```

src/ggml-cpu/simd-mappings.h
```diff
@@ -0,0 +1,884 @@
+#pragma once
+
+#include "ggml-cpu-impl.h"
```
ggerganov, line 3: Should this be simply:

```suggestion
#include "ggml-impl.h"
```

Apart from the `GGML_FP32_TO_FP16` and `GGML_FP16_TO_FP32` macros, which are defined in `ggml-impl.h`, it does not seem to need anything eles from `ggml-cpu-impl.h`. Though double-check with the CI just in case.

src/ggml-cpu/vec.h
```diff
@@ -0,0 +1,806 @@
+// Vectorized functions for fundamental operations
+
+#pragma once
+
+#include "ggml-impl.h"
+#include "simd-mappings.h"
+#include "ggml.h"
+
+#include <assert.h>
```
ggerganov, line 9: Should be moved to `vec.cpp` and changed to `#include <cassert>`.

src/ggml-cpu/vec.h
```diff
@@ -0,0 +1,806 @@
...
+
+#define UNUSED GGML_UNUSED
```
ggerganov, line 11: This macro can be defined per source file (instead of here in the header), as a shortcut for `GGML_UNUSED`. But I think it would be better to just keep using `GGML_UNUSED` to avoid clashes, since `UNUSED` seems like something that could be defined in some other header.

Anyway, this is very minor and no changes needed for this PR.

ggerganov (approved):

src/ggml-cpu/simd-mappings.h
cmdr2: I saw a few lines like these in `ggml-cpu-impl.h`, which define `__F16C__` (which is used in `simd-mappings.h`):

```cpp
// __FMA__ and __F16C__ are not defined in MSVC, however they are implied with AVX2/AVX512
#if defined(_MSC_VER) && (defined(__AVX2__) || defined(__AVX512F__))
#ifndef __FMA__
#define __FMA__
#endif
#ifndef __F16C__
#define __F16C__
#endif
#endif
```

src/ggml-cpu/vec.h
cmdr2: Done, thanks! Used `GGML_UNUSED` in `vec.cpp` and `ops.cpp`, moved this define back into `ggml-cpu.c`.

src/ggml-cpu/vec.h
cmdr2: Done, thanks!

src/ggml-cpu/ops.h
cmdr2: Done, thanks! I think I need to upgrade my g++, it doesn't show these warnings even with `-DGGML_ALL_WARNINGS=1`

cmdr2:
Thanks for the review comments!

I've fixed the suggestions, except the one about `ggml-cpu-impl.h` (since it defines a few macros like `__FP16C__` based on certain conditions). https://github.com/ggml-org/ggml/blob/f06264eda2e2bf6e814db5a32bbf42e0b2b1ed98/src/ggml-cpu/ggml-cpu-impl.h#L42-L74

Maybe we could move those lines into a different file (possibly in `simd-mappings.h`)?

src/ggml-cpu/ops.h
ggerganov: This is with Apple's clang.

ggerganov:
> I've fixed the suggestions, except the one about ggml-cpu-impl.h (since it defines a few macros like __FP16C__ based on certain conditions).

I missed those. In this case, let's leave the include as it is.

> Maybe we could move those lines into a different file (possibly in simd-mappings.h)?

I think these CPU feature defines have to remain in `ggml-cpu-impl.h`. So I think we are good for now.

ggerganov:
Yes, squash-merge and then I'll sync to `llama.cpp` and `whisper.cpp` to see if there are any issues there.

cmdr2:
Tested the latest changes (PR comments) on ggml-ci . All tests pass - https://github.com/ggml-org/ggml/commit/6f4ce79f73ccc9e66334a563b7b190c090601060

Squash-merging..

cmdr2:
The [CI build failure](https://github.com/ggml-org/ggml/commits/master/) on the `master` branch is surprising.

I created a new branch (from master) and ran the CI again, and it ran successfully - https://github.com/ggml-org/ggml/commits/cmdr2-test-master-after-cpu-refactor/

And the CUDA tests pass on my local PC as well.

The CI build failure on the `master` branch is for this line: `  MUL_MAT(type_a=iq1_s,type_b=f32,m=16,n=1,k=256,bs=[1,1],nr=[1,1],per=[0,1,2,3],v=0): [MUL_MAT] NMSE = 0.000648890 > 0.000500000 �[1;31mFAIL�[0m`

Not sure why there was a transient error in the mean-square-error for `mul_mat`.

ggerganov:
Some of the tests occasionally exceed the error threshold, so it is expected to have them failing from time to time. I think this is such case, though let's keep an eye out and see if it happens again.

PR #1177, opened 2025-04-04
ggml : simplify Arm fp16 CPU logic

ggerganov:
ref https://github.com/ggml-org/llama.cpp/pull/12732, https://github.com/ggml-org/ggml/pull/1176

- The ARM_NEON + MSVC checks are not needed any more because we don't support ARM + MSVC (https://github.com/ggml-org/llama.cpp/pull/10890). ARM + Clang is the recommended alternative.
- The `ggml_fp16_internal_t` seems to no longer have any usage, so it can be removed
- The CUDA check in `ggml-impl.h` is needed in order to fix the build with old CUDA compilers v11. These compilers don't support `__fp16` (see https://github.com/ggml-org/llama.cpp/pull/10616), so we don't include Arm headers such as `arm_neon.h` to avoid the compiler seeing this type. Note that the CPU code which is not compiled with `nvcc` can still include Arm headers and use `__fp16`.
- Similar check is done for MUSA per https://github.com/ggml-org/llama.cpp/pull/11843
- The Arm headers included in `ggml-cpu-impl.h` are redundant because they are already included by `ggml-impl.h`. So they are removed to deduplicate the code.

### DIFF
--- a/src/ggml-cpu/simd-mappings.h
+++ b/src/ggml-cpu/simd-mappings.h
@@ -71,7 +71,7 @@
     #define GGML_F16x8              float16x8_t
     #define GGML_F16x8_ZERO         vdupq_n_f16(0.0f)
     #define GGML_F16x8_SET1(x)      vdupq_n_f16(x)
-    #define GGML_F16x8_LOAD(x)      vld1q_f16((const ggml_fp16_internal_t *)(x))
+    #define GGML_F16x8_LOAD(x)      vld1q_f16((const __fp16 *)(x))
     #define GGML_F16x8_STORE        vst1q_f16
     #define GGML_F16x8_FMA(a, b, c) vfmaq_f16(a, b, c)
     #define GGML_F16x8_ADD          vaddq_f16
@@ -99,7 +99,7 @@
     #define GGML_F16_VEC_ZERO           GGML_F16x8_ZERO
     #define GGML_F16_VEC_SET1           GGML_F16x8_SET1
     #define GGML_F16_VEC_LOAD(p, i)     GGML_F16x8_LOAD(p)
-    #define GGML_F16_VEC_STORE(p, r, i) GGML_F16x8_STORE((ggml_fp16_internal_t *)(p), (r)[i])
+    #define GGML_F16_VEC_STORE(p, r, i) GGML_F16x8_STORE((__fp16 *)(p), (r)[i])
     #define GGML_F16_VEC_FMA            GGML_F16x8_FMA
     #define GGML_F16_VEC_ADD            GGML_F16x8_ADD
     #define GGML_F16_VEC_MUL            GGML_F16x8_MUL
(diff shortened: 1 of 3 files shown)


src/ggml-impl.h
```diff
@@ -311,29 +311,24 @@ GGML_API void ggml_aligned_free(void * ptr, size_t size);
 
 // FP16 to FP32 conversion
 
+// 16-bit float
+// on Arm, we use __fp16
+// on x86, we use uint16_t
 #if defined(__ARM_NEON)
```
cmdr2, line 317: Shouldn't there be a corresponding ` && (!defined(__MSC_VER)` here? Since we're only removing the CUDA check

src/ggml-cpu/simd-mappings.h
```diff
@@ -71,7 +71,7 @@
     #define GGML_F16x8              float16x8_t
     #define GGML_F16x8_ZERO         vdupq_n_f16(0.0f)
     #define GGML_F16x8_SET1(x)      vdupq_n_f16(x)
-    #define GGML_F16x8_LOAD(x)      vld1q_f16((const ggml_fp16_internal_t *)(x))
+    #define GGML_F16x8_LOAD(x)      vld1q_f16((const __fp16 *)(x))
```
cmdr2, line 74: We'd still need `reinterpret_cast<__fp16>` for C++ right?

ggerganov:
@max-krasnyansky Could you check if this branch builds on an Arm + MSVC machine (I think you mentioned you have one)? Not sure if the MSVC supports `__fp16` - if it does, we can simplify this code here. Thanks.

src/ggml-impl.h
ggerganov: I think if MSVC supports `__fp16`, we don't need this check. But maybe it was added because it does not support it. If that's the case, we'll have to bring it back and fallback to the reference implementation.

cmdr2: > I think if MSVC supports __fp16, we don't need this check.

Yeah true. The original PR said: `can't use native __fp16 type as it's not supported by MSVC` - https://github.com/ggml-org/llama.cpp/pull/3007

Maybe it has changed since that PR, but I can't find anything to this on the internet though.

ggerganov: Aha, this explains it.

But on the other hand, we have a MSVC+Arm CI job in `llama.cpp` that appears to be passing with this change:

https://github.com/ggml-org/llama.cpp/actions/runs/14262095042/job/39975786169

So maybe it was indeed fixed?

src/ggml-cpu/simd-mappings.h
cmdr2: Sorry, please ignore. My brain disappeared for a second :)

src/ggml-impl.h
cmdr2: Sounds good. This could be merged in, and if there are any reports of regression on MSVC+Arm Neon, then it's easy to fix it back.

cmdr2 (approved):

src/ggml-impl.h
ggerganov: > MSVC cannot be used to build for Arm, the CPU backend does not allow it.

This is because of the issue with the Arm inline assembly that is not supported by the MSVC compiler, correct?

slaren: Support for Arm MSVC was removed when fixing the CPU feature check with `GGML_NATIVE` in cmake, since it would require writing a specific implementation for it. You could still build with MSVC without the inline assembly kernels before that, it would just be slower. Since Clang is always available and shipped with Visual Studio, there was no reason to spend time on that.

ggerganov: Got it. So I think it that case, this PR should be good to merge?

ggerganov: I did some additional digging and found the need for the CUDA and MUSA checks. They do seem to be necessary so I added comments in the code and updated the OP to try to make things a bit more clear. This PR now just removes the `ggml_fp16_internal_t` type which is now redundant and cleans up some macros and include logic.

Let me know if you see any red flags.

slaren (approved):

PR #1178, opened 2025-04-04
cuda : deterministic and faster k_copy_src1_to_contiguous

CISC:
Upstreamed from ikawrakow/ik_llama.cpp#283 and ikawrakow/ik_llama.cpp#313

### DIFF
--- a/src/ggml-cuda/ggml-cuda.cu
+++ b/src/ggml-cuda/ggml-cuda.cu
@@ -1986,7 +1986,7 @@ static inline void prepare_row_mappings(ggml_backend_cuda_context& ctx, int64_t 
     std::vector<char> ids_host(ggml_nbytes(ids));
     const char * ids_dev = (const char *) ids->data;
     CUDA_CHECK(cudaMemcpyAsync(ids_host.data(), ids_dev, ggml_nbytes(ids), cudaMemcpyDeviceToHost, stream));
-    // CUDA_CHECK(cudaStreamSynchronize(stream));
+    CUDA_CHECK(cudaStreamSynchronize(stream));
 
     std::vector<mmid_row_mapping> rmapping(ids->ne[1]*n_ids);
     moe_counts.resize(n_as, 0);
@@ -2017,6 +2017,7 @@ static inline void prepare_row_mappings(ggml_backend_cuda_context& ctx, int64_t 
     for (int i = 0; i < (int)n_as; ++i) cum_moe_counts[i] -= moe_counts[i];
 
     CUDA_CHECK(cudaMemcpyAsync(dev_row_mapping.get(), rmapping.data(), cum_moe_counts[n_as]*sizeof(mmid_row_mapping), cudaMemcpyHostToDevice, stream));
+    CUDA_CHECK(cudaStreamSynchronize(stream));
 }
 
 static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * dst) {
@@ -2061,6 +2062,7 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor * 
         std::vector<char> ids_host(ggml_nbytes(ids));
         const char * ids_dev = (const char *) ids->data;
         CUDA_CHECK(cudaMemcpyAsync(ids_host.data(), ids_dev, ggml_nbytes(ids), cudaMemcpyDeviceToHost, stream));
+        CUDA_CHECK(cudaStreamSynchronize(stream));
         for (int64_t iid1 = 0; iid1 < ids->ne[1]; iid1++) {
             for (int64_t id = 0; id < n_ids; id++) {
                 const int32_t i02 = *(const int32_t *) (ids_host.data() + iid1*ids->nb[1] + id*ids->nb[0]);
--- a/src/ggml-cuda/ggml-cuda.cu
+++ b/src/ggml-cuda/ggml-cuda.cu
@@ -1942,35 +1942,19 @@ struct mmid_row_mapping {
     int32_t i2;
 };
 
-static __global__ void k_copy_src1_to_contiguous(const char * __restrict__ src1_original, char * __restrict__ src1_contiguous,
-                                                 int * __restrict__ cur_src1_row, mmid_row_mapping * __restrict__ row_mapping,
-                                                 const char * __restrict ids, int64_t i02, size_t ids_nb1, size_t ids_nb0,
-                                                 int64_t ne11, int64_t ne10,
-                                                 size_t nb11, size_t nb12) {
-    int32_t iid1 = blockIdx.x;
-    int32_t id = blockIdx.y;
+static __global__ void k_copy_src_to_contiguous(const char * __restrict__ src_original, char * __restrict__ src_contiguous,
+                                                  const mmid_row_mapping * __restrict__ row_mapping,
+                                                  int64_t ne10, int64_t ne11, size_t nb11, size_t nb12) {
+    int32_t i = blockIdx.x;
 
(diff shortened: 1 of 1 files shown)


src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1991,6 +1975,50 @@ static __global__ void k_copy_dst_from_contiguous(char * __restrict__ dst_origin
...
+                                       const ggml_tensor * ids, std::vector<int>& moe_counts, std::vector<int>& cum_moe_counts,
+                                       ggml_cuda_pool_alloc<mmid_row_mapping>& dev_row_mapping) {
+
+    GGML_ASSERT(moe_counts.empty() && cum_moe_counts.empty());
+
+    auto stream = ctx.stream();
+
+    std::vector<char> ids_host(ggml_nbytes(ids));
+    const char * ids_dev = (const char *) ids->data;
+    CUDA_CHECK(cudaMemcpyAsync(ids_host.data(), ids_dev, ggml_nbytes(ids), cudaMemcpyDeviceToHost, stream));
```
JohannesGaessler, line 1988: This code results in a race condition. At the end of the function `ids_host` is deallocated and `ids_host.data()` becomes a dangling pointer which can no longer be used safely. Crucially, CUDA device code is by design being run asynchronously from the host code. So without an explicit synchronization there is no guarantee whatsoever that the call to `cudaMemcpyAsync` is executed before the vector is deallocated.

src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2005,11 +2033,6 @@ static void ggml_cuda_mul_mat_id(ggml_backend_cuda_context & ctx, ggml_tensor *
     const int64_t n_as = ne02;
     const int64_t n_ids = ids->ne[0];
 
-    std::vector<char> ids_host(ggml_nbytes(ids));
-    const char * ids_dev = (const char *) ids->data;
-    CUDA_CHECK(cudaMemcpyAsync(ids_host.data(), ids_dev, ggml_nbytes(ids), cudaMemcpyDeviceToHost, stream));
-    CUDA_CHECK(cudaStreamSynchronize(stream));
```
JohannesGaessler, line 2011: Here a CPU<->GPU synchronization call is being removed. Since synchronization is very bad for performance it could be that this is responsible for a large part of the speedup.

JohannesGaessler (changes requested):
The code on master is deterministic in terms of results. What is not deterministic is how the work is distributed to CUDA blocks but I don't think that matters.

src/ggml-cuda/ggml-cuda.cu
```diff
@@ -1991,6 +1975,50 @@ static __global__ void k_copy_dst_from_contiguous(char * __restrict__ dst_origin
...
+            const int32_t row_id_i = *(const int32_t *) (ids_host.data() + iid1*ids->nb[1] + id*ids->nb[0]);
+            if (row_id_i >= 0 && row_id_i < n_as) {
+                rmapping[cum_moe_counts[row_id_i]++] = {(int)id, (int)iid1};
+            }
+        }
+    }
+
+    for (int i = 0; i < (int)n_as; ++i) cum_moe_counts[i] -= moe_counts[i];
+
+    CUDA_CHECK(cudaMemcpyAsync(dev_row_mapping.get(), rmapping.data(), cum_moe_counts[n_as]*sizeof(mmid_row_mapping), cudaMemcpyHostToDevice, stream));
```
JohannesGaessler, line 2019: This code results in a race condition. At the end of the function `ids_host` is deallocated and `ids_host.data()` becomes a dangling pointer which can no longer be used safely. Crucially, CUDA device code is by design being run asynchronously from the host code. So without an explicit synchronization there is no guarantee whatsoever that the call to `cudaMemcpyAsync` is executed before the vector is deallocated.

CISC:
A quick benchmark after synchronization changes to show that it's still faster:

| Model             | CPU             | GPU        | n_ubatch |     test |        t/s master |     t/s deterministic_k_copy_src1 |
| :---------------- | :-------------- | :--------- | -------: | -------: | ----------------: | ------------------: |
| bailingmoe 16B Q8_0 | Core i7-9700K   | RTX 3090Ti |     2048 |   pp1024 |   3230.35 ± 22.55 |             3962.19 ± 37.60 |
| bailingmoe 16B Q8_0 | Core i7-9700K   | RTX 3090Ti |     2048 |   pp2048 |   3912.80 ± 24.77 |             4992.81 ± 43.04 |
| bailingmoe 16B Q8_0 | Core i7-9700K   | RTX 3090Ti |     2048 |   pp4096 |   3685.56 ± 21.49 |             4631.87 ± 26.74 |

ggerganov:
@CISC Please preserve the authorship of these patches as it was already suggested (https://github.com/ggml-org/ggml/pull/1174#issuecomment-2775253823).

CISC:
> @CISC Please preserve the authorship of these patches as it was already suggested ([#1174 (comment)](https://github.com/ggml-org/ggml/pull/1174#issuecomment-2775253823)).

It feels wrong to fake authorship of the actual commits. Attribution in the PR itself (which can not be a direct upstream of the code anyway) should be sufficient?

Anyway, maybe it's better to ask @ikawrakow directly what his feelings about this is?

ikawrakow:
> Anyway, maybe it's better to ask @ikawrakow directly what his feelings about this is?

In general I don't mind people copying my code, else I wouldn't be publishing it under a MIT license. But `llama.cpp` is special in this regard, so I think it is best to do the right thing. But as IANAL, and as I would like to avoid that someone comes along and starts teaching me that because of the blah blah convention blah blah blah, I leave it up to you and the project owner to determine what the right thing is.

ikawrakow:
> The code on master is deterministic in terms of results. What is not deterministic is how the work is distributed to CUDA blocks but I don't think that matters.

It is not my observation. I get a different result each time I run the code on `llama.cpp` master (on the same hardware, using the same parameters). The reason for this is that `src1` rows end up in a different order in the contiguous copy, which then influence the result as floating point operations are not associative. It is not wrong, but it is still not deterministic. If in doubt, run a few perplexity calculations and observe how the values are different in each run on master, but are always the same with this version of the kernel.

CISC:
> In general I don't mind people copying my code, else I wouldn't be publishing it under a MIT license. But `llama.cpp` is special in this regard, so I think it is best to do the right thing. But as IANAL, and as I would like to avoid that someone comes along and starts teaching me that because of the blah blah convention blah blah blah, I leave it up to you and the project owner to determine what the right thing is.

What address would you like attributed, actual or the one in AUTHORS?

JohannesGaessler:
>It is not my observation. I get a different result each time I run the code on llama.cpp master (on the same hardware, using the same parameters). The reason for this is that src1 rows end up in a different order in the contiguous copy, which then influence the result as floating point operations are not associative. It is not wrong, but it is still not deterministic. If in doubt, run a few perplexity calculations and observe how the values are different in each run on master, but are always the same with this version of the kernel.

You're right, I misremembered how the code works.

CISC:
I'm reopening a new PR with co-authorship, the address in AUTHORS seems to link correctly.

ikawrakow:
> I'm reopening a new PR with co-authorship, the address in AUTHORS seems to link correctly.

Interesting approach. Because at some point in the past I contributed to this repository, from there on any original work I produce can be freely copied into here, and the attribution is covered by me being included in a list of a few hundred authors with a no-reply address?

CISC:
> Interesting approach. Because at some point in the past I contributed to this repository, from there on any original work I produce can be freely copied into here, and the attribution is covered by me being included in a list of a few hundred authors with a no-reply address?

Only work published under a compatible license of course. The AUTHORS list is generated from git history, which is the only manageable way in a project with so many contributors I guess.

Anyway, I'm more than happy to make any additional attributions you wish, it is your work, and I'm very grateful for it!

PR #1181, opened 2025-04-06
cpu : simd gelu fp32

CISC:
Upstreamed from ikawrakow/ik_llama.cpp#9 (fused soft cap may be added separately later).


ikawrakow:
As the original author of the code in this PR, I'm OK with the PR as submitted.

PR #1187, opened 2025-04-10
ggml: fix #1186 - don't include arm_neon.h when using CUDA 12 with ARM Neon

### LINKED ISSUE #1186: ARM NEON with CUDA 12 fails to build because nvcc tries to compile arm_neon.h
Very small, subtle bug at line 313 in ggml-impl.h macro to disable NVCC:

#if defined(__ARM_NEON) && !(defined(__CUDACC__) && __CUDA_VER_MAJOR <= 11) & !(defined(__MUSACC__))

should be

#if defined(__ARM_NEON) && !(defined(__CUDACC__)) && (__CUDA_VER_MAJOR <= 11) & !(defined(__MUSACC__))

At least, when I changed that  ggml/llama.cpp compiled successfully, where as it was breaking before on 52991e3.

I would do a PR, but I haven't taken the time to learn the rules of committing to this specific project (which I love dearly - great job!)

cmdr2:
`arm_neon.h` wasn't included previously for any version of CUDA, whereas it gets included now for CUDA 12 (which is a bug).

Scenarios:
```
- ARM Neon
 - MSVC:
  - prev: Include header, use uint16, use fallback fp conversion functions
  - current (ok): Include header, use fp16, use actual fp conversion functions
  - expected: Include header, use fp16, use actual fp conversion functions
 - GCC:
  - prev: include header, use fp16, use actual fp conversion functions
  - current (ok): include header, use fp16, use actual fp conversion functions
  - expected: include header, use fp16, use actual fp conversion functions
 - CUDA <= 11
  - prev: don't include header, use uint16, use actual fp conversion functions
  - current (ok): don't include header, use fallback fp conversion functions
  - expected: don't include header, use fallback fp conversion functions
 - CUDA 12
  - prev: don't include header, use fp16, use actual fp conversion functions
  - current (bug!): include header (bug), use fp16, use actual fp conversion functions
  - expected: don't include header, use fp16, use actual fp conversion functions
- Not ARM Neon:
 - prev: don't include header, use other functions
 - current (ok): don't include header, use other functions
 - expected: don't include header, use other functions
```

### DIFF
--- a/src/ggml-impl.h
+++ b/src/ggml-impl.h
@@ -16,6 +16,14 @@
 #include <arm_sve.h>
 #endif // __ARM_FEATURE_SVE
 
+#if defined(__ARM_NEON) && !defined(__CUDACC__) && !defined(__MUSACC__)
+// if YCM cannot find <arm_neon.h>, make a symbolic link to it, for example:
+//
+//   $ ln -sfn /Library/Developer/CommandLineTools/usr/lib/clang/13.1.6/include/arm_neon.h ./src/
+//
+#include <arm_neon.h>
+#endif
+
 #if defined(__F16C__)
 #include <immintrin.h>
 #endif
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Got it. The reason I was confused is because I thought that the nvcc compiler needs to pick `__fp16` from `arm_neon.h`. But it seems that it will have it defined somewhere internally.

ggerganov:
Running the llama.cpp CI just to make sure all is good and will merge.

cmdr2:
I guess it works, since colintoal confirmed on the issue that this fix compiles. I'm not sure how it works tbh, maybe it's an intrinsic data type for nvcc on ARM? Anyway, good to know.

ggerganov:
Yes, tbh I am also not 100% confident I understand how `__fp16` is handled by the `nvcc` compiler and it's different versions.

Btw, another report for this was submitted in the `llama.cpp` repo and the proposed fix is the same: https://github.com/ggml-org/llama.cpp/pull/12872. So it seems this should be good.

PR #1193, opened 2025-04-14
Fix a few small Windows / MSVC build issues

tlemo:
The latest MSVC reports a few compilation problems (#1194)

```
ggml\tests\test-pad-reflect-1d.cpp(28): error C7555: use of designated initializers requires at least '/std:c++20'
ggml\tests\test-pad-reflect-1d.cpp(89): error C7555: use of designated initializers requires at least '/std:c++20'
ggml\tests\test-pad-reflect-1d.cpp(88): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
ggml\tests\test-pad-reflect-1d.cpp(148): error C7555: use of designated initializers requires at least '/std:c++20'
ggml\tests\test-pad-reflect-1d.cpp(147): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```


slaren:
It would be preferable to fix the code to require only C++17, rather than changing the standard which will create more issues with compilers that still do not support C++20.

tlemo:
> It would be preferable to fix the code to require only C++17, rather than changing the standard which will create more issues with compilers that still do not support C++20.

Sure. I updated the changes to fix the MSVC build and stick to C++17.

tests/test-pad-reflect-1d.cpp
```diff
@@ -85,10 +87,10 @@ void test_pad_reflect_1d(bool use_gpu) {
...
     // Test cases for different padding configurations
     {
-        params = (struct ggml_init_params){
-            .mem_size   = 16*1024*1024,
-            .mem_buffer = NULL,
-            .no_alloc   = true,
+        params = ggml_init_params{
+            16*1024*1024,
+            nullptr,
+            true
```
ggerganov, line 93: The pattern that we use is to put the members in comments like this:

```suggestion
        params = ggml_init_params{
            /*.mem_size   =*/ 16*1024*1024,
            /*.mem_buffer =*/ nullptr,
            /*.no_alloc.  =*/ true
```

tlemo: > The pattern that we use is to put the members in comments like this:

Thanks, I applied this to all the instances.

Unfortunately, there's a risk that these kind of comments and the structure will go out-of-sync. Enabling C++20 and using the designated initializers would provide more safety here, and since the set of compile which support C++17 but not C++20 is getting really small, hopefully the upgrade to C++20 will be reconsidered in the future.

ggerganov (approved):

PR #1203, opened 2025-04-29
media : rm logos

ggerganov:
Moved to: https://github.com/ggml-org/media

Generate some images using Bash + ImageMagick

llama cpp-model-qwen-3

llama cpp-feature-cache-reuse


slaren:
I imagine that the goal here is to create images to share on social media to promote ggml and related projects. It's cool to release the source code even for that, but maybe it should be in a separate repository.

PR #1226, opened 2025-05-16
Enable UTF-8 Source Encoding for MSVC via CMake

xiaobing318:
Add conditional CMake compiler options so that when building with MSVC, both C and C++ sources are compiled with UTF-8 encoding (/utf-8). By default, MSVC uses the system code page and will warn (C4819) or fail when encountering non-ASCII characters in source files. This change eliminates those encoding warnings without requiring BOMs or pragmas in the code, ensures consistent UTF-8 handling across platforms, and keeps existing build processes for other toolchains unchanged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,8 +46,6 @@ option(GGML_BACKEND_DL   "ggml: build backends as dynamic libraries (requires BU
 if (MSVC)
     add_compile_options("$<$<COMPILE_LANGUAGE:C>:/utf-8>")
     add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/utf-8>")
-    add_compile_options("$<$<COMPILE_LANGUAGE:C>:/bigobj>")
-    add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/bigobj>")
 endif()
 
 #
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -42,6 +42,12 @@ endif()
 option(BUILD_SHARED_LIBS "ggml: build shared libraries" ${BUILD_SHARED_LIBS_DEFAULT})
 option(GGML_BACKEND_DL   "ggml: build backends as dynamic libraries (requires BUILD_SHARED_LIBS)" OFF)
 
+
+if (MSVC)
+    add_compile_options("$<$<COMPILE_LANGUAGE:C>:/utf-8>")
+    add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/utf-8>")
+endif()
```
slaren, line 49: There is already a `if (MSVC)` block at the bottom of this file, move this there to keep all the MSVC-specific settings together.

xiaobing318: > There is already a `if (MSVC)` block at the bottom of this file, move this there to keep all the MSVC-specific settings together.

Move /utf-8 and /bigobj into the existing if(MSVC) block before add_subdirectory(src) so that every ggml target inherits the flags automatically. This keeps all MSVC-specific settings together while still ensuring ggml-base and ggml see the flags during their creation.

PR #1232, opened 2025-05-19
ggml : Print backtrace on uncaught C++ exceptions

danielzgtg:
The goal is to have what users call "[full logs](https://github.com/ggml-org/whisper.cpp/issues/3168)" contain the backtrace.

This is registered upon ggml_init. Also fixes a minor fd [leak](https://github.com/ggml-org/ggml/pull/1228#discussion_r2094299456) on Linux.

# Before

```text
/home/home/CLionProjects/ggml/cmake-build-debug/bin/simple-ctx
terminate called after throwing an instance of 'std::runtime_error'
  what():  test

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

# After

```text
/home/home/CLionProjects/ggml/cmake-build-debug/bin/simple-ctx
set_terminate:0: uncaught exception test
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
warning: 30	../sysdeps/unix/sysv/linux/wait4.c: No such file or directory
0x00007f1711d1a127 in __GI___wait4 (pid=311685, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
#0  0x00007f1711d1a127 in __GI___wait4 (pid=311685, stat_loc=0x0, options=0, usage=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
30	in ../sysdeps/unix/sysv/linux/wait4.c
#1  0x00007f17124c2601 in ggml_print_backtrace () at /home/home/CLionProjects/ggml/src/ggml.c:199
199	        waitpid(child_pid, NULL, 0);
#2  0x00007f17124c2731 in ggml_abort (file=0x7f171254301e "set_terminate", line=0, fmt=0x7f171254302c "uncaught exception %s") at /home/home/CLionProjects/ggml/src/ggml.c:220
220	    ggml_print_backtrace();
#3  0x00007f17124ee3a0 in ggml_uncaught_exception () at /home/home/CLionProjects/ggml/src/ggml-threading.cpp:20
20	            ggml_abort("set_terminate", 0, "uncaught exception %s", ex.what());
#4  0x00007f17120bb0da in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f17120a5a55 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f17120bb391 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x000055e5756e1763 in main () at /home/home/CLionProjects/ggml/examples/simple/simple-ctx.cpp:101
101	        throw std::runtime_error("test");
[Inferior 1 (process 311684) detached]

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

### DIFF
--- a/src/ggml-threading.cpp
+++ b/src/ggml-threading.cpp
@@ -1,4 +1,5 @@
 #include "ggml-threading.h"
+#include <exception>
 #include <mutex>
 
 std::mutex ggml_critical_section_mutex;
@@ -9,4 +10,21 @@ void ggml_critical_section_start() {
 
 void ggml_critical_section_end(void) {
     ggml_critical_section_mutex.unlock();
+}
+
+GGML_NORETURN static void ggml_uncaught_exception() {
+    if (const std::exception_ptr e{std::current_exception()}) {
+        try {
+            std::rethrow_exception(e);
+        } catch (const std::exception & ex) {
+            ggml_abort("set_terminate", 0, "uncaught exception %s", ex.what());
+        } catch (...) {
+            ggml_abort("set_terminate", 0, "unknown exception");
+        }
+    }
+    ggml_abort("set_terminate", 0, "std::terminate called");
+}
+
+void ggml_uncaught_exception_init() {
+    std::set_terminate(ggml_uncaught_exception);
 }
--- a/src/ggml-threading.h
+++ b/src/ggml-threading.h
@@ -9,6 +9,8 @@ extern "C" {
 GGML_API void ggml_critical_section_start(void);
 GGML_API void ggml_critical_section_end(void);
 
+void ggml_uncaught_exception_init(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -160,6 +160,10 @@ static void ggml_print_backtrace(void) {
     const int parent_pid = getpid();
     const int child_pid = fork();
     if (child_pid < 0) { // error
+#if defined(__linux__)
+        close(lock[1]);
+        close(lock[0]);
+#endif
         return;
     } else if (child_pid == 0) { // child
         char attach[32];
@@ -167,6 +171,7 @@ static void ggml_print_backtrace(void) {
 #if defined(__linux__)
         close(lock[1]);
         (void) !read(lock[0], lock, 1);
+        close(lock[0]);
 #endif
         // try gdb
         execlp("gdb", "gdb", "--batch",
@@ -215,6 +220,8 @@ void ggml_abort(const char * file, int line, const char * fmt, ...) {
     ggml_print_backtrace();
     abort();
 }
+
+// ggml_abort is registered with std::set_terminate by ggml-threading.cpp
 
 //
 // logging
@@ -1417,6 +1424,7 @@ struct ggml_context * ggml_init(struct ggml_init_params params) {
     if (is_first_call) {
         // initialize time system (required on Windows)
         ggml_time_init();
+        ggml_uncaught_exception_init();
 
         for (int i = 0; i < (1 << 16); ++i) {
             union {


slaren:
This could be useful, but I don't think it is a good idea to do this in a library since it would prevent users from setting their own terminate. Not sure if there is a way to do this automatically and safely, maybe check if `set_terminate` does not return null and restore the previous handlers if so?

WilliamTambellini:
Just agreeing with Diego that for production onprem deployment, we actually dont want any third party to print the stack.

danielzgtg:
> set_terminate ... null

@slaren GCC seems to register its own nonnull handler. I changed my code to call it afterwards so as to not completely overwrite it.

It's only set once so users are free to overwrite our handler or set their own. Us doing this would be just like how a lot of Python ML projects let their dependencies enable [rich](https://pypi.org/project/rich/)'s traceback.

> dont want any third party to print the stack. We also dont want a third to abort

@WilliamTambellini This only makes C++ match the behavior of `ggml_abort` in C, which prints the stack. [Your pending PR](https://github.com/ggml-org/ggml/pull/1110/files#diff-034dc72828ab28a69045ff244db29850a5d41a5226f5ed530f209b02df1ff208) added code containing the C macros `GGML_ASSERT` and `GGML_ABORT`. How should not printing the stack be done on the C side?

slaren:
I am thinking that it may be preferable to do this in a static initializer so that it is available before calling `ggml_init`, and also by doing it early the chance of overwriting the user's handler is reduced. It would also be better to add a new C++ file to this rather than reusing `ggml-threading.cpp`, since this has nothing to do with threading. A file named `ggml.cpp` should be ok.

danielzgtg:
> static initializer

Do any third-party users use `dlopen` on us?

To lower the risk and address WilliamTambellini's concerns, I'm thinking of just making this opt-in. Then there can be follow-up PRs for llama.cpp and whisper.cpp.

> better to add a new C++ file

I agree

danielzgtg:
I only use C/C++ with ggml. Exactly, dlopen calls the static initializers, and I don't know the effects in other languages. The call might feel more "random" and harder to debug than a `ggml_init`, and it's worse if the dlopen is in generated code.

Since nobody from other languages has spoken up, I'll start implementing [your suggestion](https://github.com/ggml-org/ggml/pull/1232#issuecomment-2914458882), and leave hypothetical problems for later.

slaren:
Also, we should check for the environment variable `GGML_NO_BACKTRACE` and avoid setting the handler if it is set.

slaren (approved):

ggerganov:
This seems to cause a build error here: https://github.com/ggml-org/whisper.cpp/actions/runs/15318203014/job/43096006624#step:11:128

rotemdan:
The assertion `GGML_ASSERT(prev != ggml_uncaught_exception)` in the static initializer is now causing a runtime error:
```
GGML_ASSERT(prev != ggml_uncaught_exception) failed
```

Whenever `llama.cpp` (consequently `ggml`) dynamic library, like `libllama.so` is being:
* Dynamically loaded, like with `dlopen` or `LoadLibrary`.
* Unloaded with `dlclose` or `FreeLibrary`
* Loaded again, causing the assertion to fail.

This prevents switching between different builds at runtime (like switching between a CPU and a CUDA build).

Other than removing the assertion, a solution proposed by Gemini was to use `std::once_flag` and `std::call_once` to ensure the handler is only set once, instead of the assertion.

rotemdan:
Another related issue is that if `ggml` is loaded as a dynamic library (possibly via `llama.cpp`) and via `dlopen` or `LoadLibrary` then unloaded by `dlclose` or `FreeLibrary`, then this code:

```cpp
    const auto prev{std::get_terminate()};
    GGML_ASSERT(prev != ggml_uncaught_exception);
    previous_terminate_handler = prev;
    std::set_terminate(ggml_uncaught_exception);
```

The current handler `ggml_uncaught_exception` may become a dangling pointer, since it's in a memory space that has been deallocated when the library was unloaded.

Now some really surprising thing I noticed, either before or after loading the `ggml` library, compiled on MSVC / Windows 11 is that calls to:

```cpp
auto r = std::get_terminate();
printf("%p\n", r);
```

Always show:
```
0000000000000000
```

Meaning that even if I call 
```cpp
std::set_terminate(nullptr)
```

after unloading the library, it has no effect. The assertion error still occurs, because apparently `prev` is always 0.

So basically on Windows, something doesn't seem to be right with this approach. Maybe it should be reconsidered. I don't have the technical knowledge needed to explain why it's `get_terminate` is returning 0 there.

## Update

After researching further, `std::get_terminate()` is not always returning `null` under a simple MSVC console application. 

The null return values seems to maybe have to do with the particular Node.js / V8 environment I'm compiling into, for a N-API addon (involving `node-gyp`). It could be that one of the options it passes is modifying the default behavior of runtime exception handling.

rotemdan:
I'll try to summarize what I have so far:

* The current approach for setting a global `terminate_handler` via `set_terminate` is only safe for either static linking, or dynamic linking at compile-time
* If the `ggml` dynamic library is loaded and unloaded at runtime, then both the function `ggml_uncaught_exception` and `previous_terminate_handler` would become dangling pointers, since they are pointing to a memory space that has been deallocated
* The second time `ggml` is loaded, it will always fail the assertion `GGML_ASSERT(prev != ggml_uncaught_exception)`
* Even if `std::set_terminate(nullptr)` is called after unloading the library (given the user has deeply researched the code, and became informed of the behavior of `set_terminate`), there's no way to reset `previous_terminate_handler`, meaning that it will remain a dangling pointer, though the way the code is written currently, it will not be accessed until `ggml_uncaught_exception_init` is called again
* On some platforms, like when compiling as a Node.js N-API addon (on Windows and possibly other platforms), `std::set_terminate` doesn't do anything and `std::get_terminate()` always returns `nullptr`, so the static initializer doesn't actually apply any handler, but the assertion still consistently fails, although I haven't been able to reason why
* Setting `GGML_NO_BACKTRACE` environment variable could technically allow bypassing the static initializer, but setting an environment variable adds complexity and is not desirable practical to use in some situations (for example, when deploying a library that uses `ggml` as a dynamically-linked dependency)

My suggestion is to allow to disable the static initializer via a preprocessor directive, to make it conditionally compiled.

The only practical options I see currently to bypass this is to fork the code to add the directive or disable the static initializer completely.

rotemdan:
Gemini has suggested an alternative solution.

Basically, we can wrap the code in a class, and put an instance of it in a static variable, such that once the library is unloaded, the destructor would be called, which will prevent the dangling pointers and other unsafe situations. 

Here is the code Gemini 2.5 Pro generated, after a few iterations:

```cpp
#include "ggml-impl.h"

#include <cstdlib>
#include <exception>

// Use an anonymous namespace to keep the implementation private to this file.
namespace {

/**
 * @brief An RAII class to safely install and uninstall a custom terminate handler.
 *
 * A single static instance of this class is created at library load time.
 * Its constructor sets a custom terminate handler to print a backtrace.
 * Its destructor, called automatically on library unload, restores the original handler.
 * This prevents dangling function pointers if the library is dynamically unloaded and reloaded.
 */
class GgmlTerminateHandler {
public:
    GgmlTerminateHandler() {
        // Do not install the handler if the user has opted out.
        if (getenv("GGML_NO_BACKTRACE")) {
            return;
        }

        // Store the existing handler.
        s_previous_handler = std::get_terminate();

        // If our handler is already installed, do nothing. This prevents issues
        // on library reload where the previous instance might not have been
        // destructed yet in some complex scenarios.
        if (s_previous_handler == custom_terminate_handler) {
            return;
        }

        std::set_terminate(custom_terminate_handler);
    }

    ~GgmlTerminateHandler() {
        // Restore the previous handler only if our handler is still active.
        // This is a safeguard in case another part of the program has
        // replaced the terminate handler after us.
        if (std::get_terminate() == custom_terminate_handler) {
            std::set_terminate(s_previous_handler);
        }
    }

private:
    // The custom handler that will be called on std::terminate().
    GGML_NORETURN static void custom_terminate_handler() {
        ggml_print_backtrace();
        if (s_previous_handler) {
            s_previous_handler();
        }
        abort(); // Fallback in case the previous handler returns.
    }

    // The previously registered terminate handler.
    // Must be static to be accessible from the static custom_terminate_handler.
    static std::terminate_handler s_previous_handler;
};

// Define the static member variable.
std::terminate_handler GgmlTerminateHandler::s_previous_handler = nullptr;

// The single static instance that manages the handler's lifetime.
// Its constructor is called on library load, and its destructor on unload.
static GgmlTerminateHandler g_ggml_terminate_handler_instance;

} // namespace
```

This can be a place to start. I'll work from this and give a pull request once I ensure it's correct.

## Edit: Why this approach is still not fully safe and why 100% safety is likely nearly impossible

When the initializer starts, if the previous termination handler belongs to a dynamic library, and later that library is unloaded, it will become a dangling pointer, that is possibly going to crash the application when called via `s_previous_handler()`.

Same problem occurs when trying to open multiple instances of the `ggml` library concurrently (say one for `cpu` build and one for `cuda` build). If one unloads, the other still calls its handler, in its own handler.

For these reasons I'd suggest to also add a preprocessor directive that would allow to completely enable / disable this feature.

danielzgtg:
@rotemdan What is your use case for dlclose? Many libraries don't support unloading.

rotemdan:
Node.js addon, with support to loading multiple versions of the llama.cpp shared libraries, compiled for different backends, side-by-side on the same process, or switch between backends at runtime (which requires unloading of the shared libraries to do cleanly).

The wrapper is structured as a class that can be loaded and unloaded. 

Currently, the only thing that seems to prevent GGML and llama.cpp to allow full runtime dynamic linking and unlinking is the assertion added in this pull request. Otherwise it seems to work correctly (tested with the modified code by Gemini).

(Also, I'm not 100% sure if I remember correctly, but the failing assertion is not just failing during unloading (dlclose). It's failing when loading a second version of the shared library even without closing the first one)

danielzgtg:
> switch between backends at runtime

I wrote https://github.com/danielzgtg/ggml/tree/feat/backtraceDlClose based on the destructor idea. Does it work for this?

> second version of the shared library

Do the Vulkan/CUDA libraries support this?

PR #1237, opened 2025-05-23
Add support for google-benchmark

lunochod:
This PR aims to add initial support for Google Benchmark to GGML, which could lead to building a suite of benchmarks complementing the unit tests. This will allow contributors to work on GGML's performance and the various backends and maintainers to test for changes in performance over time.

I would also like to use the PR to discuss keeping external GGML dependencies small while including standard tools for benchmarking, testing, and logging.


lunochod:
@ggerganov Any feedback?

ggerganov:
Thanks. We already have a tool for benchmarking the performance of the ops: `test-backend-ops` and it's more generic since it works with all ggml ops.

PR #1240, opened 2025-05-27
ggml : install dynamic backends

rgerganov:
Make sure dynamic backends are installed in $CMAKE_INSTALL_LIBDIR


slaren:
The problem is that dl backends cannot be loaded from the lib dir, since the loader only checks the current directory and the directory of the executable.

rgerganov:
We can also load the backend from `$GGML_BACKEND_PATH`, so we can instruct users to set `GGML_BACKEND_PATH=$INSTALL_DIR/lib/backend.so`. Another option for ggml developers is to use
`ggml_backend_load_all_from_path(const char * dir_path)` and set `dir_path` to `$INSTALL_DIR/lib`. In both cases we just need backends to be installed on the file system, somewhere under `$INSTALL_DIR` (not necessarily in `$INSTALL_DIR/lib`)

slaren (approved):

PR #1243, opened 2025-05-28
ggml-cpu : Add GGML_OPENMP_FFAST_MATH for sine autovectorization

danielzgtg:
End-user time savings of 12.99% was observed on a previous ggml version. This was upstreamed from my work at https://github.com/mmwillet/ggml/pull/2 . I only need sine but cos is included for symmetry.

GCC with OpenMP supports this optimization. This is opt-in as libmvec's output differs from libm.

# Before

ggml_compute_forward_sin wasn't vectorized at all. When calculating sine, it suffered the overhead of a call for each element. During profiling using an outdated fork, each GGML_OP_SIN was ~5–10 ms.

```console
$ nm libggml-cpu.so | grep sinf
                 U sinf@GLIBC_2.2.5
```

# After

Now it works in blocks of 4 or 8 elements at a time and approximates more roughly. During profiling using an outdated fork, each GGML_OP_SIN was ~3 ms.

```console
$ nm libggml-cpu.so | grep sinf
                 U sinf@GLIBC_2.2.5
                 U _ZGVbN4v_sinf@GLIBC_2.22
                 U _ZGVdN8v_sinf@GLIBC_2.22
```


slaren:
Does this have any advantages over AVX intrinsics such as `_mm256_sin_ps`?

danielzgtg:
_mm256_sin_ps is [not available](https://stackoverflow.com/a/31979341/10477326) on any compiler besides Intel

danielzgtg:
Here is my plan:
* Use `_ZGVdN8v_sinf` on GCC and Clang on Linux. It's part of the Linux-only glibc
* Use `_mm256_sin_ps` with the Intel compiler on any platform.
* Otherwise, use a lookup table if the flag (will rename to `GGML_CPU_FAST_TRIG`) is still specified
* Undo my splitting into multiple files

`#pragma openmp simd` is unnecessary despite being related to [where I got this idea from](https://sourceware.org/glibc/wiki/libmvec). Removing the pragma in the example and only adding the ffast-math worked equally well.

Draftifying this until I get to rebasing the downstream fork. Using intrinsics requires more testing than compiler flags.

PR #1248, opened 2025-05-29
ggml : fix pkg-config include path

rgerganov:
CMake is installing public headers in CMAKE_INSTALL_INCLUDEDIR, not CMAKE_INSTALL_INCLUDEDIR/ggml.

This patch fixes building external programs which use 'pkg-config --cflags ggml'


ggerganov (approved):

rgerganov:
+1 for switching to `#include <ggml/ggml.h>`

PR #1255, opened 2025-05-31
Allow stand-alone build of simple example

dg0yt:
This helps to validate the installed package.

It also shows a benefit of the other change: An in-build `ggml::ggml` ALIAS enables a uniform link library name regardless of ggml being used as a sub-project or via `find_package`.

### DIFF
--- a/examples/simple/CMakeLists.txt
+++ b/examples/simple/CMakeLists.txt
@@ -1,16 +1,30 @@
 #
+# This example can be built as part of ggml or
+# as a separate project with an installed ggml.
+
+cmake_minimum_required(VERSION 3.14)
+project(ggml-simple)
+
+if (NOT TARGET ggml::ggml)
+    # Building as a separate project
+    find_package(ggml CONFIG REQUIRED)
+
+    set(CMAKE_CXX_STANDARD 17)
+endif()
+
+#
 # simple-ctx
 
 set(TEST_TARGET simple-ctx)
 add_executable(${TEST_TARGET} simple-ctx.cpp)
-target_link_libraries(${TEST_TARGET} PRIVATE ggml)
+target_link_libraries(${TEST_TARGET} PRIVATE ggml::ggml)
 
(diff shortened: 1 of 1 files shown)


dg0yt:
And the next step is to make CMake config actually work.
Preferably by installing exported configuration.

ggerganov (approved):

slaren (commented):
I am not sure about this. It would be ok to have an example specifically for this purpose, but just randomly changing one of the examples to be a separate project, without any documentation or explanation in the code, is just going to be confusing.

ggerganov:
Maybe for now just add the `ggml::ggml` alias and don't change the example?

dg0yt:
Well, I just use this example to validate the installed cmake config (and in a variation, pkgconfig) of the vcpkg port. Your CI could do the same.

I explicitly tried to find a simple existing example. The additions should not cause much confusion - It is really just very basic CMake, nicely illustrating expected basic usage.

dg0yt:
> Your CI could do the same.

Proof of concept: https://github.com/dg0yt/ggml/pull/1

dg0yt:
- There is no change for in-build experience.
- There is a an extra comment explaining that it can be used in two ways.
- There is a proof of concept how to use it for ggml CI.
  (And it includes fixes for relevant problems.)

What else would be needed to remove concerns?

dg0yt:
Hm, I expected this to be a no-brainer. I wanted this to be merged before updating the ggml-org ports in vcpkg.

dg0yt:
... and before upstreaming https://github.com/dg0yt/ggml/pull/1.

ggerganov:
Create a new very trivial example called `test-cmake` and update its CMakeLists.txt for this purpose. The implementation in `test-cmake.cpp` can just call `ggml_backend_load_all();` and exit.

dg0yt:
Valid, but obsolete.

PR #1260, opened 2025-06-03
Add in-build ggml::ggml ALIAS library

dg0yt:
Enable uniform linking with subproject and with find_package.


dg0yt:
Split from #1255.

slaren (approved):

PR #1263, opened 2025-06-06
sam : add default case for unsupported prompt types

danbev:
This commit adds a default case in the `embed_prompt_sparse` lambda function.

The motivation for this change is that currently the following warning is generated when compiling:
```console
/ggml/examples/sam/sam.cpp: In lambda function:
/ggml/examples/sam/sam.cpp:1499:5: warning:
control reaches end of non-void function [-Wreturn-type]
 1499 |     }();
      |     ^
```

### DIFF
--- a/examples/sam/sam.cpp
+++ b/examples/sam/sam.cpp
@@ -1460,7 +1460,7 @@ prompt_encoder_result sam_encode_prompt(
     ggml_set_name(inp, "prompt_input");
     ggml_set_input(inp);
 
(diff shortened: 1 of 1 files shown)


danbev:
@ggerganov Would it be alright to give me permission/rights to commit to this repository so I can merge PRs?

ggerganov:
Done!

danbev:
@ggerganov Thanks!

PR #1273, opened 2025-06-13
ggml : disable warnings for tests when using MSVC

danbev:
This commit disables warnings for tests on windows when using MSVC.

The motivation for this is that this brings the build output more inline with what Linux/MacOS systems produce.

There is still one warning generated for the tests which is:
```console
  Building Custom Rule C:/ggml/tests/CMakeLists.txt
cl : command line  warning D9025: overriding '/DNDEBUG' with '/UNDEBUG' [C:\ggml\build\tests\test-arange.vcxproj]
  test-arange.cpp
  test-arange.vcxproj -> C:\ggml\build\bin\Release\test-arange.exe
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,6 +367,7 @@ if (MSVC)
         /wd4005  # Macro redefinition
         /wd4244  # Conversion from one type to another type, possible loss of data
         /wd4267  # Conversion from 'size_t' to a smaller type, possible loss of data
+        /wd4305  # Conversion from 'type1' to 'type2', possible loss of data
         /wd4566  # Conversion from 'char' to 'wchar_t', possible loss of data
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -413,4 +414,20 @@ if (MSVC)
         disable_msvc_warnings(simple-ctx)
         disable_msvc_warnings(simple-backend)
     endif()
+
+    if (GGML_BUILD_TESTS)
+        disable_msvc_warnings(test-mul-mat)
+        disable_msvc_warnings(text-arange)
```
ggerganov, line 420:
```suggestion
        disable_msvc_warnings(test-arange)
```

ggerganov (approved):

PR #1274, opened 2025-06-13
Add `ggml_roll`

Acly:
This adds `ggml_roll` ([torch.roll](https://docs.pytorch.org/docs/stable/generated/torch.roll.html)) as a new operation. It shifts tensor elements with wrap behavior.

It is used by eg. SWIN transformer to implement shift for its window attention.

### DIFF
--- a/include/ggml.h
+++ b/include/ggml.h
@@ -489,6 +489,7 @@ extern "C" {
         GGML_OP_UPSCALE, // nearest interpolate
         GGML_OP_PAD,
         GGML_OP_PAD_REFLECT_1D,
+        GGML_OP_ROLL,
         GGML_OP_ARANGE,
         GGML_OP_TIMESTEP_EMBEDDING,
         GGML_OP_ARGSORT,
@@ -1800,6 +1801,17 @@ extern "C" {
             struct ggml_tensor  * a,
             int                   p0,
             int                   p1);
+
+    // Move tensor elements by an offset given for each dimension. Elements that
+    // are shifted beyond the last position are wrapped around to the beginning.
+    GGML_API struct ggml_tensor * ggml_roll(
+            struct ggml_context * ctx,
+            struct ggml_tensor  * a,
+            int                   shift0,
+            int                   shift1,
+            int                   shift2,
+            int                   shift3);
+
 
     // Ref: https://github.com/CompVis/stable-diffusion/blob/main/ldm/modules/diffusionmodules/util.py#L151
     // timesteps: [N,]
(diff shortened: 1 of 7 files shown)


src/ggml-cpu/ops.cpp
```diff
@@ -6793,6 +6793,73 @@ void ggml_compute_forward_pad_reflect_1d(
...
+
+    const int s0 = dst->op_params[0];
+    const int s1 = dst->op_params[1];
+    const int s2 = dst->op_params[2];
+    const int s3 = dst->op_params[3];
+
+    const int64_t total = ne1 * ne2 * ne3;
+    const int64_t per_thread = (total + params->nth) / params->nth;
+    const int64_t start = params->ith * per_thread;
+    const int64_t end   = MIN(start + per_thread, total);
```
slaren, line 6825: `std::min` would be better, the `MIN` macro is a leftover from the time this code was compiler as C instead of C++.

include/ggml.h
```diff
@@ -1801,6 +1802,17 @@ extern "C" {
...
+    // Move tensor elements by an offset given for each dimension. Elements that
+    // are shifted beyond the last position are wrapped around to the beginning.
+    GGML_API struct ggml_tensor * ggml_roll(
+            struct ggml_context * ctx,
+            struct ggml_tensor  * a,
+            int                   shift0,
+            int                   shift1,
+            int                   shift2,
+            int                   shift3);
+
```
ggerganov, line 1814: Should we change the signature like so:

```suggestion
    // Move tensor elements by an offset given for each dimension. Elements that
    // are shifted beyond the last position are wrapped around to the beginning.
    // If tensor b is provided, the shifts become: shiftX' = shiftX + b[X]
    GGML_API struct ggml_tensor * ggml_roll(
            struct ggml_context * ctx,
            struct ggml_tensor  * a,
            struct ggml_tensor  * b, // optional I64 [GGML_MAX_DIMS]
            int                   shift0,
            int                   shift1,
            int                   shift2,
            int                   shift3);

```

This in theory would allow to pass the shifts as dynamic graph input data, which makes the graph static in terms of nodes (and op params). Could be useful for graph reuse.

slaren: I wouldn't bother with this unless there is some use case that you are thinking about, otherwise this is likely to end increasing complexity for no real reason. If you decide to add this, for clarity I would recommend adding a different function that takes only the tensor, rather than trying to do everything with a single function and optional parameters.

ggerganov (approved):

PR #1276, opened 2025-06-16
ggml-cpu : remove stdlib include from repack.cpp

danbev:
This commit removes the inclusion of `<cstdlib>`.

The motivation for this change is that this source file does not seem to use any functions from this header and the comment about `qsort` is a little misleading/confusing.

### DIFF
--- a/src/ggml-cpu/repack.cpp
+++ b/src/ggml-cpu/repack.cpp
@@ -14,7 +14,6 @@
 #include <cmath>
 #include <cstring>
 #include <cassert>
-#include <cstdlib> // for qsort
 #include <cstdio>  // for GGML_ASSERT
 
 #include "repack.h"


ggerganov:
There will be some more refactoring of this code soon (https://github.com/ggml-org/llama.cpp/issues/14134#issuecomment-2976659333) and after we sync them from `llama.cpp` we can merge this change, to avoid conflicts.

danbev:
> There will be some more refactoring of this code soon ([ggml-org/llama.cpp#14134 (comment)](https://github.com/ggml-org/llama.cpp/issues/14134#issuecomment-2976659333)) and after we sync them from `llama.cpp` we can merge this change, to avoid conflicts.

I believe the sync was been [done](https://github.com/ggml-org/ggml/commit/eb6c858f7fb1c3226656f0b2b01f529a7c6b6914) so perhaps we can merge this now (just needs an approval).

PR #1280, opened 2025-06-20
tests: Enable missing tests

ckastner:
Just noticed by chance. If this is intentional, feel free to insta-close.

On my end, test time increased from ~11sec to ~17sec.


slaren:
These look like very old tests that may no longer be relevant to ggml.

ckastner:
OK, closing then. If you want an MR to remove the test, let me know.

PR #1285, opened 2025-06-24
ggml-cpu : "align corners" for bilinear upscale/downscale

Acly:
This PR
* adds `GGML_SCALE_MODE_BILINEAR_ALIGN_CORNERS` for `ggml_upscale`, which behaves consistent to Torch's `interpolate(mode='bilinear', align_corners=True)`
* allows using `ggml_upscale` to downscale too (no change in implementation)
* adds some tests

I'm not sure what the initial reason was for not allowing downscale, maybe because "nearest" was the only option and that can be accomplished with strides? In any case, for bilinear interpolation it's useful and not uncommon.

It does make the naming and API a bit awkward, if you think it makes sense to rename or add an additional function I'm happy to do that.

### DIFF
--- a/include/ggml.h
+++ b/include/ggml.h
@@ -1763,14 +1763,14 @@ extern "C" {
             float                 p1);
 
     enum ggml_scale_mode {
-        GGML_SCALE_MODE_NEAREST,
-        GGML_SCALE_MODE_BILINEAR,
+        GGML_SCALE_MODE_NEAREST  = 0,
+        GGML_SCALE_MODE_BILINEAR = 1,
 
         GGML_SCALE_MODE_COUNT
     };
 
     enum ggml_scale_flag {
-        GGML_SCALE_ALIGN_CORNERS = (1 << 8)
+        GGML_SCALE_FLAG_ALIGN_CORNERS = (1 << 8)
     };
 
     // interpolate
--- a/src/ggml-cpu/ops.cpp
+++ b/src/ggml-cpu/ops.cpp
@@ -6636,7 +6636,7 @@ static void ggml_compute_forward_upscale_f32(
         }
     } else if (mode == GGML_SCALE_MODE_BILINEAR) {
         float pixel_offset = 0.5f;
-        if (mode_flags & GGML_SCALE_ALIGN_CORNERS) {
+        if (mode_flags & GGML_SCALE_FLAG_ALIGN_CORNERS) {
             pixel_offset = 0.0f;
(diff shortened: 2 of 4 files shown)


slaren:
> It does make the naming and API a bit awkward, if you think it makes sense to rename or add an additional function I'm happy to do that.

Maybe it would be good to mark `ggml_upscale_ext` as deprecated, and add `ggml_interpolate` as an alias. cc @ngxson 

Also, if we are going to add a new function, I would prefer if `align_corners` becomes a separate parameter rather than part of the mode enum.

Acly:
> I would prefer if align_corners becomes a separate parameter rather than part of the mode enum

Any particular reason? The enum is perhaps a bit uncomfortably long, but between 
```
ggml_interpolate(..., GGML_SCALE_MODE_BILINEAR_ALIGN_CORNERS)
ggml_interpolate(..., GGML_SCALE_MODE_BILINEAR, true)
```
I prefer the enum at callsites. It also prevents combining align-corners with modes that don't support it.

slaren:
Because fundamentally they are different parameters, and bundling every combination of parameters into an enum can very quickly become unmaneagable. A bitfield would also work, e.g. `GGML_SCALE_MODE_BILINEAR | GGML_SCALE_MODE_ALIGN_CORNERS`.

Acly:
Changed align-corners parameter to a separate bit flag, and added `ggml_interpolate`.

I didn't rename the OP and all the implementations in various backends, not sure if that's worth the hassle.

include/ggml.h
```diff
@@ -1763,8 +1763,14 @@ extern "C" {
             float                 p1);
 
     enum ggml_scale_mode {
-        GGML_SCALE_MODE_NEAREST  = 0,
-        GGML_SCALE_MODE_BILINEAR = 1,
+        GGML_SCALE_MODE_NEAREST,
+        GGML_SCALE_MODE_BILINEAR,
```
slaren, line 1767: Don't remove the enum value integers, they are there to facilitate interop.

include/ggml.h
```diff
@@ -1763,8 +1763,14 @@ extern "C" {
...
+
+        GGML_SCALE_MODE_COUNT
+    };
+
+    enum ggml_scale_flag {
+        GGML_SCALE_ALIGN_CORNERS = (1 << 8)
```
slaren, line 1773: Enum values are always prefixed with the full enum name
```suggestion
    enum ggml_scale_flag {
        GGML_SCALE_FLAG_ALIGN_CORNERS = (1 << 8)
```

PR #1286, opened 2025-06-27
ggml : add version function to get lib version

danbev:
This commit adds a function `ggml_version()` to the ggml library that returns the version of the library as a string.

The motivation for this is that it can be useful to be able to programmatically check the version of the ggml library being used.

Usage:
```c
printf("GGML version: %s\n", ggml_version());
```
Output:
```console
GGML version: 0.0.2219
```


slaren:
If this is extended with a function that returns the commit id, it could be useful for https://github.com/ggml-org/llama.cpp/pull/14368

Issue #1296, opened 2025-07-10
ARM compilation error for SME


peardox:
I've tried this on a Ampere Altra and Pi5

```
cmake -B build -DGGML_CPU_ARM_ARCH=OFF -DGGML_NATIVE=OFF -DGGML_CPU_ALL_VARIANTS=ON -DGGML_BACKEND_DL=ON -DBUILD_SHARED_LIBS=ON
cmake --build build

```

Results in,,,

> cc1: error: unknown value ‘armv9.2-a+dotprod+fp16+sve+i8mm+sve2+sme’ for ‘-march’
> cc1: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a armv8.6-a armv8.7-a armv8.8-a armv8-r armv9-a native
> gmake[2]: *** [ggml/src/CMakeFiles/ggml-cpu-armv9.2_2.dir/build.make:76: ggml/src/CMakeFiles/ggml-cpu-armv9.2_2.dir/ggml-cpu/ggml-cpu.c.o] Error 1
> gmake[1]: *** [CMakeFiles/Makefile2:1517: ggml/src/CMakeFiles/ggml-cpu-armv9.2_2.dir/all] Error 2
> 

Suggest a new switch is added to CMakeLists.txt ...

`option(GGML_SKIP_ARM9 "Skip ARM 9 compilation" OFF)
`

This would then be triggered later on as...

```
if(NOT GGML_SKIP_ARM9)
    ggml_add_cpu_backend_variant(armv9.2_1    DOTPROD FP16_VECTOR_ARITHMETIC SVE MATMUL_INT8 SME)
    ggml_add_cpu_backend_variant(armv9.2_2    DOTPROD FP16_VECTOR_ARITHMETIC SVE MATMUL_INT8 SVE2 SME)
endif()
```

I'll write a PR if nobody else does...

ggerganov:
If the compiler does not support some variants I think you should not use `-DGGML_CPU_ALL_VARIANTS=ON` in the first place.

peardox:
By extension no compiler should  use GGML_CPU_ALL_VARIANTS as none currently support any as-yet unreleased opcodes that will be introduced over the coming years (and eventually end up in the compiler)

Also, by that argument, MSVC should not be used as it does not support AMX

>   if (NOT MSVC)
>       # MSVC doesn't support AMX
[...]

PR #1297, opened 2025-07-10
Add optional -DGGML_SKIP_ARM9 switch

peardox:
This only affects very specific shared library ARM dynamically loaded backend builds

See #1296


peardox:
Changed logic to specifically check for GCC < 14 and if so not build SME variants

peardox:
Better solution discussed in #1296

PR #1312, opened 2025-07-24
contrib : recommend PRs to llama.cpp

ggerganov:
Merging core changes only from `llama.cpp` will make synchronization and resolving conflicts much simpler. Adding recommendations for contributors to move their PRs to https://github.com/ggml-org/llama.cpp

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,1 @@
+*Note: for `ggml`-related changes (including to the CMake build system), please open a PR in https://github.com/ggml-org/llama.cpp. Doing so will make your PR more visible and more likely to be reviewed.*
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1 @@
+*Note: for `ggml`-related changes (including to the CMake build system), please open a PR in https://github.com/ggml-org/llama.cpp. Doing so will make your PR more visible and more likely to be reviewed.*
```
slaren, line 1: The wording of "ggml-related changes" could be more clear, everything in this repository can be understood as being related to ggml. Maybe something like "changes to the ggml core library"?

dg0yt:
I see the recommendation, but maintaining a separate port for ggml in vcpkg, patches are developed and tested for this repo. Barriers for upstreaming patches are ... barriers.

The vcpkg port of ggml could switch to the vendored sources in llama.cpp. But this would probably need some kind of recommendation from upstream. It has an effect on the appropriate version scheme.

ggerganov:
I'll handle your PRs to the `ggml` repo as we made this change in the middle of your work.

We will probably have to make some reorganization of the projects soon, so this won't remain for very long.

PR #1313, opened 2025-07-24
Fix ggml-vulkan 32 bit builds

dg0yt:
The pipeline member can be cast to VkPipeline.
This is a VkPipeline_T* on 64 bit but a uint64_t on 32 bit. Cf. VK_DEFINE_NON_DISPATCHABLE_HANDLE documentation.

### DIFF
--- a/src/ggml-vulkan/ggml-vulkan.cpp
+++ b/src/ggml-vulkan/ggml-vulkan.cpp
@@ -1256,7 +1256,7 @@ static void ggml_vk_create_pipeline_func(vk_device& device, vk_pipeline& pipelin
         vk::DebugUtilsObjectNameInfoEXT duoni;
         duoni.objectType = vk::ObjectType::ePipeline;
         duoni.pObjectName = pipeline->name.c_str();
-        duoni.objectHandle = reinterpret_cast<uint64_t>(static_cast<VkPipeline_T*>(pipeline->pipeline));
+        duoni.objectHandle = reinterpret_cast<uint64_t>(static_cast<VkPipeline>(pipeline->pipeline));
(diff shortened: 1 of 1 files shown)


dg0yt:
Fixes FTBFS for vcpkg ~x86-windows and~ arm-neon-android.

Not fixing x86-windows:
~~~
D:\b\ggml\src\e98bd99269-dabfc489e2.clean\src\ggml-vulkan\ggml-vulkan.cpp(1343): error C2440: 'reinterpret_cast': cannot convert from 'VkPipeline' to 'uint64_t'
D:\b\ggml\src\e98bd99269-dabfc489e2.clean\src\ggml-vulkan\ggml-vulkan.cpp(1343): note: Conversion is a valid standard conversion, which can be performed implicitly or by use of static_cast, C-style cast or function-style cast
~~~

0cc4m (approved):
This makes sense to me and still compiles fine on 64-bit. I can't easily check 32-bit.

0cc4m:
Hmm, I missed your post about the Windows issue. Maybe @mtavenrath (who wrote this initially) or @jeffbolznv could give it a try on Windows?

Acly:
Replacing `reinterpret_cast<uint64_t>` with c-style cast `(uint64_t)` seems like the easiest way that works in all cases

jeffbolznv:
I got the same error as https://github.com/ggml-org/ggml/pull/1313#issuecomment-3114918667. Changing the reinterpret cast to `(uint64_t)` fixed it.

dg0yt:
> Changing the reinterpret cast to `(uint64_t)` fixed it.

I will add this change later. 
I might insert a `static_assert(sizeof(VkPipeline) == sizeof(uin64_t))` to prevent surprises.

dg0yt:
No static_assert added, but `/*reinterpret_cast*/` for those who occasionally grep for this keyword.

PR #1317, opened 2025-07-29
fix dequant_funcs.comp

tophyr:
`data_a` is decorated with layout specifiers, so copying it causes the destination to receive those layout specifiers as well. this is illegal for function-local arrays. `spirv` accepts this code for compilation, but the resulting bytecode is against the rules so `spirv-opt` barfs.

### DIFF
--- a/src/ggml-vulkan/vulkan-shaders/dequant_funcs.comp
+++ b/src/ggml-vulkan/vulkan-shaders/dequant_funcs.comp
@@ -442,8 +442,12 @@ vec2 get_dm(uint ib, uint a_offset) {
 
 #if defined(DATA_A_IQ1_M)
 vec2 get_dm(uint ib, uint a_offset) {
-    const uint16_t[4] scales = data_a[a_offset + ib].scales;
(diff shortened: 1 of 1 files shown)


tophyr:
Hi @ggerganov! It's been a pretty hot minute since I submitted this, so I'll need to page some context back in to update it if necessary - but is this now closed _because_ it's unnecessary? If I recall correctly, the origin of this PR was that the vcpkg port of `ggml` required some patching in order to build against the then-latest version of vulkan tooling: The newer tooling enforced a syntax change that eliminated some UB or something.

ggerganov:
Hi, I noticed the source file has changed and assumed this is outdated.

@0cc4m Could you chime in if this change is still needed?

tophyr:
lemme see if i can update the PR and provide better repro steps with `spirv` version as well.

tophyr:
It is indeed still necessary: Building `ggml` with `spirv-tools 1.4.341.0`, but _without_ this patch, fails.

Relevant `vcpkg` PR: https://github.com/microsoft/vcpkg/pull/48304

jeffbolznv:
I think this is a glslang bug that I fixed a long time ago. What version of glslangvalidator are you using?

PR #1323, opened 2025-08-09
ggml: fix ggml_conv_1d_dw bug

jason-ni:
The ggml_conv_1d_dw doesn't work. There are 2 logical errors:
1. After im2col expansion, the matrix multiply should use the "new_a" instead of original "a".
2. When reshape the matrix multiply result to input dimentions, it didn't consider padding mode that derive different length with input.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -4213,9 +4213,9 @@ struct ggml_tensor * ggml_conv_1d_dw(
 
     struct ggml_tensor * im2col = ggml_im2col(ctx, new_a, new_b, s0, 0, p0, 0, d0, 0, false, GGML_TYPE_F16);
 
-    struct ggml_tensor * result = ggml_mul_mat(ctx, im2col, a);
+    struct ggml_tensor * result = ggml_mul_mat(ctx, im2col, new_a);
 
-    result = ggml_reshape_3d(ctx, result, b->ne[0], b->ne[1], 1);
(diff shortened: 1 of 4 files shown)


ggerganov:
This breaks the existing WavTokenizer implementation in `llama.cpp`:

```bash
llama-tts --tts-oute-default -p "Hi, this is a test message." && ffplay output.wav 

...
llama.cpp/ggml/src/ggml.c:3018: GGML_ASSERT(ggml_can_mul_mat(a, b)) failed
```

Could you take a look at adjusting the operator to work for both use cases?

jason-ni:
> This breaks the existing WavTokenizer implementation in `llama.cpp`:
> 
> ```shell
> llama-tts --tts-oute-default -p "Hi, this is a test message." && ffplay output.wav 
> 
> ...
> llama.cpp/ggml/src/ggml.c:3018: GGML_ASSERT(ggml_can_mul_mat(a, b)) failed
> ```
> 
> Could you take a look at adjusting the operator to work for both use cases?

Thanks for your reminding. The Pytorch grouped Conv1D weight parameter  is of shape (out_channels, input_channels/groups, kernel_size). In the specialized depthwise case, weight dimension should be (out_channels, 1, kernel_size). So the ne[1] of tensor a should be aways 1. I think the unsqeeze of tensor a is not necessary. Now in the latest commit, both cases work fine.

ggerganov (approved):

PR #1332, opened 2025-08-18
cuda : remove obsolete sources

ggerganov:


Green-Sky:
BTW, those files are still in the codeowners file over at llama.cpp
https://github.com/ggml-org/llama.cpp/blob/master/CODEOWNERS

ggerganov:
> BTW, those files are still in the codeowners file over at llama.cpp https://github.com/ggml-org/llama.cpp/blob/master/CODEOWNERS

Thanks for spotting - will remove them in a bit.

PR #1334, opened 2025-08-22
ci : add github release job

danbev:
This commit adds a GitHub Actions workflow to automate the release process. Currently this will only create an archive of the sources for ggml when a tag is pushed.

The motivation for this is that when we start releasing versions of ggml using semantic versioning it can be nice to have the sources needed for ggml to be deployed as a github release. This enables CMake users that use `FetchContent` efficiently specify the the zip file instead of cloning, and the zip file will only contain the files needed and not all the source files which is the case with the "Source Code (zip)" that is automatically created by the release job.

Example usage with `FetchContent`:
```cmake
cmake_minimum_required(VERSION 3.14)
project(ggml_example)

set(CMAKE_CXX_STANDARD 17)

include(FetchContent)
FetchContent_Declare(ggml
    URL https://github.com/danbev/ggml/releases/download/v0.1.1-test/ggml-src.zip
    DOWNLOAD_EXTRACT_TIMESTAMP TRUE
)

FetchContent_MakeAvailable(ggml)

add_executable(ggml_example main.cpp)
target_link_libraries(ggml_example ggml)
```
And with the following `main.cpp` file:
```c++
 #include <iostream>
 #include <ggml.h>

 int main() {
     std::cout << "GGML Version: " << ggml_version() << std::endl;
     return 0;
 }
```
This could then be built using:
```console
$ cmake -S . -B build
$ cmake --build build
$ ./build/ggml_example
GGML Version: 0.0.2472
```

An example release can be found here:
https://github.com/danbev/ggml/releases/tag/v0.1.1-test

Refs: https://github.com/ggml-org/ggml/issues/1333

----

I'm not ignoring this notice, it was just easier to try this out on my fork as it involved CI as opposed going via llama.cpp.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,16 +15,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v4
 
-    - name: Configure CMake
-      run: |
-        cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
-
-    - name: Build source package
-      run: |
-        cmake --build build --target ggml_package
-        mkdir -p release
-        cp build/ggml-src.zip release/
-
     - name: Create Release
       id: create_release
       uses: ggml-org/action-create-release@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Configure CMake
+      run: |
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
+
+    - name: Build source package
+      run: cmake --build build --target ggml_package
+
+    - name: Create Release
+      uses: softprops/action-gh-release@v2
```
slaren, line 26: It's better to avoid using external actions in release jobs since there is a security risk. We already have a forked `ggml-org/action-create-release@v1`, used in llama.cpp.

danbev:
If this looks alright, it would be nice to get this in place so that we can move forward with introducing the semantic versioning for ggml in https://github.com/ggml-org/llama.cpp/pull/15499.

slaren:
I am not completely sure that this is necessary, the zip files that github generates are already very small. I will leave the final review up to @ggerganov.

danbev:
> I am not completely sure that this is necessary, the zip files that github generates are already very small.

We could remove this additional zip that I added, and just use the sources zip that is automatically generated?
Either way we would have the release job there for tags when needed later.

ggerganov:
> We could remove this additional zip that I added, and just use the sources zip that is automatically generated?

Yes, let's use the auto-generated ones.

danbev:
I've removed the ggml_src.zip I added, and now only the default source .zip will be released:
https://github.com/danbev/ggml/releases/tag/v1.1.5-test

ggerganov (approved):

PR #1338, opened 2025-09-05
sync : llama.cpp

ggerganov:


ggerganov:
Hm, this error has never occurred before when rebase + merge:

image

PR #1339, opened 2025-09-07
Ignore CLion IDE files

distlibs:
Ignore CLion IDE files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ compile_commands.json
 CMakeSettings.json
 .vs/
 .vscode/
+.idea/
 .clangd
 
 .venv/


danbev:
Perhaps this could be added to a local/global git exclude (.git/info/exclude) instead?

clibdev:
I think `.idea`/ belongs in the repository's `.gitignore` rather than in a local exclude, since JetBrains IDEs (IntelliJ, CLion, PyCharm, etc.) are commonly used and multiple contributors may generate this directory. Keeping it in `.gitignore` avoids accidental commits of IDE configs, aligns with practices in similar projects (e.g. `llama.cpp`
https://github.com/ggml-org/llama.cpp/blob/master/.gitignore#L31).

danbev (approved):

PR #1340, opened 2025-09-10
ci : fix ARM MATMUL_INT8 feature detection

danbev:
This commit adds a check for GGML_MACHINE_SUPPORTS_i8mm when enabling MATMUL_INT8 features, ensuring that i8mm intrinsics are only used when the target hardware actually supports them.

The motivation for this is to fix CI build failures where the feature detection correctly identifies that i8mm is not supported (+noi8mm flag), but MATMUL_INT8 preprocessor definitions are still enabled, causing the compiler to attempt using vmmlaq_s32 intrinsics without i8mm support.


danbev:
Closing this and I'll open separate pull requests in llama.cpp for these issues to get some feedback.

PR #1341, opened 2025-09-10
CUDA : conditionally add cuda architectures

gjasny:
See [GPUs supported](https://en.wikipedia.org/wiki/CUDA#GPUs_supported)


src/ggml-cuda/CMakeLists.txt
```diff
@@ -25,10 +25,14 @@ if (CUDAToolkit_FOUND)
         if (GGML_NATIVE AND CUDAToolkit_VERSION VERSION_GREATER_EQUAL "11.6" AND CMAKE_VERSION VERSION_GREATER_EQUAL "3.24")
             set(CMAKE_CUDA_ARCHITECTURES "native")
         else()
+            if (CUDAToolkit_VERSION VERSION_LESS "13")
+                list(APPEND CMAKE_CUDA_ARCHITECTURES 50-virtual 61-virtual 70-virtual)
+            endif ()
```
JohannesGaessler, line 30:
```suggestion
                set(CMAKE_CUDA_ARCHITECTURES 50-virtual 61-virtual 70-virtual)
            endif()
```

gjasny: I don't see how that improves the situation. If the block is skipped, the later additions to CMAKE_CUDA_ARCHITECTURES also just append.

If you'd like to make it explicit I could add a
```cmake
set(CMAKE_CUDA_ARCHITECTURES)
```
before the blocks. But that feels redundant because all of the code is enclosed by a
```
if (NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
```

Please advise.

PR #1354, opened 2025-09-25
ggml : prepare for development of 0.9.3-dev

danbev:


ggerganov:
This step of creating a "dev" version seems a bit redundant. I think we can skip it?

danbev:
> This step of creating a "dev" version seems a bit redundant. I think we can skip it?

Yes, I guess that is not need. It was mainly a way to differentiate between a release and a development release. But we have the ggml_commit so it could be used to verify this. I'll remove this from the script tomorrow :+1:

PR #1363, opened 2025-09-30
ggml : bump version to 0.9.4

ggerganov:


ggerganov:
@danbev The homebrew package for some reason becomes dirty:

image

Note the suffix of `libggml-0.9.4-dirty.dylib`.

I tried to repro the formula, but it's not dirty in my local copy:

https://github.com/Homebrew/homebrew-core/blob/08beafa3f0d47125ebd6e08de19c8bee78f367c5/Formula/l/llama.cpp.rb

Not sure what is causing this. Wondering if we should just remove the `dirty` suffix logic?

danbev:
> Not sure what is causing this. Wondering if we should just remove the `dirty` suffix logic?

I wonder if this has to do with how it checks out llama.cpp and then perhaps something in the git repo is touched and this is getting picked up by GGML_GIT_DIRTY variable. 

~Lets just remove this, we can always get the commit from the build if we need to determine that actual build a project is using even if the version matches.~ Actually that is not correct as the commit would be the same. Perhaps we could remove this from the version but if the checkout is dirty then we set the commit to indicate this?

PR #1374, opened 2025-10-24
Add circular tiling support (for making seamless textures)

Phylliida:
Adds additional functions:

- ggml_conv_2d_circular
- ggml_conv_2d_dw_direct_circular
- ggml_conv_2d_direct_circular
- ggml_pad_circular

These tweak the inputs to the shaders (which are modified in cuda, cpp, and vulkan) to cause them to behave in a "circular" way (like on a torus, the boundaries wrap around).

This is useful for making seamless textures, for example here is a texture generated using this command in stable-diffusion-cpp:

```
./bin/sd  --diffusion-model /home/bepis/Documents/world/models/qwen-image-lighting-8steps-V1.0-Q4_K_S.gguf  --vae /home/bepis/Documents/world/models/qwen_image_vae.safetensors  --qwen2vl /home/bepis/Documents/world/models/Qwen2.5-VL-7B-Instruct-IQ4_XS.gguf --cfg-scale 2.5 --sampling-method euler -p "onion" --circular --seed 420
```
output

By default this should be enough to make stable diffusion circular. For qwen image, you need to also adjust the rope settings, see my PR here https://github.com/leejet/stable-diffusion.cpp/pull/914 for the stable-diffusion-cpp side of it.

### DIFF
--- a/include/ggml.h
+++ b/include/ggml.h
@@ -2179,13 +2179,6 @@ GGML_API struct ggml_tensor * ggml_conv_transpose_2d_p0(
             int64_t               ne3,
             uint32_t              mode); // ggml_scale_mode [ | ggml_scale_flag...]
 
-    enum ggml_pad_mode {
-        GGML_PAD_MODE_ZERO     = 0,
-        GGML_PAD_MODE_CIRCULAR = 1,
-    };
-
-    GGML_API void ggml_set_pad_mode(struct ggml_tensor * tensor, enum ggml_pad_mode mode);
-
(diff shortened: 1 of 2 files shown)


CISC:
Please be aware you may have better luck submitting this PR (with tests) to `llama.cpp` instead, see [CONTRIBUTING.md](https://github.com/ggml-org/ggml/blob/master/CONTRIBUTING.md).

Phylliida:
@CISC ty, made a pr here https://github.com/ggml-org/llama.cpp/pull/16985

PR #1376, opened 2025-10-27
Rewrite simple-backend to use sched and ggml_backend_load_all

jeffbolznv:
I wanted to have an example that uses the more modern "sched" and backend loading interface.

### DIFF
--- a/examples/simple/simple-backend.cpp
+++ b/examples/simple/simple-backend.cpp
@@ -24,7 +24,9 @@ struct simple_model {
     ggml_backend_t backend {};
     ggml_backend_t cpu_backend {};
     ggml_backend_sched_t sched {};
-    struct ggml_context * ctx {};
+
+    // storage for the graph and tensors
+    std::vector<uint8_t> buf;
 };
 
 // initialize data of matrices to perform matrix multiplication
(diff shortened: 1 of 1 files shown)


jeffbolznv:
@slaren please review.

examples/simple/simple-backend.cpp
```diff
@@ -29,111 +17,90 @@ static void ggml_log_callback_default(ggml_log_level level, const char * text, v
...
-    struct ggml_init_params params0 = {
-        /*.mem_size   =*/ buf_size,
-        /*.mem_buffer =*/ buf.data(),
-        /*.no_alloc   =*/ true, // the tensors will be allocated later by ggml_gallocr_alloc_graph()
-    };
-
-    // create a temporally context to build the graph
-    struct ggml_context * ctx0 = ggml_init(params0);
+struct ggml_cgraph * build_graph(simple_model& model) {
+    struct ggml_cgraph  * gf = ggml_new_graph(model.ctx);
```
slaren: While not strictly necessary for this example since the graph is only built once, applications usually use a different `ggml_context` for the graph that is reset every time a new graph is created (effectively freeing all the tensors allocated for the previous graph). It may be a better example if this separation between model context and graph context is maintained here.

PR #1382, opened 2025-11-09
sync : llama.cpp

ggerganov:


ggerganov:
Flagging this failed case: https://github.com/ggml-org/ggml/actions/runs/19208683722/job/54907640675?pr=1382#step:3:16157
Not sure if it is expected. cc @0cc4m @jeffbolznv

jeffbolznv:
I haven't seen that failure locally, and I don't think we've touched that code in the last few days. If it keeps happening I'll look into it.

PR #1390, opened 2025-11-22
metal: implement cross-entropy and count-equal ops for MNIST

xi-guo-0:
This PR implements the missing Metal backend operations required to run [examples/mnist](https://github.com/ggml-org/ggml/issues/990)


iliailmer (commented):
Hey! 

I am working on a similar PR, I was pointed towards your work, I wanted to ask a couple of questions to understand our codes' differences/similarities better. Maybe we can join forces and add cross entropy together. I'm relatively new to Metal dev work.

I left a couple of comments.

src/ggml-metal/ggml-metal.metal
```diff
@@ -9920,3 +10108,51 @@ kernel void kernel_opt_step_sgd_f32(
 
     x[gid] = x[gid] * (1.0f - pars[0] * pars[1]) - pars[0] * g[gid];
 }
+
+typedef decltype(kernel_cross_entropy_loss<int>) kernel_cross_entropy_loss_t;
+
+template [[host_name("kernel_cross_entropy_loss_f32")]] kernel kernel_cross_entropy_loss_t kernel_cross_entropy_loss<float>;
+template [[host_name("kernel_cross_entropy_loss_f16")]] kernel kernel_cross_entropy_loss_t kernel_cross_entropy_loss<half>;
+template [[host_name("kernel_cross_entropy_loss_i32")]] kernel kernel_cross_entropy_loss_t kernel_cross_entropy_loss<int>;
```
iliailmer: Can you help me understand what is the use case for i32 here? Would that mean we are casting the `logits_ptr` into `int *`? I think it should only be f32 and f16 since logits are always floats

src/ggml-metal/ggml-metal.metal
```diff
@@ -9920,3 +10108,51 @@ kernel void kernel_opt_step_sgd_f32(
...
+template [[host_name("kernel_cross_entropy_loss_i16")]] kernel kernel_cross_entropy_loss_t kernel_cross_entropy_loss<short>;
+
+typedef decltype(kernel_cross_entropy_loss_back<int>) kernel_cross_entropy_loss_back_t;
```
iliailmer:
```suggestion
typedef decltype(kernel_cross_entropy_loss_back<float>) kernel_cross_entropy_loss_back_t;
```

ggerganov:
Note that the count_equal op has just been merged.

src/ggml-metal/ggml-metal.metal
```diff
@@ -1878,6 +1878,194 @@ typedef decltype(kernel_sum_rows<false>) kernel_sum_rows_t;
...
+
+    if (sgitg == 0 && tiisg == 0) {
+        shmem[0] = -row_loss;
+    }
+    threadgroup_barrier(mem_flags::mem_threadgroup);
+
+    if (tgpig.x == 0 && sgitg == 0 && tiisg == 0) {
+        float total_loss = 0.0f;
+        for (int i = 0; i < nrows; i++) {
+            total_loss += dst[i];
```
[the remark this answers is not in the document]
xi-guo-0: Good catch. You are absolutely right—the previous logic caused a race condition and failed to write to dst correctly. I have removed the global accumulation loop. The kernel now computes the loss for each row and writes it directly to dst[tgpig.x] (row index).

src/ggml-metal/ggml-metal.metal
xi-guo-0: You are right. I have removed the i32 and i16 templates since logits are always floating-point numbers (f32 or f16).

src/ggml-metal/ggml-metal.metal
```diff
@@ -1879,6 +1879,196 @@ typedef decltype(kernel_sum_rows<false>) kernel_sum_rows_t;
...
+        if (tiisg == 0) shmem[sgitg] = row_loss;
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+        if (sgitg == 0) {
+            row_loss = (tiisg < nsg) ? shmem[tiisg] : 0.0f;
+            row_loss = simd_sum(row_loss);
+        }
+    }
+
+    if (sgitg == 0 && tiisg == 0) {
+        dst[tgpig.x] = -row_loss;
```
iliailmer: i think this is still out of bounds potentially

PR #1394, opened 2025-12-05
chores: set `CMAKE_RUNTIME_OUTPUT_DIRECTORY` for non standalone build

HerrCai0907:
Some backend depends on `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to create temporary file like metal backened.
Missing `CMAKE_RUNTIME_OUTPUT_DIRECTORY` will cause some cmake error like permission denied (try to copy file to root).
This PR wants to setup a default path for `CMAKE_RUNTIME_OUTPUT_DIRECTORY` when it does not exist.


DaAwesomeP:
Ran into this as well!

PR #1428, opened 2026-02-25
sync : llama.cpp

ggerganov:


include/ggml.h
```diff
@@ -730,10 +730,6 @@ extern "C" {
     GGML_API size_t  ggml_type_size(enum ggml_type type);             // size in bytes for all elements in a block
     GGML_API size_t  ggml_row_size (enum ggml_type type, int64_t ne); // size in bytes for all elements in a row
 
-    GGML_DEPRECATED(
-    GGML_API double ggml_type_sizef(enum ggml_type type), // ggml_type_size()/ggml_blck_size() as float
-    "use ggml_row_size() instead");
-
```
ggerganov: @ckastner After merging this sync, we'll make a new ggml release. Should we bump the minor version because of this change here removing a function from the `ggml.h` header?

PR #1465, opened 2026-04-27
ggml-cuda: warp-cooperative conv_transpose_1d, MUL_MAT_VEC + ADD + ADD fusion, GGML_CUDA_PERF_LOGGER env var

Zbig9000:
## ggml-cuda: three independent perf / diagnostic improvements

This PR bundles three logically-distinct ggml-cuda changes that came
out of profiling / optimising chatterbox.cpp text-to-speech on RTX
5090 (Blackwell, sm_120, CUDA Toolkit 12.8).  Each one is a separate
commit so it can be cherry-picked / squashed individually if you
prefer to land them one at a time.

| # | Commit                                                              | Headline                                                              |
|---|---------------------------------------------------------------------|-----------------------------------------------------------------------|
| 1 | `test: add HiFT-realistic shapes to test_conv_transpose_1d`         | +5 test-backend-ops cases at HiFT-vocoder shapes                      |
| 2 | `ggml-cuda: warp-cooperative conv_transpose_1d kernel`              | **~42× faster** at HiFT shapes                                        |
| 3 | `test: extend test_mul_mat_vec_fusion with decoder-realistic shapes`| +100 test-backend-ops cases (`m=1, n>=1024, k>=1024`)                 |
| 4 | `ggml-cuda: fuse MUL_MAT_VEC + ADD(bias) + ADD(residual)`           | **−12 % total GPU time** on transformer decoders                      |
| 5 | `ggml-cuda: add GGML_CUDA_PERF_LOGGER env var`                      | Cross-backend perf diagnostic mirroring `GGML_VK_PERF_LOGGER`         |

Total: **+571 / -43 lines** across `src/ggml-cuda/` and
`tests/test-backend-ops.cpp`.

---

### Problem
Current scalar kernel: one CUDA thread per output pixel, scanning all
`IC × IL` inputs with a per-iteration skip branch:

```cuda
for (int c = 0; c < src0_ne2; c++) {                // c in [0, IC)
    for (int i = 0; i < src1_ne0; i++) {            // i in [0, IL)
        if (!(idx >= i*s0 && idx < i*s0 + src0_ne0)) {
            continue;                               // skip ~99% of iterations
        }
        accumulator += src0[...] * src1[...];
    }
}
```

At HiFT-vocoder-realistic shapes (`L=303, IC=80, K=16, s0=8`) this is
~24 K loop iterations per output pixel with the skip branch firing on
~99 % of iterations.  HiFT decode spends 67 % of total GPU time in
this kernel (4 700 µs / call) on RTX 5090.

### Solution
1. **Narrow the input range** to what actually contributes:
   `i ∈ [⌈(ol - K + 1)/s0⌉, ⌊ol/s0⌋] ∩ [0, IL-1]`.  At `K=16, s0=8`
   this is **2 iterations of `i`** instead of `IL=O(100)`.  Skip
   branch eliminated entirely.
2. **Parallelise IC reduction** across the warp (32 threads); reduce
   via 5 stages of `__shfl_xor_sync`.
3. Block size: 256 → 32 (one warp per output pixel).

### Perf (RTX 5090, CUDA 12.8)
* `conv_transpose_1d` kernel time: **4 700 µs → 110 µs per call (~42×)**
* HiFT decode total GPU time: **67 % → 1.6 %**
* End-to-end speedup on chatterbox 232-token utterance: **4.7× faster HiFT**

### Tests
Existing 116 `test_conv_transpose_1d` cases continue to pass.  Commit
1 of this PR adds 5 HiFT-realistic shapes that exercise the
warp-cooperative reduction at scale (`IC > 32` multi-warp accumulation,
`K > s0` inner-loop unroll).  These pass on **both** the legacy
scalar kernel and the new warp-cooperative kernel — they're a
regression guard for future changes, not a "fail before, pass after"
pattern.

| Test phase                                                | Result        |
|-----------------------------------------------------------|---------------|
| `test-backend-ops -o CONV_TRANSPOSE_1D -b CUDA0` baseline | 116/116 PASS  |
| `…` after commit 1 (test-only)                            | 121/121 PASS  |
| `…` after commit 2 (kernel rewrite)                       | 121/121 PASS  |

Files: `src/ggml-cuda/conv-transpose-1d.cu`,
`src/ggml-cuda/conv-transpose-1d.cuh`, `tests/test-backend-ops.cpp`.

---

### Problem
ggml-cuda's fusion engine knows the 2-op pattern `MUL_MAT_VEC +
ADD(bias)` but not the 3-op pattern that includes the residual.  The
graph shape `((mat * y) + bias) + residual` is common in transformer
attention-output and FFN-output blocks; ggml-vulkan already fuses it
via the `MUL_MAT_ADD_ADD` shader.  Profiling chatterbox vs Vulkan
showed CUDA paid ~67 ms / utterance more in stand-alone ADD launches.

### Solution
1. Extend `ggml_cuda_mm_fusion_args_*` (in `common.cuh`) with an
   `x_residual` field.  When set together with `x_bias`, the matmul-vec
   kernel performs `dst = mat * y + bias + residual` in a single
   dispatch.
2. Detect the `{MUL_MAT, ADD, ADD}` pattern in
   `ggml_cuda_graph_evaluate_and_capture` (placed **above** the
   existing 2-op fusion so the greedy match prefers the larger).
   Resolve `bias_tensor` / `residual_tensor` from the cgraph;
   reject patterns with broadcasting on either ADD (matches the 2-op
   fusion's existing constraint).
3. Update `mul_mat_vec_q` (`mmvq.cu`) and `mul_mat_vec_f` (`mmvf.cu`)
   templates to consume `x_residual` — added after bias and any GLU,
   matching ggml-vulkan's `MUL_MAT_ADD_ADD` execution order.

Only `MUL_MAT` (not `MUL_MAT_ID`) is handled — the residual ADD
pattern doesn't apply to MoE expert routing.  The new code path
falls back gracefully to plain dispatch when the fusion isn't
applicable.

### Perf (RTX 5090, CUDA 12.8, chatterbox Turbo Q4_0, 232-token prompt)
* Total GPU time per utterance: **−12 %**
* `MUL_MAT_VEC q4_0` op-bucket time: **−47 ms / utterance**
* CUDA ↔ Vulkan gap on long prompts: **1.29× → 1.13×**

### Tests
`test_mul_mat_vec_fusion::build_graph` already exercises the
`((mm * y) + bias) + residual` pattern via:

```cpp
ggml_tensor * ffn_up = ggml_mul_mat(ctx, up, cur);
if (with_bias) {
    ffn_up = ggml_add(ctx, ffn_up, up_bias);          // 2nd op
}
ggml_tensor * out = with_gate ? build_gate(ctx, ffn_gate, ffn_up) : ffn_up;
out = ggml_add(ctx, out, bias2);                      // 3rd op  ← residual
```

Commit 3 adds 100 new cases at decoder-realistic shapes (`m=1,
n ∈ {1024, 3072}, k=1024`) that catch FP-precision and per-row stride
issues that don't surface at the existing `m=1, n=32, k=256` cases.

| Test phase                                                  | Result        |
|-------------------------------------------------------------|---------------|
| `test-backend-ops -o MUL_MAT_VEC_FUSION -b CUDA0` baseline  | 300/300 PASS  |
| `…` after commit 3 (test-only)                              | 400/400 PASS  |
| `…` after commit 4 (fusion patch, **first attempt**)        | **2/400 FAIL** at `batch_dims=[4,2]` |
| Bug found: `x_residual` missing `(sample_dst, channel_bias)` offset in `mmvf.cu` | — |
| `…` after fix                                               | 400/400 PASS  |

The bug was latent in the original prototype patch (developed against
chatterbox where `ne[2]==ne[3]==1` makes the offset 0).  Upstream's
broader test coverage caught it immediately when stacked against
`batch_dims=[4,2]` — a good demonstration of why the test-before-
change discipline matters.  Fixed in the same commit (`mmvf.cu` now
applies the offset symmetric to `x_bias`).

Files: `src/ggml-cuda/common.cuh`, `src/ggml-cuda/ggml-cuda.cu`,
`src/ggml-cuda/mmvf.cu`, `src/ggml-cuda/mmvq.cu`,
`tests/test-backend-ops.cpp`.

---

### Problem
ggml-vulkan ships `GGML_VK_PERF_LOGGER=1` and prints a structured
table after every compute graph; ggml-cuda has no equivalent.
Cross-backend perf characterisation today requires `nsys` (heavy,
NVIDIA-only, sometimes needs root for hardware counters) or one-off
manual instrumentation.

### Solution
A symmetric env var that prints the same output format as
`vk_perf_logger`:

```
----------------
CUDA Timings:
MUL_MAT q4_0 m=3072 n=383 k=1024: 24 x 241.979 us = 5807.507 us
FLASH_ATTN_EXT (64,16,411,1): 24 x 5996.571 us = 143917.704 us
…
Total time: 22480.220 us.
```

Same prefix structure (`----------------`, `<Backend> Timings:`,
per-op rows, `Total time:` summary), same number formatting — so
existing cross-backend grep / awk one-liners work unchanged.

### Implementation notes
* Meyers-singleton `ggml_cuda_perf_logger`, off by default.  Only the
  function-local-static `getenv` check runs in the hot path when the
  env var is unset.
* RAII `scope` helper records `cudaEventRecord(start)` on
  construction and `cudaEventRecord(end)` on destruction — one
  scope per dispatched op, including before any `continue` taken
  by the fusion fast-paths.
* `cudaStreamSynchronize` + `flush_and_print` is called at the end of
  every `ggml_backend_cuda_graph_compute` so elapsed times are
  readable before the events are re-used.
* CUDA Graphs auto-disable when the env var is set (otherwise events
  would either re-record over still-pending events or get hidden
  inside `cudaGraphLaunch`).
* Destructor flushes any pending data but **does not** call
  `cudaEventDestroy` — the singleton's dtor runs at static
  destruction time (after `main()`), where the CUDA driver may
  already be torn down.  OS reclaim is safe; pool is bounded.

### Tests
The env var is a stderr-only diagnostic that doesn't change op
output, so there's no natural insertion point in `test-backend-ops`
for testing it.  Manual smoke verifies `GGML_CUDA_PERF_LOGGER=1
test-backend-ops test -o ADD -b CUDA0` produces well-formed `CUDA
Timings:` blocks (99 blocks across the ADD test sweep, matching
`vk_perf_logger`'s format byte-for-byte).

Functional / integration testing of the env var path lives downstream
in chatterbox.cpp's `scripts/test-cuda-perf-logger.sh` (4 phases:
default-silent, env-on-output-format, env+graphs-mutual-exclusion,
aggregate-time-bound) — happy to port a stripped-down version if
you'd prefer a unit test in this PR.

Files: `src/ggml-cuda/ggml-cuda.cu`, `src/ggml-cuda/common.cuh`.

---

## Combined regression — 12189/12189 PASS

The three changes are independent (different files / different code
paths), but they're stacked on the same branch in this PR.  Final
sanity check on the combined `pr-cuda-bundle`:

```
$ ./build/bin/test-backend-ops test -b CUDA0
…
  12189/12189 tests passed
  Backend CUDA0: OK
2/2 backends passed
OK
```

That's `12084 (baseline @ 8be60f8) + 5 (HiFT cases) + 100
(decoder-shape cases) = 12189`, no regressions.  Bench logs
(per-PR + combined) are available on request.

---

## Reproduction

```bash
git fetch origin pr-cuda-bundle
git checkout pr-cuda-bundle
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DGGML_CUDA=ON -DGGML_BUILD_TESTS=ON
cmake --build build --target test-backend-ops -j
./build/bin/test-backend-ops test -b CUDA0

# verify the perf logger:
GGML_CUDA_PERF_LOGGER=1 ./build/bin/test-backend-ops test -o ADD -b CUDA0
```

---

## Splitting if preferred

If you'd rather review / land these one at a time, the 5 commits map
1:1 onto three separable PRs:

* commits 1+2 → conv_transpose_1d kernel rewrite
* commits 3+4 → 3-op fusion
* commit  5  → perf logger

The corresponding individual PR descriptions (with a bit more detail
per piece) are at the following branches in the same fork, ready to
push:
[`pr1-conv-transpose-1d-cuda-warp-cooperative`](https://github.com/Zbig9000/ggml/commits/pr1-conv-transpose-1d-cuda-warp-cooperative),
[`pr2-mul-mat-vec-add-add-fusion-cuda`](https://github.com/Zbig9000/ggml/commits/pr2-mul-mat-vec-add-add-fusion-cuda),
[`pr3-cuda-perf-logger`](https://github.com/Zbig9000/ggml/commits/pr3-cuda-perf-logger).
Happy to convert this PR into three separate ones if reviewers
prefer.

---

## Companion findings (not in this PR)

While preparing this work, profiling on RTX 5090 (Blackwell, sm_120)
revealed that `ggml_cuda_fattn_mma_get_config` has no Blackwell entry
— `ampere_mma_available(cc)` returns `true` for any `cc >= 800`, so
sm_120 silently uses the Ampere config tuned for sm_80.  This is the
single biggest remaining `FLASH_ATTN_EXT` perf gap to ggml-vulkan on
chatterbox-style workloads (~67 ms / utterance).  Tuning a Blackwell
config empirically requires either ncu hardware counters or a
multi-day parameter sweep on Blackwell hardware — best done by
maintainers with multi-Blackwell access.  Happy to file a separate
issue / discussion thread for this if there's interest.

The new `GGML_CUDA_PERF_LOGGER` shipped in this PR makes such an
A/B sweep convenient (no rebuild required to switch variants when
combined with a small follow-up env var to override the picker).

### DIFF
--- a/src/ggml-cuda/common.cuh
+++ b/src/ggml-cuda/common.cuh
@@ -1467,11 +1467,23 @@ struct ggml_cuda_mm_fusion_args_host {
     const ggml_tensor * x_bias = nullptr;
     const ggml_tensor * gate = nullptr;
     const ggml_tensor * gate_bias = nullptr;
+    // Residual tensor added to the matmul output AFTER bias and (if any) gate.
+    // When both x_bias and x_residual are set the kernel performs
+    //     dst = mat * y + bias + residual
+    // in a single dispatch, mirroring ggml-vulkan's MUL_MAT_ADD_ADD shader and
+    // saving the launch overhead of a stand-alone GGML_OP_ADD per residual
+    // connection.  Used by the 3-op MUL_MAT + ADD(bias) + ADD(residual) fusion
+    // detected in ggml_backend_cuda_graph_compute.  Must have ne[0] ==
+    // dst->ne[0] and the same shape as the bias-add output (no broadcasting).
+    // Set to nullptr for normal 2-op MUL_MAT + ADD(bias) fusion or unfused
+    // dispatch.
+    const ggml_tensor * x_residual = nullptr;
     ggml_glu_op glu_op;
 };
 struct ggml_cuda_mm_fusion_args_device {
     const void * x_bias = nullptr;
     const void * gate = nullptr;
     const void * gate_bias = nullptr;
+    const void * x_residual = nullptr;  // see _host counterpart for semantics
     ggml_glu_op glu_op;
 };
--- a/src/ggml-cuda/common.cuh
+++ b/src/ggml-cuda/common.cuh
@@ -1194,7 +1194,12 @@ struct ggml_cuda_graph {
 
     bool is_enabled() const {
         static const bool disable_cuda_graphs_due_to_env = (getenv("GGML_CUDA_DISABLE_GRAPHS") != nullptr);
-        return !(disable_due_to_gpu_arch || disable_cuda_graphs_due_to_env);
+        // Disable graphs when the per-op perf logger is on: graph capture
+        // would either hide individual-op timings inside cudaGraphLaunch
+        // or re-record over still-pending events on subsequent launches.
+        // See ggml-cuda.cu's ggml_cuda_perf_logger comment for context.
+        static const bool disable_cuda_graphs_due_to_perf_logger = (getenv("GGML_CUDA_PERF_LOGGER") != nullptr);
+        return !(disable_due_to_gpu_arch || disable_cuda_graphs_due_to_env || disable_cuda_graphs_due_to_perf_logger);
     }
 #endif
 };
--- a/src/ggml-cuda/conv-transpose-1d.cu
+++ b/src/ggml-cuda/conv-transpose-1d.cu
@@ -1,57 +1,110 @@
 #include "conv-transpose-1d.cuh"
 
-static  __global__ void conv_transpose_1d_kernel(
+// One CUDA warp (32 threads) cooperatively computes one output pixel
+// dst[oc, ol] (== dst[ol + OL*oc] in linear index, since we keep ne2/ne3 == 1).
+//
+// Grid : (OL, OC, 1)
+// Block: (32, 1, 1) — exactly one warp; sized below as CUDA_CONV_TRANPOSE_1D_BLOCK_SIZE.
+//
+// Two perf-critical changes vs the original "1 thread per output pixel + scan
+// the full IC*IL grid + skip via conditional" implementation:
+//
+//   1. Narrow the input position i to the small range that actually
+//      contributes:
+//          out[ol, oc] = sum over (ic, i, ki) of  k[ki, oc, ic] * x[i, ic]
+//          subject to  i*s0 + ki == ol,  0 <= ki < K,  0 <= i < IL
+//      ⇒    i ∈ [ ceil((ol - K + 1)/s0), floor(ol/s0) ] ∩ [0, IL-1]
+//      typically (KS=16, s0=8) this is 2 iterations of i instead of IL=O(100).
+//
+//   2. Parallelise the IC reduction across the warp (each thread handles a
+//      strided slice of IC) and finalise with __shfl_xor_sync.  This gives
+//      32× useful work per warp on top of the i-range narrowing.
+//
+// Layouts (matching the original kernel and the Vulkan / Metal patches):
+//   src0 (kernel) : [K, OC, IC] row-major  → element (ki, oc, ic) at
+//                                            ic*(OC*K) + oc*K + ki
+//   src1 (input)  : [IL, IC]    row-major  → element (i, ic) at ic*IL + i
+//   dst           : [OL, OC]    row-major  → element (ol, oc) at oc*OL + ol
+//
+// Limitation (unchanged from the original kernel): only ne1==ne3==1 is
+// supported; the host-side wrapper enforces that via the contiguous +
+// shape assertions.
+static __global__ void conv_transpose_1d_kernel(
         const int s0, const int p0, const int d0, const int output_size,
         const int src0_ne0, const int src0_ne1, const int src0_ne2, const int src0_ne3,
         const int src1_ne0, const int src1_ne1, const int src1_ne2, const int src1_ne3,
-        const int dst_ne0, const int dst_ne1, const int dst_ne2, const int dst_ne3,
-        const float * src0, const float * src1,  float * dst) {
-    int global_index = threadIdx.x + blockIdx.x * blockDim.x;
-    if (global_index >= output_size) {
+        const int dst_ne0,  const int dst_ne1,  const int dst_ne2,  const int dst_ne3,
+        const float * __restrict__ src0, const float * __restrict__ src1, float * __restrict__ dst) {
+
+    const int ol = blockIdx.x;
+    const int oc = blockIdx.y;
+    if (ol >= dst_ne0 || oc >= dst_ne1) {
         return;
     }
 
-    int out_index = global_index / dst_ne0;
+    const int K  = src0_ne0;
+    const int OC = dst_ne1;
+    const int IC = src0_ne2;
+    const int IL = src1_ne0;
 
-    float accumulator = 0;
+    // Range of input positions i that contribute to this output pixel.
+    int i_start = (ol - K + 1 + s0 - 1) / s0;   // ceil((ol - K + 1) / s0)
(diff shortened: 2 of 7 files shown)


Zbig9000:
Filed companion issue #1466 documenting the Blackwell flash-attn config gap that I mentioned in this PR's last section. The `GGML_CUDA_PERF_LOGGER` shipped here is what makes that A/B sweep tractable for maintainers with multi-Blackwell hardware.

PR #1476, opened 2026-05-05
sync : llama.cpp

ggerganov:


ggerganov:
@taronaeo The python deps are failing to install on the self-hosted runner:

https://github.com/ggml-org/ggml/actions/runs/25366621707/job/74379012178#step:3:155

Not sure how to fix it. Any ideas?

taronaeo:
> @taronaeo The python deps are failing to install on the self-hosted runner:
> 
> https://github.com/ggml-org/ggml/actions/runs/25366621707/job/74379012178#step:3:155
> 
> Not sure how to fix it. Any ideas?

This https://github.com/ggml-org/ggml/blob/ac6f7b44f60fde0091f0b3d99afde48f8c99b13a/requirements.txt#L11 and this https://github.com/ggml-org/ggml/blob/ac6f7b44f60fde0091f0b3d99afde48f8c99b13a/requirements.txt#L4 are mismatching. The latest version of `torchvision` deprecated support for `torch` 2.5.X and since we are forcing `torch` to match versions within 2.5.X, pip can't find a suitable version to install.

I would suggest that we update and align the version requirements with llama.cpp (i.e., bumping torch version to `torch~=2.6.0`). Also to apply supply-chain hardening, same as what we did in llama.cpp since we currently are accepting versions greater than 0.15.2 for torchvision.

This problem is isolated to the GGML repository only.

PR #1479, opened 2026-05-06
requirements: sync requirements.txt with llama.cpp versions

taronaeo:
ref: https://github.com/ggml-org/ggml/pull/1476#issuecomment-4378332120

This PR fixes the CI error referenced above and attempts to match the package dependency requirements with llama.cpp to ensure consistency. Also applies supply-chain hardening, the same that was done in llama.cpp.

Tested to be working on sg-hl1-ci-nvidia-vulkan-cm runner (where the `pip install -r` failed @ Python 3.13. Not sure what to test in this repository so please advise if needed.

cc: @ggerganov

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ torchvision~=0.21.0
 transformers==5.5.1
 gguf>=0.1.0
 keras==3.5.0
-tensorflow==2.18.0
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Thanks!

PR #1485, opened 2026-05-10
ggml-cuda : handle OW > 65535 in im2col (2D and 3D)

CrispStrobe:
`im2col_cuda` and `im2col_3d_cuda` both dispatch with
`block_nums.y = OW`. CUDA caps grid Y at 65535. Conv1d encoders on
raw 16 kHz audio with T > 65535 (~ 4 s) trip the limit — e.g. SEANet
at 11 s lands at OW = 176000 — and the launch returns
`invalid configuration argument`.

Fix: clamp `block_nums.y` to `MIN(OW, MAX_GRIDDIM_Y)` and loop inside
the kernel with stride `MAX_GRIDDIM_Y`. Same in-kernel stride pattern
already used for the z axis (`MAX_GRIDDIM_Z`). Both 2D `im2col_kernel`
and 3D `im2col_3d_kernel` need the same fix. Bit-identical for
OW ≤ 65535 (single iteration of the new outer loop).

### Verification

Tested on T4 / Jetson Orin with a SEANet encoder running on 11 s /
16 kHz audio (im2col reaching OW ≈ 176000); pre-fix launch returns
`invalid configuration argument`, post-fix runs to completion.
Existing `test-backend-ops` im2col cases unchanged.

### DIFF
--- a/src/ggml-cuda/im2col.cu
+++ b/src/ggml-cuda/im2col.cu
@@ -1,5 +1,6 @@
 #include "im2col.cuh"
 
+#define MAX_GRIDDIM_Y 65535
 #define MAX_GRIDDIM_Z 65535
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


CISC:
Please submit this to `llama.cpp` instead, it's not likely to get reviewed here.

CrispStrobe:
Thanks for the steer @CISC — that matches what the commit log shows (recent `src/ggml-cuda/` commits all come in as `(llama/NNNNN)` syncs) and what `README.md` notes about development happening in llama.cpp. Closing this and re-opening against `ggml-org/llama.cpp` so it lands where the CUDA reviewers actually look. I'll link the new PR here once it's up.

CrispStrobe:
For posterity: refiled as [ggml-org/llama.cpp#22944](https://github.com/ggml-org/llama.cpp/pull/22944).

JohannesGaessler:
I think it would make sense to ping ggml-org maintainers not just for PRs to the llama.cpp repository but for PRs to the standalone ggml repository as well.

CISC:
> I think it would make sense to ping ggml-org maintainers not just for PRs to the llama.cpp repository but for PRs to the standalone ggml repository as well.

I guess, the PR template notice does not seem to be very effective. :)

PR #1487, opened 2026-05-12
docs : add MTP to GGUF Type slot

mishig25:
Adds `MTP` as a third value for the `Type` slot (alongside `LoRA` and `vocab`) to cover Multi-Token Prediction / speculative-decoding draft modules shipped beside a base model. Updates the regex in both spots and adds an example + test case.

@ggerganov previously OK'd adding `MTP`.

**Open question:** the spec puts `Type` after `Encoding`, so the canonical form here is `…-Q4_K_M-MTP.gguf`. Some community releases place `MTP` before the encoding (`…-MTP-Q4_K_M.gguf`). Happy to flip if you'd rather match current filenames.

### DIFF
--- a/docs/gguf.md
+++ b/docs/gguf.md
@@ -44,6 +44,7 @@ The components are:
   - If missing, then file is by default a typical gguf tensor model file
   - `LoRA` : GGUF file is a LoRA adapter
   - `vocab` : GGUF file with only vocab data and metadata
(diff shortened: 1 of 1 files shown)


mishig25:
Closed by an unintentional branch rename on my fork — superseded by #1488 with identical changes.

PR #1488, opened 2026-05-12
docs : add MTP to GGUF Type slot

mishig25:
Adds `MTP` as a third value for the `Type` slot (alongside `LoRA` and `vocab`) to cover Multi-Token Prediction / speculative-decoding draft modules shipped beside a base model. Updates the regex in both spots and adds an example + test case

### DIFF
--- a/docs/gguf.md
+++ b/docs/gguf.md
@@ -44,6 +44,7 @@ The components are:
   - If missing, then file is by default a typical gguf tensor model file
   - `LoRA` : GGUF file is a LoRA adapter
   - `vocab` : GGUF file with only vocab data and metadata
+  - `MTP` : GGUF file contains Multi-Token Prediction heads (a speculative-decoding draft module), intended to be loaded alongside a base model of matching architecture and version
 1. **Shard**: (Optional) Indicates and denotes that the model has been split into multiple shards, formatted as `<ShardNum>-of-<ShardTotal>`.
     - *ShardNum* : Shard position in this model. Must be 5 digits padded by zeros.
       - Shard number always starts from `00001` onwards (e.g. First shard always starts at `00001-of-XXXXX` rather than `00000-of-XXXXX`).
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

docs/gguf.md
```diff
@@ -54,7 +55,7 @@ The components are:
...
 
-To validate you can use this regular expression `^(?<BaseName>[A-Za-z0-9\s]*(?:(?:-(?:(?:[A-Za-z\s][A-Za-z0-9\s]*)|(?:[0-9\s]*)))*))-(?:(?<SizeLabel>(?:\d+x)?(?:\d+\.)?\d+[A-Za-z](?:-[A-Za-z]+(\d+\.)?\d+[A-Za-z]+)?)(?:-(?<FineTune>[A-Za-z0-9\s-]+))?)?-(?:(?<Version>v\d+(?:\.\d+)*))(?:-(?<Encoding>(?!LoRA|vocab)[\w_]+))?(?:-(?<Type>LoRA|vocab))?(?:-(?<Shard>\d{5}-of-\d{5}))?\.gguf$` which will check that you got the minimum BaseName, SizeLabel and Version present in the correct order.
+To validate you can use this regular expression `^(?<BaseName>[A-Za-z0-9\s]*(?:(?:-(?:(?:[A-Za-z\s][A-Za-z0-9\s]*)|(?:[0-9\s]*)))*))-(?:(?<SizeLabel>(?:\d+x)?(?:\d+\.)?\d+[A-Za-z](?:-[A-Za-z]+(\d+\.)?\d+[A-Za-z]+)?)(?:-(?<FineTune>[A-Za-z0-9\s-]+))?)?-(?:(?<Version>v\d+(?:\.\d+)*))(?:-(?<Encoding>(?!LoRA|vocab|MTP)[\w_]+))?(?:-(?<Type>LoRA|vocab|MTP))?(?:-(?<Shard>\d{5}-of-\d{5}))?\.gguf$` which will check that you got the minimum BaseName, SizeLabel and Version present in the correct order.
```
ngxson, line 58: I think is might be incorrect, we went with the `mtp-` prefix in this PR: https://github.com/am17an/llama.cpp/pull/9/changes#diff-03b361169a690c5ac8e77460aeba18d833d2b78babab92b7b5bb721fc34947c9R612-R620

ngxson: also `mmproj-` as prefix is pretty much a standard now, we should probably also add it to the regex

julien-c: yes 👍

ngxson: feel free to open a follow-up PR @mishig25

mishig25: Opened #1496 : moves `MTP` out of `Type` and introduces a `Sidecar` prefix slot covering both `mtp-` and `mmproj-`. wdyt?

julien-c: actually it changed later to be like this no?
image

julien-c: or am i missing something

julien-c: where as the unsloth ones don't encode MTP in the gguf filename
image

mishig25: Unsloth-style (MTP in repo name, clean filename): the entire repo is dedicated to MTP variants, so MTP is implied by the repo name. Each file inside just needs to disambiguate by quant (Qwen3.6-27B-Q4_K_M.gguf, Qwen3.6-27B-Q5_K_M.gguf, etc.). See [unsloth/Qwen3.6-27B-MTP-GGUF](https://huggingface.co/unsloth/Qwen3.6-27B-MTP-GGUF/tree/main) repo

ngxson: yes, indeed if the model already come with MTP support, it's always better to have both main model + LLM in the same GGUF, it does save a bit of VRAM that way.

the case where MTP and main model are separate GGUFs is mostly for eagle3-style models

julien-c: do you have an example repo for eagle3-style?

mishig25: > do you have an example repo for eagle3-style?

so far, every Eagle3 repo I found ships safetensors only ([nvidia/gpt-oss-120b-Eagle3-v3](https://huggingface.co/nvidia/gpt-oss-120b-Eagle3-v3), [openbmb/MiniCPM4.1-8B-Eagle3](https://huggingface.co/openbmb/MiniCPM4.1-8B-Eagle3), [thoughtworks/Qwen3-8B-Eagle3](https://huggingface.co/thoughtworks/Qwen3-8B-Eagle3))

PR #1489, opened 2026-05-13
docs(gguf): extend Encoding slot to support percentage-mixed recipes

mishig25:
Extends the `<Encoding>` slot to allow a multi-component recipe form for files with a genuinely mixed byte distribution (e.g. asymmetric MoE quants where routed experts use a different ggml type than attention projections). Single-token Encodings are unchanged.

### New form

```
DeepSeek-V4-Flash-256x8.4B-v1.0-55IQ2_XXS-34Q2_K-07Q8_0-03F16.gguf
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                Encoding: ~55% IQ2_XXS, ~34% Q2_K, ~7% Q8_0, ~3% F16
```

- `<pct>` is 1–3 digits, components listed descending by byte share.
- Only components above ~2% should appear; the sum need not equal 100.

### Regex change

```diff
-(?<Encoding>(?!LoRA|vocab|MTP)[\w_]+)
+(?<Encoding>(?!LoRA|vocab|MTP)[\w_]+(?:-\d{1,3}[A-Z][\w_]*)*)
```

Verified backward-compatible against every existing example in the doc — Encoding/Type/Shard captures remain identical.

### Originating use case

[`huggingface.co/antirez/deepseek-v4-gguf`](https://huggingface.co/antirez/deepseek-v4-gguf) — DeepSeek V4 Flash shipped with routed experts at IQ2_XXS/Q2_K and the rest at Q8_0/F16. No single `LLAMA_FTYPE_MOSTLY_*` value captures this split honestly. Parser implementation in [`huggingface.js#2170`](https://github.com/huggingface/huggingface.js/pull/2170).

### DIFF
--- a/docs/gguf.md
+++ b/docs/gguf.md
@@ -40,6 +40,7 @@ The components are:
     - If model is missing a version number then assume `v1.0` (First Public Release)
     - This can be derived from gguf metadata `general.version`
 1. **Encoding**: Indicates the weights encoding scheme that was applied to the model. Content, type mixture and arrangement however are determined by user code and can vary depending on project needs.
+   - For files where the byte distribution is genuinely mixed (e.g. asymmetric MoE quants where routed experts use a different ggml type than attention projections), the Encoding may instead be a hyphen-joined sequence of `<pct><quant>` tokens listed in descending order of byte share, where `<pct>` is a 1–3 digit percentage and `<quant>` is a ggml tensor type name (e.g. `IQ2_XXS`, `Q2_K`, `Q4_K`, `Q8_0`, `F16`, `F32`, `BF16`). Only components representing more than ~2% of bytes should appear, so the sum need not equal 100. Example: `55IQ2_XXS-34Q2_K-07Q8_0-03F16` means the file's bytes are ~55% IQ2_XXS, ~34% Q2_K, ~7% Q8_0, ~3% F16.
 1. **Type**: Indicates the kind of gguf file and the intended purpose for it
   - If missing, then file is by default a typical gguf tensor model file
   - `LoRA` : GGUF file is a LoRA adapter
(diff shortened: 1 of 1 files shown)


mishig25:
Verified the regex locally against the full example set in the doc — all parse correctly:

```js
const ggufRegex = /^(?<BaseName>[A-Za-z0-9\s]*(?:(?:-(?:(?:[A-Za-z\s][A-Za-z0-9\s]*)|(?:[0-9\s]*)))*))-(?:(?<SizeLabel>(?:\d+x)?(?:\d+\.)?\d+[A-Za-z](?:-[A-Za-z]+(\d+\.)?\d+[A-Za-z]+)?)(?:-(?<FineTune>[A-Za-z0-9\s-]+))?)?-(?:(?<Version>v\d+(?:\.\d+)*))(?:-(?<Encoding>(?!LoRA|vocab|MTP)[\w_]+(?:-\d{1,3}[A-Z][\w_]*)*))?(?:-(?<Type>LoRA|vocab|MTP))?(?:-(?<Shard>\d{5}-of-\d{5}))?\.gguf$/;
```

```
✓ Mixtral-8x7B-v0.1-KQ2.gguf
    BaseName = Mixtral | SizeLabel = 8x7B | Version = v0.1
    Encoding = KQ2 | Type = undefined | Shard = undefined
✓ Hermes-2-Pro-Llama-3-8B-v1.0-F16.gguf
    BaseName = Hermes-2-Pro-Llama-3 | SizeLabel = 8B | Version = v1.0
    Encoding = F16 | Type = undefined | Shard = undefined
✓ Grok-100B-v1.0-Q4_0-00003-of-00009.gguf
    BaseName = Grok | SizeLabel = 100B | Version = v1.0
    Encoding = Q4_0 | Type = undefined | Shard = 00003-of-00009
✓ Qwen3-27B-v1.0-Q4_K_M-MTP.gguf
    BaseName = Qwen3 | SizeLabel = 27B | Version = v1.0
    Encoding = Q4_K_M | Type = MTP | Shard = undefined
✓ DeepSeek-V4-Flash-256x8.4B-v1.0-55IQ2_XXS-34Q2_K-07Q8_0-03F16.gguf
    BaseName = DeepSeek-V4-Flash | SizeLabel = 256x8.4B | Version = v1.0
    Encoding = 55IQ2_XXS-34Q2_K-07Q8_0-03F16 | Type = undefined | Shard = undefined
✓ DeepSeek-V4-Flash-256x8.4B-v1.0-95Q4_K-04Q8_0.gguf
    BaseName = DeepSeek-V4-Flash | SizeLabel = 256x8.4B | Version = v1.0
    Encoding = 95Q4_K-04Q8_0 | Type = undefined | Shard = undefined
```

Notes on what the trailing `(?:-\d{1,3}[A-Z][\w_]*)*` alternation does and doesn't absorb:

- **Single-token Encodings** (`KQ2`, `F16`, `Q4_0`, `Q4_K_M`) — the trailing pattern matches 0 times, Encoding ends at the same boundary as before.
- **Type-slot collisions** (`-MTP` after Encoding) — `MTP` has no leading digit, so the trailing pattern can't absorb it; Type slot still captures cleanly.
- **Shard-slot collisions** (`-00003-of-00009` after Encoding) — `\d{1,3}` would match up to 3 digits of `00003`, but the next char is another `0`, not `[A-Z]`. Backtracks through `\d{1,3}` and fails. Shard slot still captures cleanly.

One pre-existing inconsistency I noticed unrelated to this PR: the doc lists `Hermes-2-Pro-Llama-3-8B-F16.gguf` (without `-v1.0-`) but the regex requires the Version slot, so it doesn't actually match that filename. I tested with the explicit-version form. Happy to fix the example in a follow-up if useful.

PR #1515, opened 2026-05-28
ggml-cpu : fix op_expm1 to use expm1f instead of expf(x) - 1

Dev-X25874:
## Problem

`op_expm1` in `src/ggml-cpu/unary-ops.cpp` was implemented as:

```c
static inline float op_expm1(float x) {
    return expf(x) - 1.0f;
}
```

This loses all significant bits for small `x`. When `x ≈ 1e-8f`, `expf(x)` rounds
to exactly `1.0f` in single precision, so `expf(x) - 1.0f = 0.0f` — a 100% relative
error against the correct answer `~1e-8`.

## Fix

Delegate to the standard `expm1f(x)`, which is designed precisely to avoid this
cancellation. The same file already uses `expm1f` correctly in `op_elu` and `op_xielu`.

```c
static inline float op_expm1(float x) {
    return expm1f(x);
}
```

## Test

Added `test_unary_expm1_near_zero` to `tests/test-backend-ops.cpp`. It runs
`GGML_UNARY_OP_EXPM1` on F32 inputs drawn from `[-1e-6, 1e-6]`, where the naive
implementation produces NMSE >> 1 and the fixed implementation stays within `1e-6`.

### DIFF
--- a/src/ggml-cpu/unary-ops.cpp
+++ b/src/ggml-cpu/unary-ops.cpp
@@ -74,7 +74,7 @@ static inline float op_log(float x) {
 }
 
 static inline float op_expm1(float x) {
-    return expf(x) - 1.0f;
+    return expm1f(x);
 }
 
 static inline float op_softplus(float x) {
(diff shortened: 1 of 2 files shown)


ggerganov (commented):
Could you submit this PR to https://github.com/ggml-org/llama.cpp?

PR #1516, opened 2026-05-28
security: harden GGUF parser against memory exhaustion DoS

Ashutosh0x:
## Summary

The GGUF parser has two resource exhaustion vulnerabilities that allow a crafted GGUF file (as small as ~160 bytes) to trigger multi-gigabyte memory allocations and crash any application that loads it.

### 1. Excessive MAX limits for string/array metadata (DoS)

`GGUF_MAX_STRING_LENGTH` and `GGUF_MAX_ARRAY_ELEMENTS` are both set to **1 GiB** (`1024*1024*1024`). A malicious GGUF file can declare a single KV string of length ~1 GiB, forcing the parser to attempt a ~1 GiB allocation before any data is actually read. On systems with limited memory, this causes immediate OOM.

**Fix:** Reduced both limits from 1 GiB to 64 MiB. This is still far above any legitimate model metadata size (typical metadata is < 1 MiB) while preventing trivial DoS.

### 2. Missing upper bound for `general.alignment` (DoS / potential overflow)

The `general.alignment` KV value is validated as a power of 2, but has **no upper bound**. A malicious file can set alignment to `2^30` (1 GiB), causing:
- Massive padding between every tensor (`GGML_PAD(tensor_size, 1GiB)`)
- Total data section size overflow or multi-GiB allocation from a tiny file
- On 32-bit systems, the alignment value can cause integer overflow in padding calculations

**Fix:** Added an upper bound of 1 MiB (`1 << 20`) for alignment. This is well above any real alignment requirement (typically 32-4096 bytes for SIMD/GPU memory alignment).

## Impact

- **Attack vector:** User loads a crafted GGUF model file (e.g., downloaded from a model hub)
- **Effect:** Application OOM crash (Denial of Service)
- **Affected code:** `gguf_init_from_reader()` in `src/gguf.cpp`
- **File size:** A malicious GGUF file triggering this can be as small as ~160 bytes

## Changes

- `src/gguf.cpp`: Reduced `GGUF_MAX_STRING_LENGTH` and `GGUF_MAX_ARRAY_ELEMENTS` from 1 GiB to 64 MiB
- `src/gguf.cpp`: Added `GGUF_MAX_ALIGNMENT` (1 MiB) upper bound check after the existing power-of-2 validation

## Related

- Reported via [huntr.com](https://huntr.com)
- Related but distinct from CVE-2026-33298 (integer overflow in `ggml_nbytes`, fixed in b7824)

CC: @ggerganov @JohannesGaessler

### DIFF
--- a/src/gguf.cpp
+++ b/src/gguf.cpp
@@ -15,8 +15,11 @@
 #include <string>
 #include <vector>
 
-#define GGUF_MAX_STRING_LENGTH  (1024*1024*1024)
-#define GGUF_MAX_ARRAY_ELEMENTS (1024*1024*1024)
+// Security: reduced from 1 GiB to 64 MiB to prevent memory exhaustion DoS
+// via crafted GGUF files with oversized string or array metadata.
+// 64 MiB is still far above any legitimate model metadata size.
+#define GGUF_MAX_STRING_LENGTH  (64*1024*1024)
+#define GGUF_MAX_ARRAY_ELEMENTS (64*1024*1024)
 
 #ifdef _WIN32
 #    define gguf_ftell _ftelli64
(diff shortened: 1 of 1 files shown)


Ashutosh0x:
Hey @ggerganov @JohannesGaessler - this PR addresses two resource exhaustion vectors in the GGUF parser that allow a tiny crafted file (~160 bytes) to trigger multi-GiB allocations and crash any application loading it. The fixes are minimal and backward-compatible: reducing the MAX constants from 1 GiB to 64 MiB (still way above any real metadata), and capping alignment at 1 MiB. Both issues were discovered during security testing and reported via huntr.com. These are distinct from CVE-2026-33298. Happy to adjust the limits if you have different preferences!

JohannesGaessler:
I don't agree that this is an issue that needs resolving.

JohannesGaessler:
I don't have a large sample size but if this happens more frequently we should maybe consider banning huntr.com since it seems to encourage bad vulnerability reports.

Ashutosh0x:
Thanks for the feedback @JohannesGaessler, I appreciate your time reviewing this.

You raise a fair point. Looking at the code more carefully, I see that the 
bytes_remain check (line 349) already prevents oversized string allocations when the file is small, and the std::bad_alloc catch in gguf_read_emplace_helper handles allocation failures gracefully. I should have analyzed these existing safeguards more carefully before opening the PR.

Regarding the alignment upper bound — I noticed that GGML_PAD(tensor_size, alignment) with a very large power-of-2 alignment value can still cause excessive padding in the data section size calculation (lines 773-780), even with the existing overflow checks. But if you consider this acceptable given the local-execution threat model, I understand.

I'm genuinely interested in contributing to ggml — not just for bounty farming. If this particular change isn't useful, I'm happy to close the PR and look for other ways to help. Are there any open issues or areas where contributions would be more welcome?

Thanks again for taking the time to respond.

Ashutosh0x:
Closing this PR based on @JohannesGaessler's feedback. The existing nbytes_remain and bad_alloc checks already handle the resource exhaustion case. I've opened a separate PR (#1517) for a different issue I found during the analysis — a division-by-zero crash via zero-valued tensor dimensions. Thanks for the review.

Issue #1506, opened 2026-05-24
mul_mat_id produces wrong output with K-quantized source weights


OriPekelman:
## Summary

`ggml_mul_mat_id` produces degenerate output when the per-expert weight stack (`as`) is K-quantized (Q4_K, Q5_K, Q6_K). F32, F16, and Q8_0 sources work correctly. Reproduced on CPU with vendored ggml; we believe CUDA is also affected but haven't isolated.

## What we see

Two GGUFs of OLMoE-1B-7B-Instruct, identical model, identical inference code, only the expert weight dtype changes:

| GGUF | expert dtype | output for "The capital of France is" |
|---|---|---|
| `Q4_K_M` (from `Meshwa/OLMoE-1b-7b-0924-Instruct-gguf`) | mostly Q4_K, some Q6_K | "Dub Dub Dub Dub Dub Dub Dub Dub" |
| `q8_0` (same repo) | Q8_0 | "called Paris." |

Both runs use:
- Same router weights (F16)
- Same attention weights (Q4_K for one, Q8_0 for the other)
- Same model graph: `router · x → softmax → top_k → 3× mul_mat_id → silu·up → broadcast-weight → sum-across-K`

Only the **expert weight dtype** passed to the `mul_mat_id`'s `as` parameter differs. Q4_K produces degenerate output; Q8_0 produces coherent factual text.

## Why we suspect a kernel gap

`tests/test-backend-ops.cpp::test_mul_mat_id` only exercises mul_mat_id with F32, F16, and Q8_0 source weights (see test registrations around lines 8264, 8408, 8438 in HEAD as of this writing). K-quants aren't in the test matrix. Our observation is consistent with mul_mat_id either:
- Falling back to a wrong arithmetic path for K-quant sources, or
- Mis-indexing into the K-quant block layout (K-quants have a hierarchical scale + per-sub-block offset structure that's significantly different from the Q4_0 / Q8_0 per-block scale).

`ggml_mul_mat` (non-id) **does** work with K-quant sources on similar tensor shapes — we have OLMoE's attention weights at Q4_K on the same model and the per-head Q/K/V matmuls produce correct output. So K-quants aren't broken in general; just specifically through the mul_mat_id dispatch.

## Suggested test addition

```cpp
// tests/test-backend-ops.cpp, in init_gguf_tests or similar:
for (auto t : {GGML_TYPE_Q4_K, GGML_TYPE_Q5_K, GGML_TYPE_Q6_K}) {
[...]

PR #1518, opened 2026-05-28
tests : add OLMoE-sized K-quant shapes to test_mul_mat_id (ref #1506)

devYRPauli:
The existing `test_mul_mat_id` matrix tops out at `m=512, k=256` with `n_mats in {4,8}, n_used in {1,2,4}`. Production MoE models use much larger shapes - OLMoE-1B-7B is 64 experts with top-8 routing (`gate/up` projection `out=1024 in=2048`, `down` projection `out=2048 in=1024`).

This adds that topology as preventive regression coverage: `Q2_K, Q3_K, Q4_K, Q5_K, Q6_K` at `n in {1, 32}` for both projection shapes, 20 cases total.

All 20 pass on current master. Verified on an M1 Pro with a Release build: a CPU-only run gives 885/885 MUL_MAT_ID cases passed, and a Metal build passes 3/3 backends with zero failures.

Context: this started as coverage for #1506, which reported `mul_mat_id` corruption with K-quants at OLMoE shape. That issue has since been closed - the op-level isolation showed `mul_mat_id` was correct on master, and the reporter traced the corruption to their own loader. So there is no bug to fix here. The value is that the shapes real MoE models actually use were not covered before, and a future regression at MoE scale would now show up in CI.

### DIFF
--- a/tests/test-backend-ops.cpp
+++ b/tests/test-backend-ops.cpp
@@ -9033,6 +9033,16 @@ static std::vector<std::unique_ptr<test_case>> make_test_cases_eval() {
         test_cases.emplace_back(new test_mul_mat_id(type_a, GGML_TYPE_F32, 4, 2, false, 64, 16, 3*ggml_blck_size(type_a)));
     }
 
+    // K-quant coverage at a production MoE topology. The cases above top out at
+    // n_mats=8, m=512, k=256; OLMoE-1B-7B uses 64 experts with top-8 routing,
+    // gate_proj/up_proj out=1024 in=2048 and down_proj out=2048 in=1024.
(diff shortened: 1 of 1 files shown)


devYRPauli:
Closing this one for now. I hadn't realised new contributors are asked to keep to a single open PR, so I'm trimming mine down. Happy to reopen this once I've had a PR merged. Sorry for the extra noise!

devYRPauli:

devYRPauli:
Rebased onto master (90951f99) and trimmed.

Since #1506 was closed as a loader-side issue rather than a `mul_mat_id` bug, I have cut the cases that only existed to chase that report: the reporter-stated `n_mats=8/n_used=2` shape, the IQ-quant block, and the `b_transposed` siblings. What remains is just the real OLMoE-1B-7B topology across the five K-quants at `n in {1, 32}` - 20 cases instead of 47, so the added CI cost is modest.

Happy to trim further or drop this entirely if you would rather not grow the `test_mul_mat_id` matrix.

PR #1519, opened 2026-05-28
ggml-cpu : fix rms_norm_back wrong output under in-place aliasing (#1491)

### LINKED ISSUE #1491: ggml_rms_norm_back: backend-sched compute produces different (wrong) result than legacy compute_with_ctx
## Summary

`ggml_rms_norm_back` produces different values depending on whether
the graph is computed via the legacy `ggml_graph_compute_with_ctx`
(which gives the correct, formula-matching value) or via the backend
scheduler `ggml_backend_sched_graph_compute` on the CPU backend (which
returns a wildly different value with the wrong sign and magnitude).

Both paths use the same op (`ggml_rms_norm_back`), the same CPU
backend, and the same input tensors. Forward `ggml_rms_norm` works
correctly on both paths; the discrepancy is specific to the backward op.

Repro on `ggml` master `5725fee` (the commit in my vendored checkout),
Linux aarch64 (NVIDIA GB10), gcc 13.3, CPU backend only.

## Repro 1 — legacy `ggml_graph_compute_with_ctx` (correct)

```c
#include "ggml.h"
#include "ggml-cpu.h"
#include <stdio.h>
#include <string.h>

int main(void) {
    struct ggml_init_params params = {
        .mem_size   = 256 * 1024,
        .mem_buffer = NULL,
        .no_alloc   = false,
    };
    struct ggml_context *ctx = ggml_init(params);

    struct ggml_tensor *t_a = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, 4, 1);
    struct ggml_tensor *t_b = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, 4, 1);

    float a[4] = {1.0f, 0.0f, 0.0f, 0.0f};  // gradients (src0 per impl source)
    float b[4] = {1.0f, 0.0f, 0.0f, 0.0f};  // x (src1 per impl source)
    memcpy(t_a->data, a, sizeof a);
    memcpy(t_b->data, b, sizeof b);

    struct ggml_tensor *t_out = ggml_rms_norm_back(ctx, t_a, t_b, 1e-4f);
    struct ggml_cgraph *gf = ggml_new_graph(ctx);
    ggml_build_forward_expand(gf, t_out);
    ggml_graph_compute_with_ctx(ctx, gf, 1);

    float *out = (float *) t_out->data;
    printf("out = [%g %g %g %g]\n", out[0], out[1], out[2], out[3]);
    return 0;
}
```

**Output**: `out = [0.000799 0 0 0]`

Matches the formula in `ggml-cpu/ops.cpp::ggml_compute_forward_rms_norm_back_f32`:

```
sum_xx  = Σ x[i]²       = 1
sum_eps = sum_xx + eps*n = 1.0004
sum_xdz = Σ x[i]*dz[i]   = 1
mean_eps = sum_xx/n + eps = 0.2501
rrms   = 1/sqrt(mean_eps) ≈ 1.9996
dx[0]  = (dz[0] - x[0]*sum_xdz/sum_eps) * rrms
       = (1 - 1.0 * 1/1.0004) * 1.9996
       ≈ 0.0008
dx[1..3] = 0
```

## Repro 2 — backend scheduler `ggml_backend_sched_graph_compute` (wrong)

```c
#include "ggml.h"
#include "ggml-backend.h"
#include "ggml-cpu.h"
#include <stdio.h>

int main(void) {
    ggml_backend_load_all();
    ggml_backend_t backend = ggml_backend_init_by_type(GGML_BACKEND_DEVICE_TYPE_CPU, NULL);
    ggml_backend_sched_t sched = ggml_backend_sched_new(&backend, NULL, 1,
                                                         GGML_DEFAULT_GRAPH_SIZE, false, true);

    size_t ctx_buf_size = ggml_tensor_overhead() * GGML_DEFAULT_GRAPH_SIZE + ggml_graph_overhead();
    uint8_t *ctx_buf = (uint8_t *) calloc(1, ctx_buf_size);
    struct ggml_init_params params = {
        .mem_size   = ctx_buf_size,
        .mem_buffer = ctx_buf,
        .no_alloc   = true,
    };
    struct ggml_context *ctx = ggml_init(params);

    struct ggml_tensor *t_a = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, 4, 1);
    struct ggml_tensor *t_b = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, 4, 1);
    struct ggml_tensor *t_out = ggml_rms_norm_back(ctx, t_a, t_b, 1e-4f);

    struct ggml_cgraph *gf = ggml_new_graph(ctx);
    ggml_build_forward_expand(gf, t_out);

    ggml_backend_sched_reset(sched);
    if (!ggml_backend_sched_alloc_graph(sched, gf)) {
        fprintf(stderr, "alloc_graph failed\n"); return 1;
    }

    float a[4] = {1.0f, 0.0f, 0.0f, 0.0f};
    float b[4] = {1.0f, 0.0f, 0.0f, 0.0f};
    ggml_backend_tensor_set(t_a, a, 0, sizeof a);
    ggml_backend_tensor_set(t_b, b, 0, sizeof b);

    ggml_backend_sched_graph_compute(sched, gf);

    float out[4];
    ggml_backend_tensor_get(t_out, out, 0, sizeof out);
    printf("out = [%g %g %g %g]\n", out[0], out[1], out[2], out[3]);
    return 0;
}
```

**Output**: `out = [-3.9976 0 0 0]`

`-3.9976` is suspicious: it equals roughly `-2 * (1/sqrt(0.2501))` = `-2 * rrms`. Doesn't match any documented formula I can recover from the impl source.

## What's different

Tried during debugging:
- Calling `ggml_set_input` on the two input tensors before
  `ggml_backend_sched_alloc_graph` — no change.
- Downloading the input tensors immediately after
  `ggml_backend_tensor_set` and before `ggml_backend_sched_graph_compute`
  — values are correct (`[1 0 0 0]` for both), so the uploads are landing.
- Forward `ggml_rms_norm` via the same backend-sched setup — works
  correctly (parity verified at multiple shapes against a hand-rolled
  reference in our `TinyNN.rms_norm` smoke test).
- `softmax_back` via `ggml_soft_max_ext_back` through the same
  backend-sched setup — works correctly (parity verified against
  hand-rolled reference). So it's not a "backward ops in general are
  broken via sched" issue, specifically `rms_norm_back`.

## Where I'm using this

Building a Ruby-FFI bridge from a Spinel-AOT-compiled toy transformer
to ggml — repo at
[`OriPekelman/toy_ruby_neural_network`](https://github.com/OriPekelman/toy_ruby_neural_network),
the bridge code in `tinynn/`. The shim around `tnn_session_new` uses
backend-sched (so we get the same code path for CPU and CUDA), which
is why this matters for us — but the repro reproduces in pure C using
just `ggml`, so the issue is upstream.

`TinyNN.rms_norm_back` is currently bound but its parity smoke is
disabled until this is resolved. Happy to test any candidate fix
against our toy transformer's backward pass.

devYRPauli:
Fixes #1491.

### Bug

`GGML_OP_RMS_NORM_BACK` is listed in `ggml_op_can_inplace`, so the backend scheduler may allocate `dst` over the buffer of `src0` (= `dz`, the upstream gradient). On the CPU backend, `ggml_compute_forward_rms_norm_back_f32` was implemented as a four-step sequence:

```c
ggml_vec_cpy_f32  (ne00, dx, x);                       // dx = x  (overwrites dz if dst aliases src0)
ggml_vec_scale_f32(ne00, dx, (float)(-sum_xdz)/sum_eps);
ggml_vec_acc_f32  (ne00, dx, dz);                      // re-reads dz from now-overwritten memory
ggml_vec_scale_f32(ne00, dx, rrms);
```

When `dst` aliases `src0`, the `+= dz` step reads the values written in the first step instead of the original gradient. With the issue's exact inputs `dz = x = [1, 0, 0, 0]`, `eps = 1e-4`, this produces `[-3.9976, 0, 0, 0]` instead of the correct `[0.000799, 0, 0, 0]` — sign flipped, magnitude ~5000× off.

### Fix

Replace the multi-step sequence with a single-pass FMA loop that reads each element before writing it. This is in-place safe under aliasing of `dst` with either `src0` or `src1`. The pattern matches the existing CUDA implementation at `ggml-cuda/norm.cu:205-207`, which has always been correct.

### Regression test

Adds `tests/test-rms-norm-back-inplace.cpp`. Compares `ggml_graph_compute_with_ctx` (no aliasing) against `ggml_backend_sched_graph_compute` (may alias) for three input shapes including the issue's exact reproducer. Without the fix, all three cases fail with significant divergences. With the fix, all pass.

### Out of scope

Quick audit on M1 Pro turned up two pre-existing unrelated failures that reproduce on master without these changes — flagging here in case they help triage:

- `test-backend-ops grad -o SUM` on Metal: `SUM(type=f32,ne=[33,256,1,1],permute=[1,0,2,3])` reports `MAA = 0.997 > 1e-4`.
- `test-backend-ops grad -o RMS_NORM` on Metal: `RMS_NORM(...,inplace=1)` hits `GGML_ASSERT(!node->view_src || ...)` in `ggml_build_backward_expand` (`ggml.c:7037`).

SYCL `rms_norm_back` implementation wasn't audited here; should probably get the same review since it likely uses similar logic.

### DIFF
--- a/src/ggml-cpu/ops.cpp
+++ b/src/ggml-cpu/ops.cpp
@@ -3974,12 +3974,18 @@ static void ggml_compute_forward_rms_norm_back_f32(
                 // dx := scale(dx, rrms)
                 float * dx = (float *) ((char *) dst->data + i01*nb1 + i02*nb2 + i03*nb3);
 
-                // dx[i00] = (x*(-sum_xdz/sum_eps) + dz) / sqrtf(mean_eps)
-                ggml_vec_cpy_f32  (ne00, dx, x);
-                // ggml_vec_scale_f32(ne00, dx, -mean_xdz/mean_eps);
-                ggml_vec_scale_f32(ne00, dx, (float)(-sum_xdz)/sum_eps);
-                ggml_vec_acc_f32  (ne00, dx, dz);
-                ggml_vec_scale_f32(ne00, dx, rrms);
+                // dx[i00] = (dz + x*(-sum_xdz/sum_eps)) * rrms
+                //
+                // GGML_OP_RMS_NORM_BACK is listed in ggml_op_can_inplace, so the
+                // scheduler may alias dst with src0 (dz) or src1 (x). A single
+                // fused read-before-write loop is safe under either aliasing.
+                // A multi-step cpy/scale/acc/scale sequence is NOT safe — if dst
+                // aliases dz, the +=dz step would re-read overwritten memory.
+                // See: https://github.com/ggml-org/ggml/issues/1491
+                const float scale_x = (float) (-sum_xdz) / sum_eps;
+                for (int64_t i00 = 0; i00 < ne00; i00++) {
+                    dx[i00] = (dz[i00] + x[i00] * scale_x) * rrms;
+                }
             }
         }
     }
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -233,6 +233,15 @@ if (NOT GGML_BACKEND_DL)
     set_property(TEST ${TEST_TARGET} PROPERTY ENVIRONMENT "LLVM_PROFILE_FILE=${TEST_TARGET}.profraw")
 
     #
+    # test-rms-norm-back-inplace
+
+    set(TEST_TARGET test-rms-norm-back-inplace)
+    add_executable(${TEST_TARGET} ${TEST_TARGET}.cpp)
+    target_link_libraries(${TEST_TARGET} PRIVATE ggml)
+    add_test(NAME ${TEST_TARGET} COMMAND $<TARGET_FILE:${TEST_TARGET}>)
+    set_property(TEST ${TEST_TARGET} PROPERTY ENVIRONMENT "LLVM_PROFILE_FILE=${TEST_TARGET}.profraw")
+
+    #
     # test-pad-reflect-1d
 
     set(TEST_TARGET test-pad-reflect-1d)
--- a/tests/test-rms-norm-back-inplace.cpp
+++ b/tests/test-rms-norm-back-inplace.cpp
@@ -1,0 +1,186 @@
+// Regression test for ggml-org/ggml#1491: ggml_rms_norm_back produces wrong
+// output via the backend scheduler on the CPU backend when GGML_OP_RMS_NORM_BACK
+// is allocated in-place over its src0 (gradient) input.
+//
+// The op is listed in ggml_op_can_inplace, so the allocator may reuse a parent
+// buffer (src0 = dz or src1 = x) as the destination buffer. The CPU
+// implementation must remain correct under that aliasing.
+//
+// This test runs the same graph two ways:
+//   (1) ggml_graph_compute_with_ctx     — buffers allocated by the legacy
+//                                          context (no aliasing)
+//   (2) ggml_backend_sched_graph_compute — buffers allocated by the scheduler
+//                                          (may alias src0 with dst)
+// and asserts the two results agree element-wise.
+
+#include "ggml.h"
+#include "ggml-alloc.h"
+#include "ggml-backend.h"
+#include "ggml-cpu.h"
+
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <vector>
+
+struct test_case {
+    const char *           name;
+    int64_t                ne0;
+    int64_t                ne1;
+    std::vector<float>     a;   // gradient input (src0, dz)
+    std::vector<float>     b;   // forward x  (src1)
+    float                  eps;
+};
+
+static bool run_legacy(const test_case & tc, std::vector<float> & out) {
+    const size_t n_elem = (size_t) tc.ne0 * (size_t) tc.ne1;
+    const size_t mem_size = 4 * (sizeof(float) * n_elem) + 256 * 1024;
+
+    ggml_init_params params = {};
+    params.mem_size   = mem_size;
+    params.mem_buffer = nullptr;
+    params.no_alloc   = false;
+
+    ggml_context * ctx = ggml_init(params);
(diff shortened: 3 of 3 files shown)


src/ggml-cpu/ops.cpp
```diff
@@ -3974,12 +3974,18 @@ static void ggml_compute_forward_rms_norm_back_f32(
...
+                // GGML_OP_RMS_NORM_BACK is listed in ggml_op_can_inplace, so the
+                // scheduler may alias dst with src0 (dz) or src1 (x). A single
+                // fused read-before-write loop is safe under either aliasing.
+                // A multi-step cpy/scale/acc/scale sequence is NOT safe — if dst
+                // aliases dz, the +=dz step would re-read overwritten memory.
+                // See: https://github.com/ggml-org/ggml/issues/1491
+                const float scale_x = (float) (-sum_xdz) / sum_eps;
+                for (int64_t i00 = 0; i00 < ne00; i00++) {
+                    dx[i00] = (dz[i00] + x[i00] * scale_x) * rrms;
+                }
```
ggerganov, line 3988: Could you submit just this change alone in a PR to https://github.com/ggml-org/llama.cpp?

No need for a regression test.

devYRPauli:
Done — opened ggml-org/llama.cpp#24305 with just this change. Thanks!

devYRPauli:
Closing this one out. The fix landed in master via ggml-org/llama.cpp#24305 - the standalone version you asked for. I confirmed current master already has the single-pass loop in `ggml_compute_forward_rms_norm_back_f32` (with the #1491 note), so this PR is now redundant and only conflicts because it re-applies a change that's already in. Skipping the regression test per your earlier note. Thanks for the review and the steer.

Issue #1448, opened 2026-03-30
`ggml_conv_transpose_2d_p0` silently ignores batch dimension (ne[3] > 1)


janblumenkamp:
`ggml_conv_transpose_2d_p0` produces correct output for the first batch element but zeros for all subsequent batch elements when the input has `ne[3] > 1`. The output shape is computed correctly (preserving the batch dimension), but the CPU compute kernel only processes batch index 0. This issue was co-authored with Claude.

## Root cause

In `ggml/src/ggml-cpu/ops.cpp`, function `ggml_compute_forward_conv_transpose_2d_impl`:

**1. Source data permutation** only iterates over `ne12` (channels), `ne11` (height), `ne10` (width) but **never over `ne13` (batch)**:

```cpp
// permute source data (src1) from (Sw x Sh x Cin) to (Cin x Sw x Sh)
for (int i12 = 0; i12 < ne12; i12++) {      // channels
    for (int i11 = 0; i11 < ne11; i11++) {   // height
        // ... copies ne10 (width) elements
    }
}
// ne13 (batch) is never iterated — only batch 0 data is copied to wdata
```

**2. The compute loop** writes to `dst->data` using offsets that only cover a single batch slice:

```cpp
for (int i2 = ip0; i2 < ip1; i2++) { // output channels only
    float * dst_data = (float *)((char *) dst->data + i2*nb2);
    // ... writes to dst_data[(i11*stride + i01)*ne0 + i10*stride + i00]
    // no iteration over ne3 (batch)
}
```

**3. Work buffer sizing** also doesn't account for the batch dimension:

```cpp
cur += ggml_type_size(src0->type) * ne10 * ne11 * ne12;  // missing * ne13
```

## Reproducer

Feeds 3 identical batch elements through `ggml_conv_transpose_2d_p0`. Expected: all 3 outputs are identical. Actual: batch 0 is correct, batches 1–2 are all zeros.

```cpp
#include "ggml.h"
#include "ggml-alloc.h"
#include "ggml-backend.h"
#include "ggml-cpu.h"

#include <cstdio>
#include <cmath>
#include <vector>

int main() {
    const int IC = 2, OC = 2, KH = 2, KW = 2, stride = 2;
    const int IH = 3, IW = 3, BS = 3;
    const int OH = (IH - 1) * stride + KH;
    const int OW = (IW - 1) * stride + KW;

    ggml_backend_t backend = ggml_backend_cpu_init();
[...]

PR #1520, opened 2026-05-28
ggml : assert b->ne[3] == 1 in ggml_conv_transpose_2d_p0 (#1448)

devYRPauli:
Addresses #1448 via Option A from @janblumenkamp's analysis (assert at the API level, no kernel changes).

### Bug

`ggml_conv_transpose_2d_p0` accepts inputs with `b->ne[3] > 1` and returns a tensor with the correct 4D output shape, but the CPU compute kernel (`ggml_compute_forward_conv_transpose_2d_impl` in `ggml-cpu/ops.cpp`) only iterates batch index 0:

- the `src1` permutation loop iterates `ne12, ne11, ne10` but never `ne13`;
- the compute loop offsets the output by `i2*nb2` (channel) only, not by `i3*nb3` (batch);
- the work-buffer sizing omits a `*ne13` factor.

Net effect: batch 0 is correct, batches 1..N-1 are silently zero. Caught only by explicitly comparing across batch elements. The bug has been present since 8da5be2c (Aug 2023). Every existing caller (SAM example, `test-conv-transpose.c`, `test-backend-ops.cpp` parameterizations) uses `ne[3] == 1`, so it has never surfaced in CI.

### Fix

Add `GGML_ASSERT(b->ne[3] == 1)` to `ggml_conv_transpose_2d_p0`. Matches the existing `GGML_ASSERT(a->ne[3] == 1)` precedent in `ggml_conv_transpose_1d`. Turns silent data corruption into an immediate, debuggable abort.

### Why not implement batch support instead?

Option B (implementing the batch loop in CPU + Metal + Vulkan kernels) is genuinely useful but adds capability for a code path with no current caller. Option A solves the correctness problem and matches existing precedent with minimal risk. If a caller later needs batched `conv_transpose_2d_p0`, removing the assert and implementing batch support across backends is a clean follow-up.

### Verification

- Reporter's reproducer (#1448 body, 3-batch input with identity-like kernel). Before this change it prints `batch 1: sum_abs = 0.0 <-- BUG`. After this change: aborts with `GGML_ASSERT(b->ne[3] == 1) failed` at `ggml.c:4882`.
- Full `test-backend-ops test` (Metal + CPU): no regressions. All `CONV_TRANSPOSE_2D` parameterizations use `ne[3] == 1` and continue to pass.

### No test added

The natural regression test ("assert fires on `ne[3] > 1`") requires a death-test pattern (fork + check exit on SIGABRT) since `GGML_ASSERT` aborts the process. `test-backend-ops`'s compute-and-compare model can't host it. Happy to add a separate death-test executable if maintainers prefer; let me know.

Credit to @janblumenkamp for the original bug analysis, complete reproducer, and historical research.

### DIFF
--- a/src/ggml.c
+++ b/src/ggml.c
@@ -4947,6 +4947,12 @@ struct ggml_tensor * ggml_conv_transpose_2d_p0(
         struct ggml_tensor  * b,
         int                   stride) {
     GGML_ASSERT(a->ne[3] == b->ne[2]);
+    // The CPU/Metal/Vulkan compute kernels currently only iterate batch index 0,
+    // so calling with b->ne[3] > 1 would silently produce zeros for batches 1..N
+    // while reporting the correct 4D output shape. Reject at the API level —
+    // matches the GGML_ASSERT(a->ne[3] == 1) precedent in ggml_conv_transpose_1d.
+    // See: https://github.com/ggml-org/ggml/issues/1448
+    GGML_ASSERT(b->ne[3] == 1);
 
     const int64_t ne[4] = {
         ggml_calc_conv_transpose_output_size(b->ne[0], a->ne[0], stride, 0 /*p0*/),


devYRPauli:
Trimming down to one open PR to respect the new-contributor guideline. Closing this for now and I'll resubmit after my first one lands. Thanks!

devYRPauli:

devYRPauli:

Still reproducible on current master: `ggml_conv_transpose_2d_p0` asserts only `a->ne[3] == b->ne[2]`, so calling it with `b->ne[3] > 1` returns a correctly shaped 4D tensor while the kernels only ever iterate batch index 0, leaving batches 1..N silently zero. Issue #1448 is still open.

This matches the existing `GGML_ASSERT(a->ne[3] == 1)` precedent in `ggml_conv_transpose_1d` - reject at the API level rather than return wrong data.

devYRPauli:
Closing this. Upstream fixed the batch case the other way.

llama.cpp#26132 adds real multi-batch support to `ggml_conv_transpose_2d`. The CPU kernel now loops over the batch dimension, the Metal kernel handles it, and `test-backend-ops` covers the case. @ggerganov closed #1448 with that change on 2026-08-28.

This PR asserts `b->ne[3] == 1`. That guard was correct while the kernel ignored batches. Once the sync reaches this repo, the assert would reject input that now works.

Thanks to @tekinertekin for the kernel fix.

PR #1521, opened 2026-05-28
ggml-cpu : fix soft_max_back wrong output when dst aliases src1 (y)

devYRPauli:
Sibling fix to #1519 (rms_norm_back in-place aliasing). Same root-cause class.

### Bug

`GGML_OP_SOFT_MAX_BACK` is listed in `ggml_op_can_inplace`, so the backend scheduler may allocate `dst` over the buffer of either `src0` (`dy`, the upstream gradient) or `src1` (`y`, the saved softmax output). On the CPU backend, `ggml_compute_forward_soft_max_ext_back_f32` was implemented as a five-step sequence:

```c
ggml_vec_dot_f32  (nc, &dot_y_dy, 0, y, 0, dy, 0, 1);  // dot(y, dy)
ggml_vec_cpy_f32  (nc, dx, dy);                        // dx = dy
ggml_vec_acc1_f32 (nc, dx, -dot_y_dy);
ggml_vec_mul_f32  (nc, dx, dx, y);                     // re-reads y
ggml_vec_scale_f32(nc, dx, scale);
```

The allocator preferentially picks `src0` for in-place reuse, in which case `dst == dy` and the `cpy` is a no-op — historically safe. But when `src0` is ineligible (e.g. marked `GGML_TENSOR_FLAG_OUTPUT`, or used by more than one downstream consumer), the allocator falls through to `src1`. Then the `dx = dy` step overwrites `y` with `dy` values, and the subsequent `dx = dx * y` reads the corrupted buffer — producing wildly wrong gradients (diffs of ~1.5 on values of magnitude ~0.02 in the test cases below).

### Fix

Replace the multi-step sequence with a single fused read-before-write loop:

```c
float dot_y_dy = 0;
ggml_vec_dot_f32(nc, &dot_y_dy, 0, y, 0, dy, 0, 1);
for (int i = 0; i < nc; i++) {
    dx[i] = scale * (dy[i] - dot_y_dy) * y[i];
}
```

Each element's read of `dy[i]` and `y[i]` precedes its write to `dx[i]`, which is safe under aliasing of `dst` with either `src0` or `src1`. Matches the CUDA reference at `ggml-cuda/softmax.cu:268`.

### Regression test

Adds `tests/test-soft-max-back-inplace.cpp`. Compares `ggml_graph_compute_with_ctx` against `ggml_backend_sched_graph_compute` across six configurations:

- `default-*` — allocator picks src0; historically passes on both paths.
- `dy-output-*` — `ggml_set_output(t_dy)` forces allocator to fall through to src1; **previously broken, fixed by this PR**.
- `y-output-*` — allocator picks src0 (src1 ineligible); safe.
- `both-output-*` — both ineligible, fresh allocation; safe.

Without the fix: the four "default-*" and "*-output (src0 path)" cases pass, the two `dy-output-*` cases fail with large divergences. With the fix: all six pass.

### How discovered

Surfaced during the audit prompted by #1519 — same pattern, same backward-op surface (`_back` ops in `ggml_op_can_inplace`). I audited the other entries in that list: `ggml_vec_silu_backward_f32/_f16` (used by SILU_BACK) is a single-pass loop, in-place safe. `ggml_compute_forward_rope_flt` (used by ROPE_BACK) uses pair-wise rotations that the existing impl handles correctly. Only SOFT_MAX_BACK had the multi-step pattern with the same trap as RMS_NORM_BACK.

### Pre-existing unrelated issue noted

`test-backend-ops grad -o SUM` on Metal still produces `inf` for `SUM(type=f32,ne=[33,256,1,1],permute=[1,0,2,3])` — same pre-existing failure noted in #1519's body. Unrelated to this change.

### DIFF
--- a/src/ggml-cpu/ops.cpp
+++ b/src/ggml-cpu/ops.cpp
@@ -5653,12 +5653,19 @@ static void ggml_compute_forward_soft_max_ext_back_f32(
         // dx := dx * y
 
         // linear runtime, no additional memory
+        //
+        // GGML_OP_SOFT_MAX_BACK is listed in ggml_op_can_inplace, so the
+        // scheduler may alias dst with src0 (dy) or src1 (y). The previous
+        // multi-step sequence (cpy/acc1/mul/scale) was unsafe when dst aliased
+        // src1 (y): the `dx = dy` step overwrote y, and the subsequent
+        // `dx = dx * y` then read the corrupted y. A single fused
+        // read-before-write loop is safe under either aliasing and matches the
+        // CUDA reference at ggml-cuda/softmax.cu:268.
         float dot_y_dy = 0;
-        ggml_vec_dot_f32  (nc, &dot_y_dy, 0, y, 0, dy, 0, 1);
-        ggml_vec_cpy_f32  (nc, dx, dy);
-        ggml_vec_acc1_f32 (nc, dx, -dot_y_dy);
-        ggml_vec_mul_f32  (nc, dx, dx, y);
-        ggml_vec_scale_f32(nc, dx, scale);
+        ggml_vec_dot_f32(nc, &dot_y_dy, 0, y, 0, dy, 0, 1);
+        for (int i = 0; i < nc; i++) {
+            dx[i] = scale * (dy[i] - dot_y_dy) * y[i];
+        }
 
 #ifndef NDEBUG
         for (int i = 0; i < nc; ++i) {
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -233,6 +233,15 @@ if (NOT GGML_BACKEND_DL)
     set_property(TEST ${TEST_TARGET} PROPERTY ENVIRONMENT "LLVM_PROFILE_FILE=${TEST_TARGET}.profraw")
 
     #
+    # test-soft-max-back-inplace
+
+    set(TEST_TARGET test-soft-max-back-inplace)
+    add_executable(${TEST_TARGET} ${TEST_TARGET}.cpp)
+    target_link_libraries(${TEST_TARGET} PRIVATE ggml)
+    add_test(NAME ${TEST_TARGET} COMMAND $<TARGET_FILE:${TEST_TARGET}>)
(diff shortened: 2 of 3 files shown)


devYRPauli:
Going to close this for the moment to keep to a single open PR as a new contributor. I've left #1519 open since it fixes the same in-place aliasing problem in the backward ops; I'll bring this soft_max_back one back once that lands. Thanks!

devYRPauli:

devYRPauli:

Re-verified on the current tree. `GGML_OP_SOFT_MAX_BACK` is still listed in `ggml_op_can_inplace` (`src/ggml-alloc.c:44`), so the scheduler may still alias dst with src1.

Built current master (Release, CPU only, Metal/BLAS/Accelerate off, Apple M1 Pro) with only the `ops.cpp` change reverted and the new test kept:

```
[default-8x4] OK
[default-64x16] OK
[dy-output-8x4] FAIL
[dy-output-64x16] FAIL
[y-output-8x4] OK
[both-output-8x4] OK
exit=1
```

With the change applied, all six pass (exit=0).

The failing cases mark src0 as a graph output, which stops the allocator reusing it and makes it fall through to src1. That is the aliasing the current cpy/acc1/mul/scale sequence cannot survive: the `dx = dy` step overwrites y, and the later `dx = dx * y` reads back the corrupted y. The `y-output` case passes either way because aliasing dst onto src0 leaves y intact, which is why the existing tests never caught it.

The replacement is a single fused read-before-write loop, which is safe under either aliasing and matches the CUDA reference in `ggml-cuda/softmax.cu`.

devYRPauli:
I moved this fix to llama.cpp: ggml-org/llama.cpp#27096.

The same code is in llama.cpp at `ggml/src/ggml-cpu/ops.cpp`. Changes to the ggml CPU operators usually land in llama.cpp first, and they come back to this repo through `sync-ggml`. The llama.cpp PR also has a regression test. The test fails before the change and passes after it.

I close this PR to keep the review in one place. Please tell me if you prefer the fix in this repo. I will reopen it.

PR #1528, opened 2026-06-05
vulkan: add e4m3 (fp8) matmul for AMD RDNA4 — portable dequant default + opt-in native fp8

RapidMark:
## Summary

Today AMD RDNA4 (gfx12) GPUs fall back to the RDNA3 code path in ggml-vulkan — there is no
RDNA4-specific path, so the fp8 matrix-multiply hardware that RDNA4 adds (and that RDNA3 does not have)
sits completely unused. The driver even advertises RDNA4's fp8 cooperative-matrix shapes, and the
backend enumerates and logs them, then discards them because nothing consumes them.

This PR adds that missing path. It introduces `GGML_TYPE_E4M3` (scaled fp8 e4m3fn) and a Vulkan
cooperative-matrix matmul for it, plus `AMD_RDNA4` detection.

By default e4m3 is a **portable dequant-to-f16** type: the fp8 weights are stored as-is (half the bytes
of f16), decoded to f16 in-shader, and run through the standard f16 cooperative-matrix path — no fp8
extension, builds with stock glslc, and works on any f16-coopmat stack (NVIDIA, AMD RDNA3/4, Intel), so
everyone gets the VRAM/bandwidth win of 1-byte weights. On RDNA4, the opt-in build flag
`-DGGML_VULKAN_FP8_NATIVE=ON` runs the weights **natively through RDNA4's fp8 WMMA** (`fp8×fp8→f32`) — on
a Radeon AI PRO R9700 this is **~2× the matmul throughput of f16** and **1.18× faster than Q4_K** at the
whole-image level, at visually-equivalent quality. A scalar CPU `vec_dot` is included as the
`test-backend-ops` reference. The native fp8 path is RDNA4-gated; the portable e4m3 type runs on any
coopmat device; all other types and backends are untouched.

## The problem this addresses

`vk_device_architecture` enumerated `GCN, RDNA1, RDNA2, RDNA3` — **no `AMD_RDNA4`** — and the detector
returns `AMD_RDNA3` for any RDNA card with `integerDotProduct4x8BitPackedMixedSignedness`, true for both
gfx11 (RDNA3) and gfx12 (RDNA4). So an RDNA4 card is classified RDNA3. Meanwhile RDNA4's driver
**advertises fp8 cooperative-matrix shapes** that the backend enumerates, logs, and then discards — the
coopmat-config loop has a `Float16` branch and an `Sint8` branch but **no `Float8` branch**. Net: RDNA4's
fp8 matrix hardware is detected and left idle.

```
# R9700 / RADV GFX1201, vkGetPhysicalDeviceCooperativeMatrixPropertiesKHR:
  total coopmat configs: 20   |   fp8 (e4m3/e5m2) configs: 4
    M=16 N=16 K=16  A=FLOAT8_E4M3_EXT  B=FLOAT8_E4M3_EXT  C=f32  Result=f32   (+ e4m3×e5m2, e5m2×…)
```

## The change

**Core type** (`ggml.h`, `ggml-common.h`, `ggml-impl.h`, `ggml.c`, `ggml-quants.{c,h}`)
- `GGML_TYPE_E4M3`: `block_e4m3 = { fp16 d; e4m3 qs[32] }` (q8_0-shaped, 34 B), per-row scale.
  `quantize`/`dequantize_row`, type-traits, fp8 pack/unpack helpers.

**CPU reference** (`ggml-cpu/…`)
- Scalar `ggml_vec_dot_e4m3_f32` (block dequant · f32) — the correctness oracle for the Vulkan kernel in
  `test-backend-ops`; no CPU perf claim.

**Vulkan: detection + kernels** (`ggml-vulkan.cpp`, `vulkan-shaders/{mul_mm.comp, mul_mm_funcs.glsl,
types.glsl, vulkan-shaders-gen.cpp}`)
- Add `AMD_RDNA4`, detected via `VK_EXT_shader_float8` (gfx12 has fp8 WMMA + the extension; gfx11 has
  WMMA but no fp8, so the extension is absent). Widen the AMD coopmat gate to `RDNA3 || RDNA4` so the enum
  split is not a latent coopmat regression.
- **Portable e4m3 matmul (default):** `FLOAT_TYPE=float16_t`; decode fp8→f16 at load (per-row scale
  applied), feed the standard f16 coopmat path. No fp8 extension. Reads `qs` through a single `uint8`
  view. Runs on any coopmat device.
- **Native fp8 matmul (opt-in `-DGGML_VULKAN_FP8_NATIVE=ON`):** `FLOAT_TYPE=floate4m3_t`, weights loaded
  raw, pure fp8×fp8 WMMA over the whole K, f32 accumulator (the only acc the hardware advertises for
  fp8); per-row scale applied in the store epilogue; activations cast f16→fp8 in B-staging. The native
  e4m3 type, pipeline and `supports_op` are gated to `AMD_RDNA4`. Requires a glslang that supports the
  `GL_EXT_float_e4m3` extension (validated: glslang `main` @ `98beacd`, `--target-env vulkan1.2`).

**Dispatch fix** (`ggml-vulkan.cpp`, `ggml_vk_get_mul_mat_mat_pipeline`)
- e4m3 reads the activation (`src1`) as **f32** and has no q8_1/integer-dot MMQ variant. `mul_mat`
  enables `quantize_y` (quantize src1 → q8_1) whenever the device reports integer-dot support, then asks
  for the `(src0, GGML_TYPE_Q8_1)` pipeline. Guard the e4m3 branch with `src1_type != GGML_TYPE_Q8_1` so
  that request returns null → the caller drops `quantize_y` and keeps src1 as f32. Without this, on a
  build where glslc enables integer-dot, the f32-reading e4m3 shader is fed q8_1 bytes → NaN. Covers both
  the portable and native paths.

**Test** (`test-backend-ops.cpp`): `GGML_TYPE_E4M3` added to the MUL_MAT type list (+ init / to_float).
`test_mul_mat::max_nmse_err()` returns an fp8-appropriate tolerance (`1e-2`) for e4m3 — the native
fp8×fp8 path deviates from the dequant-to-f32 CPU reference by ~0.1–0.4% NMSE (expected fp8 precision),
mirroring the existing MXFP4/NVFP4 native-fp4 relaxation, tighter for e4m3's extra mantissa.

## Validation (AMD R9700 gfx1201 / RADV / Mesa 25.2.8 unless noted)

Both paths validated on the **same final tree** (the squashed commit), each with **integer-dot ON** (the
condition that triggered the NaN), on the R9700:
- **Portable default** (`GGML_VULKAN_FP8_NATIVE` off, stock glslc 2026.2, no fp8 extension):
  `test-backend-ops -o MUL_MAT` **957/957**, all 10 e4m3 cases OK on Vulkan0; full suite green; coopmat
  still engages (gate fix verified).
- **Native path** (`-DGGML_VULKAN_FP8_NATIVE=ON`, fp8-capable glslang `main` @ `98beacd`):
  `test-backend-ops -o MUL_MAT` **957/957**, all 10 e4m3 cases OK.
- **Dispatch fix** (the e4m3-with-integer-dot NaN), both configs that had been failing now pass:
  R9700 / glslc shaderc 2026.2 → **957/957**, all 10 e4m3 OK; RX 7800 XT (RDNA3) / glslc shaderc 2025.1 →
  **957/957**.
- **Native vs portable speed** (test-backend-ops perf, m=4096, k=14336; portable = what a non-fp8 card
  runs, so this is the RDNA4 native gain with no hardware confound):

  | n | portable dequant-f16 | native fp8 | speedup |
  |---|---|---|---|
  | 1 | 141.8 GFLOPS | 257.3 GFLOPS | 1.81× |
  | 4 | 538.0 | 1030 | 1.91× |
  | 8 | 969.7 | 1750 | 1.80× |
  | 512 (compute-bound) | 29.9 TFLOPS | 61.5 TFLOPS | **2.06×** |

- **End-to-end gen** (sd.cpp, Flux, 1024²/20-step/seed 42): Q4_K sampling 44.1 s → native e4m3 37.5 s
  (1.18× whole-image, ~2× at the matmul vs f16); visually-equivalent image (same-seed cross-quant
  equivalence). e4m3 is 1 byte/weight vs Q4's 0.5 — it reads *twice* the weight bytes and is still
  faster, i.e. the win is the fp8 matrix hardware, not bandwidth.
- **RDNA3 control:** a gfx11 card lacks `VK_EXT_shader_float8` → classified `AMD_RDNA3`, so the *native*
  fp8 path is correctly not built; the *portable* e4m3 type still runs there (verified on an RX 7800 XT,
  957/957 above). All other types/backends unchanged.

## Scope / non-goals
- Only `AMD_RDNA4` gains the native path; every other arch/backend is unchanged (native e4m3 type,
  pipeline, `supports_op` are RDNA4-gated). The portable e4m3 type is available on any f16-coopmat device.
- **Follow-ups (out of scope):** (1) per-block weight scaling (this PR is per-row — correct + faster;
  per-block is the natural quality follow-up); (2) RDNA4-specific warptile tuning (borrows
  `warptile_mmq`); (3) the CPU `vec_dot` is scalar (reference only).


RapidMark:
This update makes e4m3 **portable by default** — moving the native fp8 path to an opt-in build flag so
e4m3 works on any coopmat GPU, not just RDNA4 — and **fixes an activation-dispatch bug** found while
validating it. Force-pushed as a single squashed commit; two changes:

**1. Portable by default, native fp8 opt-in.** To compile the native fp8 shader we had to build glslang
from its development branch — fp8 support isn't in the released glslang that ships with the Vulkan SDK
yet. That worked on our box, but with native fp8 as e4m3's *only* path it was too fragile to ship that
way: anyone building with the stock SDK toolchain can't compile it, and non-RDNA4 cards have no fp8
hardware to run it anyway. So e4m3 is now a portable dequant-to-f16 type by default — no fp8 extension,
builds with the stock glslc everyone already has, and runs on any f16-coopmat GPU, so RDNA3 / NVIDIA /
Intel get the 1-byte-weight VRAM/bandwidth win too. The native fp8 WMMA path (~2× f16 on RDNA4) is the
opt-in.

If you want that speedup, build glslang from `main` (that's where the fp8 support lives for now) and set
`-DGGML_VULKAN_FP8_NATIVE=ON` — otherwise the portable default just works on whatever glslc you already
have.

**2. Fix an e4m3 activation-dispatch bug.** e4m3 mul_mat reads its activation (src1) as f32 and has no
q8_1 MMQ variant, but when a device's integer-dot path is active the dispatch was quantizing src1 to q8_1
and still handing it to the e4m3 pipeline — wrong input, wrong result. `ggml_vk_get_mul_mat_mat_pipeline`
now rejects a q8_1 request for e4m3, so the activation stays f32. Covers both the portable and native
paths.

Verified with `test-backend-ops -o MUL_MAT`: **957/957** on RDNA4 (R9700) for both the portable default
and the native path, and on RDNA3 (RX 7800 XT) for the portable path. The description above is updated to
match, and thanks for flagging the e4m3 issue.

En3Tho:
I wonder if these improvements can be somehow utilized in llamacpp or whispercpp too

RapidMark:
@En3Tho yes — it's implemented as a ggml-vulkan type + cooperative-matrix matmul, so it lives in the same backend llama.cpp and whisper.cpp build on; nothing here is sd.cpp-specific. The portable e4m3 default (1-byte weights decoded to f16 in-shader, stock glslc, any coopmat GPU) gives the VRAM/bandwidth win, and the opt-in native RDNA4 fp8 WMMA path helps the compute-bound matmuls — both relevant to LLM/ASR inference.

One caveat for LLMs specifically: e4m3 is 1 byte/weight, so on *size* it sits between f16 and the sub-byte K-/I-quants llama.cpp already has — the appeal there is the native fp8 hardware throughput and a simple format more than beating Q4_K on bytes. The kernel itself is reusable as-is; each project would just wire the type into its model loader.

RapidMark:
**Correction to the numbers in this comment.** The 9070 XT table I first posted showed e4m3 ≈ 5.8× over f16 — that ratio is an artifact, not a real result. The f16 baseline on that box was anomalously low: f16 came in *below* q8_0 and q4_K (2020 < 2330 < 2590 GFLOPS), which is backwards for a type with no dequant step. The f16 matmul path was underperforming on that particular driver stack, so the denominator was broken — the fp8 number wasn't inflated. Corrected figures below.

The headline still stands: the native fp8 path (`-DGGML_VULKAN_FP8_NATIVE=ON`) builds and runs on a **consumer Radeon RX 9070 XT** (RDNA4 / gfx12), not only the workstation R9700.

**End-to-end gen — the number that actually matters.** Flux Krea-dev, 1024², modified DPM++ (2M), seed 42, R9700 (RDNA4, Vulkan):

| | sampling | total |
|---|---:|---:|
| native fp8 e4m3 | **36.3 s** | 39.6 s |
| Q4_K_S | 43.3 s | 45.8 s |

→ **≈ 1.19× faster, at visually equivalent quality** (same seed/prompt, side-by-side checked).

**Matmul** (`test-backend-ops perf -o MUL_MAT`, compute-bound shape) on the R9700 under Linux/RADV, where the f16 path is healthy:

| type | GFLOPS |
|------|-------:|
| e4m3 (native fp8) | ~77260 |
| f16 | ~48300 |

→ **e4m3 ≈ 1.6× f16** at the compute-bound matmul — the realistic figure, not ≈ 5.8×. At small n (memory-bound) e4m3 is modest, as expected for a compute accelerator.

One footprint caveat worth stating: e4m3 is 8-bit, so it carries ~2× the weight VRAM of Q4 — a speed+quality win on ≥24 GB cards, while a 12 GB tier still prefers Q4. Net: native fp8 is a real ~1.2× end-to-end / ~1.6× matmul win on RDNA4 (consumer 9070 XT included), just not the ≈ 5× the broken-f16 table implied.

RapidMark:
Moved to ggml-org/llama.cpp#24987. ggml-vulkan changes land in llama.cpp first and sync back into ggml, so this belongs there. The new PR carries the validation (test-backend-ops e4m3 on NVIDIA and AMD RDNA4) and the fp8-vs-f16 numbers.

PR #1534, opened 2026-06-10
sycl: fix soft_max_f32 max reduction

someoneinjd:
Fix the second-stage max reduction in the SYCL softmax kernel when a workgroup contains more subgroups than `WARP_SIZE`.

Previously, after each subgroup wrote its partial max to `buf_iw[warp_id]`, the cross-subgroup reduction only loaded:

```cpp
max_val = buf_iw[lane_id];
```

This only reduces the first `WARP_SIZE` partial maxima. For example, on Intel GPUs, `WARP_SIZE` can be 16 while the softmax workgroup size is 1024, producing nwarps = 64. In that case, partial maxima from `buf_iw[16..63]` were ignored.
If the true row max is in one of the ignored subgroups, softmax subtracts a too-small max value. For large attention logits this can overflow `exp(x - max)` to `inf`, eventually producing `NaN`.

llama.cpp PR: https://github.com/ggml-org/llama.cpp/pull/24451

### DIFF
--- a/src/ggml-sycl/softmax.cpp
+++ b/src/ggml-sycl/softmax.cpp
@@ -56,7 +56,7 @@ static void soft_max_f32(const float *         x,
                                : block_size_template;
     const int nthreads = block_size;
     const int nwarps = nthreads / WARP_SIZE;
-    size_t nreduce = nwarps / WARP_SIZE;
+    const size_t nreduce = nwarps / WARP_SIZE;
 
     const int tid = item_ct1.get_local_id(2);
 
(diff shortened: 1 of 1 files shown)


someoneinjd:
This fix has already been merged into `llama.cpp` (https://github.com/ggml-org/llama.cpp/commit/d8a3f523c8da0111b8bcf2d73ca50819579fca4b). What should I do next to get this fix merged into `ggml`?

ggerganov:
It will be synced soon: https://github.com/ggml-org/ggml/pull/1547

PR #1536, opened 2026-06-10
CUDA/HIP: honor GGML_PREC_F32 in the flash-attention tile kernel

RapidMark:
## Summary

The flash-attention **tile** kernel (`fattn-tile.cuh`) selects fp16 vs fp32 storage for Q/K and the KQ buffer from the compile-time arch capability (`FAST_FP16_AVAILABLE`) only, and ignores the op's requested precision (`ggml_flash_attn_ext_get_prec(dst) == GGML_PREC_F32`). On a fast-fp16 arch it therefore always down-casts Q and K to fp16 before the dot product, even when the graph asked for fp32.

The fp16 down-cast is usually harmless, but it is **value/scale dependent**: with large activations and a long key length the per-element rounding of Q·K compounds across layers and sampling steps into a visibly corrupted result (a doubled / "ghosted" subject in an image-diffusion model). This is the kernel AMD GPUs fall back to for **non-GQA** attention when built without rocWMMA flash-attention, so it is reachable in practice.

This PR makes the tile kernel **honor `GGML_PREC_F32`**: when fp32 is requested on a fast-fp16 arch, Q/K/KQ are stored in fp32. Where the wider fp32 buffers cannot fit shared memory for a given head size, the kernel transparently keeps the existing fp16 path (no regression for those shapes).

## Reproduction

Chroma (Flux-schnell arch, non-GQA, head_dim 128) via stable-diffusion.cpp on an AMD RDNA4 GPU (gfx1201), HIP, ggml built **without** rocWMMA flash-attention:

- 256² / 768²: clean.
- 1024² (≈4608 keys, fp32 model): **ghosted/doubled subject.**

It is invisible to `test-backend-ops` (synthetic near-uniform inputs do not excite the fp16 rounding) and to small/low-step generations. A **CPU-backend** run of the identical graph is clean; a per-node CPU-vs-HIP comparison (`ggml_backend_compare_graph_backend`) puts the first significant divergence on the `FLASH_ATTN_EXT` node (NMSE ≈ 2.2e-2 there vs ~1e-5 on every feeding op).

## Change

- Add a `bool prec_f32` template parameter to `flash_attn_tile`; `launch_fattn_tile_switch_ncols1` reads `ggml_flash_attn_ext_get_prec(dst)` (under `#ifdef GGML_USE_HIP`) and instantiates accordingly.
- `use_fp16 = fast_fp16_avail && !prec_f32` drives conditional buffer **types** (`std::conditional_t`) and **sizes**. The per-branch `#ifdef FAST_FP16_AVAILABLE` blocks in `flash_attn_tile`, `flash_attn_tile_iter` and `flash_attn_tile_iter_KQ` become `if constexpr (use_fp16)` (plus a `kpack = use_fp16 ? 2 : 1` factor on the already-templated `T_vec_dot` buffer type), so there is no logic duplication.
- fp32 storage roughly doubles shared memory. A constexpr helper shrinks `nbatch_K` (kept a multiple of 8 and a divisor of `DV`) until the `Q_tmp + KV_tmp + KQ` allocation fits one block, and forces occupancy to 1. If no fitting `nbatch_K` exists (e.g. head size 512/576, where `Q_tmp` alone exceeds the budget) it returns 0 and the kernel **falls back to the fp16 path** for that shape — those shapes are byte-for-byte unchanged.

Scope: gated to HIP. On CUDA the tile kernel is the pre-Pascal fallback (already using the fp32 path) and modern CUDA uses the mma kernel for these shapes, so the CUDA tile path is unchanged; the gating could be extended to CUDA if desired.

## Validation

AMD Radeon AI PRO R9700 (gfx1201), ROCm 7.2.4, ggml `master`, **without** rocWMMA flash-attention, Chroma1-HD fp16, FA + mask, 1024², 20 steps, seed 42:

- **Before:** ghosted/doubled subject.
- **After:** clean, correct image (matches the CPU / CUDA / Vulkan reference and the rocWMMA-fattn output).

`test-backend-ops` unchanged; 256²/768² (already clean) stay clean; 512/576 head sizes fall back to fp16.

## Performance / when it matters

The fp32 tile path is a **shader fp32** path (no matrix cores), so it is substantially slower than the fp16 tile path (~6× on the validated 1024² case) — it is a **correctness** fix. On RDNA3/RDNA4 the fast *and* correct path is `-DGGML_HIP_ROCWMMA_FATTN=ON` (the matrix-core fp32 WMMA kernel); this fix matters most for AMD **without** rocWMMA flash-attention (e.g. RDNA2, or rocWMMA < 2.0), where the tile kernel would otherwise return silently wrong results for an fp32-precision request.

## Notes

- The fp32 `nbatch_K`/occupancy heuristic (60 KiB budget, occupancy 1, `nbatch_K` on multiples of 8) is deliberately conservative to guarantee it fits; happy to replace it with tuned per-`(DKQ, DV, ncols)` fp32 config entries (mirroring `..._nvidia_fp32`) if you prefer that style.
- Only `fattn-tile.cuh` changes; the no-mask / non-fp32 / fp16 paths are unchanged.

Surfaced via [leejet/stable-diffusion.cpp#1625](https://github.com/leejet/stable-diffusion.cpp/pull/1625) (masked flash-attention shape fix); that PR's HIP note describes this corruption as a separate ggml-side issue.

_Opening as a draft pending maintainer preference on the fp32 config style (conservative heuristic vs. explicit tuned table)._

### DIFF
--- a/src/ggml-cuda/fattn-tile.cuh
+++ b/src/ggml-cuda/fattn-tile.cuh
@@ -370,6 +370,33 @@ static constexpr __device__ int ggml_cuda_fattn_tile_get_nbatch_K(const int DKQ,
     return (ggml_cuda_fattn_tile_get_config(DKQ, DV, ncols) >> 23) & ((1 << 9) - 1);
 }
 
+// The GGML_PREC_F32 path stores Q/K/KQ in fp32 instead of fp16, which roughly doubles the
+// static shared-memory footprint of the tile kernel. On a fast-fp16 arch (where the fp16 config
+// was selected) this can exceed the 64 KiB LDS limit for a single block, so derive an LDS-safe
+// nbatch_K by shrinking it - keeping DV % nbatch_K == 0 and nbatch_K % cpy_ne == 0 - until the
+// Q_tmp + KV_tmp + KQ allocation fits one block. Occupancy is forced to 1 separately.
+static constexpr __device__ int ggml_cuda_fattn_tile_get_nbatch_K_f32(
+        const int DKQ, const int DV, const int ncols, const int nbatch_fa, const int nbatch_K_f16) {
+    constexpr int warp_size = 32;
+    const int DVp    = (DV + 2*warp_size - 1) & ~(2*warp_size - 1);
+    const int cpy_ne = ggml_cuda_get_max_cpy_bytes() / 4;
+    const int budget = (60*1024) / 4; // floats; stay safely below 64 KiB LDS for one block
+    // nbatch_K must be a multiple of 8 (flash_attn_tile_load_tile's J % 8 == 0) and divide DV.
+    // Return the largest such value <= the fp16 config's nbatch_K whose fp32 buffers fit, else 0
+    // (caller then keeps the fp16 path - fp32 storage is infeasible for this shape, e.g. DKQ 512).
+    const int step = 8;
+    for (int nk = (nbatch_K_f16 / step) * step; nk >= step; nk -= step) {
+        if ((DV % nk) != 0) {
+            continue;
+        }
+        const int floats = ncols*DKQ + (nbatch_fa*(nk + cpy_ne) + DVp - DV) + ncols*nbatch_fa;
+        if (floats <= budget) {
+            return nk;
+        }
+    }
+    return 0;
+}
+
 // TODO: deduplicate with mma-f16
 template<int warp_size, int nwarps, int I, int J, int J_padding, bool oob_check>
 static __device__ __forceinline__ void flash_attn_tile_load_tile(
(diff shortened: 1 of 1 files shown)


RapidMark:
**Validation — HIP/ROCm tile kernel, Linux.** Tested on an AMD Radeon AI PRO R9700 (gfx1201), ROCm 7.2.4, Linux, with ggml built **without** rocWMMA flash-attention, so non-GQA attention falls to this tile kernel. Repro: Chroma (Flux-schnell arch, non-GQA, head_dim 128) via stable-diffusion.cpp, masked flash-attention, 1024², identical prompt/seed across all renders; `GGML_PREC_F32` is requested on the FA node in every case.

Controlled before/after — the only difference between the two HIP renders is this PR's `fattn-tile.cuh` (stock-0.14 vs patched), same tree, same command:

- **stock-0.14 tile (fp16):** the requested `GGML_PREC_F32` is ignored, Q/K are down-cast to fp16, and the result is a corrupted/ghosted subject. NMSE 0.223 vs the rocWMMA fp32 reference.
- **this PR (fp32 tile):** honors `GGML_PREC_F32`; an instrumented build confirms the fp32 path engages for this shape (`force_f32=1, use_fp16=0`; `nbatch_K` 64→32 to fit shared memory, no fp16 fallback). NMSE 0.0157 vs rocWMMA — clean, the ghost is gone.
- **rocWMMA fp32-WMMA kernel:** an independent fp32 reference. (A CPU-backend render of the identical graph is in progress as an additional cross-check.)

**Scope:** affects the HIP/ROCm fattn **tile** kernel (the non-GQA fallback when built without rocWMMA fattn) on Linux. The CUDA mma path and the Vulkan/Windows paths are unaffected.

One open call for your preference: the precision read is gated to HIP (`#ifdef GGML_USE_HIP` in `launch_fattn_tile_switch_ncols1`). The `if constexpr (use_fp16)` refactor makes honoring `GGML_PREC_F32` on CUDA too trivial — effectively a no-op for modern CUDA (the mma kernel handles these shapes) but semantically uniform. Happy to extend the gate or keep it HIP-scoped.

A before/after/reference render comparison follows in a reply.

Marking ready for review.

RapidMark:
Moved to ggml-org/llama.cpp#24984. Backend changes land in llama.cpp and sync into this repo, so the fix belongs there. Closing this one in favor of that PR.

PR #1540, opened 2026-06-12
sync : llama.cpp

ggerganov:


danbev (approved):

lazka:
(fyi, whisper.cpp could need a sync too, currently doesn't build with current ggml)

ggerganov:
Should be fixed now.

lazka:
confirmed, thanks!

PR #1551, opened 2026-06-30
tests: Take over test-{alloc,gguf} from llama.cpp

ckastner:
They don't use anything from llama.cpp, it's ggml-only. This enables a source-level separation of llama.cpp and ggml.

The tests are removed from llama.cpp in [llama.cpp PR25179](https://github.com/ggml-org/llama.cpp/pull/25179).


ckastner:
Closing, as per [llama.cpp PR25179](https://github.com/ggml-org/llama.cpp/pull/25179#issuecomment-4890216845).

PR #1569, opened 2026-07-17
webgpu : add CONV_2D_DW (depthwise conv2d) kernel

m1el:
Implements GGML_OP_CONV_2D_DW on the WebGPU backend, ported from the Vulkan backend's conv2d_dw.comp so the index math matches the CPU reference exactly.

- conv2d_dw.wgsl: WHCN and CWHN variants selected by a compile define, each using the reference per-channel indexing (no input-channel accumulation). Slim scalar params (channels/batches/dims/stride/pad/dilation), WebGPU-style type defines, offset handling, and 2D workgroup dispatch.
- get_conv2d_dw_pipeline keyed on {types, whcn}; encode fn selects WHCN vs CWHN via ggml_is_contiguous / ggml_is_contiguous_channels; supports_op gates on those two layouts.

### DIFF
--- a/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
+++ b/src/ggml-webgpu/ggml-webgpu-shader-lib.hpp
@@ -355,6 +355,30 @@ struct ggml_webgpu_conv2d_pipeline_key_hash {
     }
 };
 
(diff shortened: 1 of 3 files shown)


m1el:
(closing in favor of contributing to llama.cpp)

PR #1573, opened 2026-07-26
opencl: bugfix  increase ref_count in  ggml_backend_opencl_init()

akleine:
This patch is important at program end in   `ggml_backend_opencl_context`'s` free()` method.

If we would not increase ref_count, we would later end with a `ref_count` of -1 and thus
some code there would not executed, e.g. flushing profiling data.  (`#ifdef GGML_OPENCL_PROFILING`)


akleine:
closed here and moved to https://github.com/ggml-org/llama.cpp/pull/26162

PR #1592, opened 2026-08-20
Kernel optimizations for Krea2 sd.cpp

pwilkin:
Note: this is not llama.cpp related, so I'm dropping it here. Full set of optimizations to bring Krea2 generation down from 77 to 29 seconds.

### DIFF
--- a/include/ggml-rpc.h
+++ b/include/ggml-rpc.h
@@ -11,7 +11,7 @@ extern "C" {
 #define RPC_PROTO_PATCH_VERSION    1
 
 #ifdef  __cplusplus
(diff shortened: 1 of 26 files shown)


pwilkin:
@0cc4m slop drop!

Green-Sky:
I think this would still go to llama.cpp though.

Also what specifically is this, you include the convrot changes, do you improve on them or are they independent.

ggerganov:
Yup, we should still merge in llama.cpp for larger CI coverage. Btw, there is this PR: https://github.com/ggml-org/llama.cpp/pull/26914 - not sure if related.

pwilkin:
@ggerganov yeah, this goes on top.

pwilkin:
@Green-Sky I include them since this is work on the sd.cpp codebase that already has them.

PR #1620, opened 2026-09-06
Fix builds on s390x

WhyNotHugo:
The CPU backend on s390x (z15/z16) is compiled with -march=z15 / -march=z16 -mvx, but these -march flags aren't passed to the linker, so it uses the baseline -march=z196, which fails to build with:

    lto1: error: hardware vector support not available on z196

Pass these flags to the linker when doing LTO. The same fix also affects riscv64 / -march=rv64gc_v.

Fixes: https://github.com/ggml-org/llama.cpp/pull/19341


WhyNotHugo:
This also fixes builds on riscv64.

PR #1626, opened 2026-09-12
sync : llama.cpp

ggerganov:


ggerganov:
@ggml-org/ggml-vulkan In this most recent sync of the ggml, we still observe the ARGSORT failures: https://github.com/ggml-org/ggml/actions/runs/34771629519/job/103762155728?pr=1626#step:3:46868

I think the failures also happen in the `llama.cpp` repo, but they are intermittent. Here is one from yesterday (https://github.com/ggml-org/llama.cpp/actions/runs/34748256417/job/103700046957#step:3:48049)

Related https://github.com/ggml-org/llama.cpp/pull/28705

jeffbolznv:
I've borrowed a turing system and am able to reproduce the intermittent failure. Looking into it..

